Issue #35, opened 2026-04-14
scripts/codex-review-check.sh — new merge-gate helper


nathanjohnpayne:
## Scope

Create `scripts/codex-review-check.sh`:

- Inputs: `PR_NUMBER`
- Runs `gh pr checks <PR>` for CI state
- Runs `gh pr view --json reviews,statusCheckRollup` for review states
- Evaluates merge gate: CI green + internal reviewer approved + Codex approving + no unaddressed P0/P1 findings
- Exits 0 if mergeable; non-zero with a one-line human-readable reason otherwise
- Read-only: never mutates PR state
- `chmod +x`

## Blocked by

Decision: Clean-PR behavior rule (Phase 1)

## Acceptance

- Script is read-only, verified by grep
- All exit codes documented in a comment at the top

---
Part of [Project #2 — Phase 4 Codex-in-GitHub automation](https://github.com/users/nathanjohnpayne/projects/2).

nathanjohnpayne:
## Refinement from #31 bootstrap (2026-04-14)

The merge gate must **not** wait for an `APPROVED` state review from Codex — it never arrives.

### Corrected merge gate

```
codex-review-check.sh exits 0 (mergeable) iff ALL of:

  (a) `gh pr checks <PR>` reports all required checks green
  (b) An `APPROVED` review from a login in `available_reviewers`
      (nathanpayne-claude, nathanpayne-cursor, nathanpayne-codex) is
      present on the PR, from an author != PR author
  (c) Codex has cleared on or after the latest commit via one of:
        • A `COMMENTED` review from `chatgpt-codex-connector[bot]`
          with no unaddressed P0/P1 inline findings on
          `/pulls/{pr}/comments` (filter by commit_id to ignore
          comments from superseded commits), OR
        • A 👍 / `+1` reaction from `chatgpt-codex-connector[bot]`
          on the PR issue (from `/issues/{pr}/reactions`), with
          reaction.created_at >= latest commit timestamp

Exits non-zero with a one-line reason otherwise. Read-only. Never
mutates PR state.
```

### What "unaddressed" means for inline findings

An inline finding is "unaddressed" if:
- It is anchored to a path:line that still exists in the current HEAD diff, AND
- It has no reply from the author (`nathanjohnpayne`) OR the reply does not explicitly rebut the finding (reply pattern: `Not applicable because ...` or `Fixed in <sha>` — exact matching TBD; may need a convention)

[...]

Issue #34, opened 2026-04-14
scripts/codex-review-request.sh — new helper


nathanjohnpayne:
## Scope

Create `scripts/codex-review-request.sh`:

- Inputs: `PR_NUMBER` (required), `REPO` (optional, defaults to current)
- Posts `@codex review` comment using `OP_PREFLIGHT_AUTHOR_PAT`
- Polls `gh pr view --json reviews,comments` every 15 seconds up to 10 minutes
- Emits JSON to stdout: `{state, submitted_at, body, findings: [...]}`
- Logs status to stderr
- Exit codes: 0 review received, 2 timeout, 3 API error
- `chmod +x`

## Blocked by

Decision: Clean-PR behavior rule (Phase 1) — determines how to interpret silence

## Acceptance

- Script is idempotent (safe to re-run mid-loop)
- JSON output is machine-parseable
- Follows template shell script conventions (`set -euo pipefail`, sourced from op-preflight env)

---
Part of [Project #2 — Phase 4 Codex-in-GitHub automation](https://github.com/users/nathanjohnpayne/projects/2).

nathanjohnpayne:
Unblocked by #27. On timeout polling `chatgpt-codex-connector[bot]` review: exit code `4` = `FALLBACK_REQUIRED`; stderr signal `MANUAL_FALLBACK_REQUIRED: no Codex app review within {timeout}s`. Non-zero exit = escalate to Phase 4b, not a broken state. Decision favors explicit review (Option 2): never treat silence as implicit approval.

nathanjohnpayne:
## Refinement from #31 bootstrap (2026-04-14)

The script needs to poll **two endpoints**, not one, and emit a richer JSON payload than the original plan assumed.

### Dual-endpoint polling

- `/pulls/{pr}/reviews` — review state + top-level body (minimally useful; Codex always uses `COMMENTED` state regardless of findings, so this doesn't distinguish "has findings" from "no findings")
- `/pulls/{pr}/comments` — inline diff comments with the actual findings. Codex's findings appear here, not in the review body. Parse for `![P{0-3} Badge]` markdown shortcodes to extract priority (P0/P1/P2/P3).
- `/issues/{pr}/reactions` — Codex's no-findings signal is a 👍 / `+1` reaction on the PR **issue**, not a review state change. The poll loop must also check reactions.

### Emitted JSON structure

```json
{
  "state": "COMMENTED",
[...]

Issue #57, opened 2026-04-15
Add local test harness for pr-review-policy.yml and agent-review.yml shell logic


nathanjohnpayne:
## Problem

The three bugs fixed in #54 / PR #55 were all discoverable only by running the workflow against a real PR. There is no local test harness for the shell logic inside `.github/workflows/pr-review-policy.yml` or `.github/workflows/agent-review.yml`'s `load-config` step. The bugs were:

1. **YAML parse failure** — caught by `ruby -r yaml` *after* I already suspected a workflow file issue
2. **awk range self-terminating** — caught only after bug 1 was fixed and the workflow started running, and only surfaced because the `External Review Check` step returned the wrong answer
3. **Glob-to-regex double-transform** — caught only after bugs 1 and 2 were fixed and the matcher *still* didn't fire
4. **Parallel bugs in `agent-review.yml`** — caught by nathanpayne-codex's external review, not by any automated check

All four were invisible in the existing CI. A simple unit-level harness would have caught every one at commit time.

## Proposal

Add `scripts/ci/test_pr_review_policy.sh` (and equivalent for `agent-review.yml`'s `load-config` step) that:

1. Runs the shell script in isolation against a fixture `review-policy.yml` containing known entries
2. Asserts `PATHS` extraction returns the expected list (vs. the empty-array failure mode of the old awk range)
3. Asserts `REVIEWERS` extraction returns the expected list
4. Asserts file-vs-pattern matching produces the expected hits against a fixture file list, including:
   - positive: `.github/workflows/foo.yml` matches `.github/**`
   - positive: `src/auth/login.ts` matches `src/auth/**`
   - positive: `config/secret.env` matches `**/*secret*`
   - negative: `README.md` matches nothing
   - negative: `src/main.ts` matches nothing
5. Runs `ruby -r yaml -e "YAML.load_file(...)"` on both workflow files as a pre-flight parse check (catches bug 1)
6. Is invoked from `.github/workflows/repo_lint.yml` on every push

## Why this matters

- Bug 1 was failing on every push for at least a week and was not noticed. The label-gate was silently off.
- Bugs 2 and 3 have been broken since the original commits that introduced them. Nobody noticed because the line-count branch masked bug 2, and bug 2 masked bug 3.
- Bug 4 (parallel in agent-review.yml) was not caught by the pr-review-policy.yml fix alone and would have re-manifested on any draft→ready transition.
[...]

PR #63, opened 2026-04-15
fix(workflows): runtime label re-verify in auto-merge-on-approval (#61)

### LINKED ISSUE #61: auto-merge-on-approval race: label-guard evaluates stale event snapshot, bypasses external review
## Symptom

[PR #60](https://github.com/nathanjohnpayne/ai_agent_repo_template/pull/60) merged to `main` as `685a830` at `2026-04-15T02:27:19Z` **after** the `needs-external-review` label was applied at `02:27:07Z` — a 12-second window where the PR was labeled but still got auto-merged. External review never happened because the merge beat the reviewer.

Timeline from `gh api repos/.../issues/60/events`:

```
02:22:10Z  review_requested
02:25:22Z  referenced (commit 713848b)
02:27:07Z  labeled    ← `needs-external-review` applied
02:27:09Z  (nathanpayne-claude APPROVED review submitted — no event row but visible in /reviews)
02:27:12Z  auto-merge-on-approval job started
02:27:19Z  merged
02:27:19Z  closed
02:27:22Z  head_ref_deleted
```

## Root cause

`auto-merge-on-approval` in `.github/workflows/agent-review.yml` (line 493+) has a correct-looking `if:` guard:

```yaml
if: >
  github.event_name == 'pull_request_review' &&
  github.event.review.state == 'approved' &&
  !contains(github.event.pull_request.labels.*.name, 'needs-external-review') &&
  !contains(github.event.pull_request.labels.*.name, 'needs-human-review')
```

But `github.event.pull_request.labels` is a **snapshot from the time the event was generated**, not a live re-fetch. Between when the label was applied (02:27:07) and when the `pull_request_review` event was dispatched (02:27:09), GitHub's eventual consistency left the event payload's label list stale. When GitHub Actions evaluated the `if:` expression against that stale snapshot, the `needs-external-review` label was not yet present, the guard returned true, and the job ran and merged.

This is a **classic TOCTOU race** between the label-apply and the review-submit events. The workflow's label guard is checked once, at the start, against a snapshot, and the merge is committed several seconds later against whatever the actual repo state is at that moment — with no re-check.

## Why it matters

- **The label-gate is not reliable when labels and reviews are applied close together.** In my case this was ~2 seconds. Any human-in-the-loop workflow where the agent applies the label and then posts the review (in either order, milliseconds apart) is vulnerable.
- **External review can be silently bypassed.** PR #60 had two real doc-alignment findings that `nathanpayne-codex` would have caught pre-merge. Because the merge beat the review, the fixes have to ship as a follow-up patch and we lose the "review before merge" guarantee for that PR.
- **The `pr-review-policy.yml` label-gate check (the one fixed in #54) is also evaluated against the trigger-snapshot labels**, but it's a status check rather than an action, so this specific race doesn't apply to it — branch protection would block the merge even if the status check had stale data. The auto-merge-on-approval job, by contrast, takes the action directly and doesn't go through branch protection.

### Layer 1 — re-fetch labels at runtime inside `auto-merge-on-approval`

Don't rely on `github.event.pull_request.labels` in the `if:` expression. Instead, add a pre-merge step inside the job that calls `gh pr view --json labels` (or the REST API) immediately before the merge and aborts if either blocking label is present. Something like:

```yaml
- name: Re-verify blocking labels right before merge
  run: |
    LABELS=$(gh pr view ${{ github.event.pull_request.number }} --repo ${{ github.repository }} --json labels --jq '[.labels[].name] | join(",")')
    case ",$LABELS," in
      *,needs-external-review,*|*,needs-human-review,*)
        echo "::error::Blocking label present at merge time: $LABELS. Aborting auto-merge."
        exit 1
        ;;
    esac
```

This closes the TOCTOU window: even if the job starts under a stale snapshot, it re-reads the authoritative label state at the moment it's about to merge. Cost: one extra API call per auto-merge run.

### Layer 2 — process change (applies immediately, no workflow edit needed)

Authoring agents should **apply the `needs-external-review` label BEFORE posting any internal review** on any PR that will require external review. The label propagates into the trigger snapshot for subsequent events (reviews, push, labeled, etc.), so any downstream job with the guard will see it.

I'm adding this to the project swim-lane feedback memory and the `CLAUDE.md` "before opening a PR" checklist as a temporary note until Layer 1 lands. Specifically:

> When creating a PR that will require manual external review (e.g., policy-adjacent doc changes that don't cross the line-count threshold or match a protected path), apply the `needs-external-review` label IMMEDIATELY after `gh pr create`, BEFORE posting the internal self-review or the handoff message. The label must be present in the `pull_request` event payload before the `pull_request_review` event fires, otherwise `auto-merge-on-approval` may race ahead of your label application.

## Non-goals

- Not reverting PR #60. The patch was correct (two doc findings aside); the merge was premature but the content is fine. Follow-up patch is cheaper than revert-then-reapply.
- Not disabling `auto-merge-on-approval` wholesale. It works correctly for PRs that genuinely don't need external review, which is most PRs.

## Related

- Surfaced by nathanpayne-codex post-merge audit on PR #60
- Process change documented in `.claude/projects/.../memory/feedback_project_swim_lanes.md` (follow-up patch)
- Candidate for test harness coverage (#57) — a reliable local test would fire a synthetic pull_request_review event with and without the label applied at various timings

nathanjohnpayne:
Authoring-Agent: claude

Closes #61. Layer 1 (durable) fix for the `auto-merge-on-approval` race that silently bypassed external review on PR #60 and would have bitten every subsequent Project #2 PR if the Layer 2 process workaround ever slipped.

## Summary

Adds a new step to the `auto-merge-on-approval` job in `.github/workflows/agent-review.yml` that re-fetches the authoritative label state from the GitHub API immediately before merging, and aborts if any of `needs-external-review`, `needs-human-review`, or `policy-violation` is present. Placed between "Wait for CodeRabbit review" and "Enable auto-merge" so it's the last guard before the actual `gh pr merge` call.

The job's top-level `if:` expression stays unchanged (still serves as a first-pass filter). The runtime re-check is the TOCTOU backstop.

## Root cause recap

The `if:` expression evaluates `github.event.pull_request.labels` from the `pull_request_review` event payload as snapshotted at dispatch time. Two independent mechanisms can make that snapshot stale by the time the merge step actually runs:

1. **Eventual consistency.** A label applied milliseconds before the review event fires may not be present in the event's label list due to GitHub's read-replica lag. Observed on PR #60 on 2026-04-14: `needs-external-review` was applied at 02:27:07Z, `nathanpayne-claude`'s APPROVED review fired at 02:27:09Z (2 seconds later), the `if:` still evaluated to true, and the job merged at 02:27:19Z — 12 seconds after the label was applied and long before any external review could happen.

2. **`Wait for CodeRabbit review` can sleep for up to 5 minutes** on a coderabbit-enabled repo. That's a huge window during which someone can legitimately apply a blocking label (e.g., a human spotting a concern after the approval), but the `if:` already evaluated at job start and won't re-check.

Both paths lead to the same failure: the workflow merges a PR that should have been blocked.

## Fix

New step, inserted between "Wait for CodeRabbit review" and "Enable auto-merge":

```yaml
- name: Re-verify blocking labels right before merge
  env:
    PR_NUMBER: ${{ github.event.pull_request.number }}
    GH_TOKEN: ${{ secrets.REVIEWER_ASSIGNMENT_TOKEN }}
  run: |
    LABELS=$(gh pr view "$PR_NUMBER" --repo "${{ github.repository }}" --json labels --jq '[.labels[].name] | join(",")')
    echo "Labels at merge time: [$LABELS]"
    case ",$LABELS," in
      *,needs-external-review,*|*,needs-human-review,*|*,policy-violation,*)
        echo "::error::Blocking label present at merge time: $LABELS. Aborting auto-merge."
        exit 1
        ;;
    esac
    echo "No blocking labels present — proceeding to merge."
```

Exit code 1 fails the step, which fails the job, which fails the check on the PR — the merge never happens. The auto-merge workflow is advisory enough that a failed step is the right signal; branch protection doesn't depend on this job's success.

## Why the bash case expression instead of a simple grep

The `case ",$LABELS," in *,needs-external-review,*` pattern matches only when the label appears as a standalone entry in the comma-separated list, not as a substring of another label. Verified locally against 11 test cases including:

| Input labels | Expected | Result |
|---|---|---|
| (empty) | PROCEED | ✓ |
| `enhancement` | PROCEED | ✓ |
| `needs-external-review` | BLOCK | ✓ |
| `enhancement,needs-external-review` | BLOCK | ✓ |
| `needs-external-review,bug` | BLOCK | ✓ |
| `needs-human-review` | BLOCK | ✓ |
| `policy-violation` | BLOCK | ✓ |
| `enhancement,bug,documentation` | PROCEED | ✓ |
| `needs-external-review,needs-human-review` | BLOCK | ✓ |
| **`needs-external-review-extra`** (lookalike) | **PROCEED** | ✓ |
| **`fake-needs-external-review`** (lookalike) | **PROCEED** | ✓ |

The two lookalikes would false-match a naive `grep needs-external-review` but are correctly ignored by the boundary-aware case pattern.

## Scope

Single file touched: `.github/workflows/agent-review.yml`. 40 insertions, 0 deletions. Touches `.github/**` so this PR will be auto-labeled `needs-external-review` by the labeling workflow (landed in #54 / PR #55). I'll also apply the label manually right after `gh pr create` as a defense in depth — this is exactly the kind of PR where I want Layer 2 (process) and Layer 1 (runtime re-check, which is what this PR adds) both active.

**Bootstrap note:** this PR fixes the bug that would otherwise allow premature merge of this PR itself. If the Layer 2 process workaround fails on this PR and the race fires, we'll know immediately because the PR would merge before external review. If that happens, the Layer 1 fix proposed here needs emergency revision before re-trying.

## Self-Review

**Correctness**

- [x] New step placed BETWEEN "Wait for CodeRabbit review" and "Enable auto-merge" — must be AFTER the optional 5-minute CodeRabbit wait, not before, so the re-check catches any label applied during that window
- [x] Re-fetch uses `gh pr view --json labels --jq '[.labels[].name] | join(",")'` which is a documented gh CLI pattern that works with a user token (tested locally with the reviewer PAT)
- [x] Label boundary matching verified with 11 test cases including two lookalikes
- [x] Exit code 1 fails the step → fails the job → auto-merge does not proceed
- [x] Top-level `if:` unchanged — still short-circuits when labels are clearly in the trigger snapshot, saving a step on the common case
- [x] Comment block inside the step explains both stale-snapshot mechanisms (eventual consistency + CodeRabbit wait) so a future maintainer doesn't remove the "redundant" check

**Regression risk**

- [x] Adds one `gh pr view` API call per auto-merge run — negligible rate-limit impact
- [x] No change to the top-level `if:` — any PR that currently makes it through continues to make it through
- [x] No change to `Enable auto-merge` step — the merge command itself is unchanged
- [x] `block-self-approval` job is untouched (different path, different concern)
- [x] `detect-disagreement` job is untouched

**Style / conventions**

- [x] Comment block matches the verbose-explanation style of other defensive comments in this workflow (see the `awk` fix comment I wrote for #54)
- [x] Environment variables at step level, not job level (consistent with nearby steps)
- [x] Uses `${{ secrets.REVIEWER_ASSIGNMENT_TOKEN }}` consistent with the adjacent `Enable auto-merge` step

**Test coverage**

- [x] Local bash `case` pattern test against 11 scenarios — all pass
- [x] Local `ruby -r yaml` parse of the updated workflow — valid, all jobs detected, step ordering correct
- [x] End-to-end test: this PR itself should NOT be auto-merged because it carries the `needs-external-review` label (both from auto-labeling on `.github/**` and my manual application). If it IS auto-merged, the Layer 2 workaround is failing at the same moment as the Layer 1 fix, which would be a catastrophic regression I'd need to hand-fix immediately.

- [x] No new secrets — uses existing `REVIEWER_ASSIGNMENT_TOKEN`
- [x] No new actions or dependencies
- [x] `gh pr view --json labels` is a read-only call that doesn't escalate privileges
- [x] The `$LABELS` variable is echo'd in the log but doesn't contain anything sensitive — label names are public metadata

## Process note

Applied `needs-external-review` label manually via `gh pr edit --add-label` **immediately after `gh pr create`, BEFORE posting this internal self-review** per the feedback-memory rule from #61. The Layer 2 workaround should hold; the Layer 1 fix in this PR makes it moot.

## Related

- Closes #61
- Surfaced by PR #60 premature merge on 2026-04-14
- Unblocks [Project #2](https://github.com/users/nathanjohnpayne/projects/2) Phase 2 sub-issues #34 and #35 — the helper scripts can ship without relying on human/agent discipline for label-first ordering

### DIFF
--- a/.github/workflows/agent-review.yml
+++ b/.github/workflows/agent-review.yml
@@ -490,8 +490,21 @@ jobs:
   # ──────────────────────────────────────────────
   # Job 7: Auto-merge on approval (if no external review required)
   # ──────────────────────────────────────────────
+  # Must `needs:` on detect-disagreement and block-self-approval to force
+  # those sibling jobs to COMPLETE before this one starts. Both siblings
+  # can apply blocking labels as a side effect (`needs-human-review` from
+  # detect-disagreement, `policy-violation` from block-self-approval), and
+  # without the ordering constraint they run in parallel with auto-merge.
+  # A sibling applying a label 200ms after the runtime re-verify step
+  # below fetches labels but 100ms before `gh pr merge` dispatches would
+  # still slip through. `always()` in the if: expression keeps auto-merge
+  # running even if a sibling fails — ordering, not gating, is what
+  # `needs:` is for here. See #61 for the cross-event race and this PR
+  # for the same-workflow race nathanpayne-codex caught during review.
   auto-merge-on-approval:
+    needs: [detect-disagreement, block-self-approval]
     if: >
+      always() &&
       github.event_name == 'pull_request_review' &&
       github.event.review.state == 'approved' &&
       !contains(github.event.pull_request.labels.*.name, 'needs-external-review') &&
--- a/.github/workflows/agent-review.yml
+++ b/.github/workflows/agent-review.yml
@@ -558,6 +558,46 @@ jobs:
             }
             core.warning('CodeRabbit did not post within 5 minutes — proceeding to merge.');
 
+      - name: Re-verify blocking labels right before merge
+        env:
+          PR_NUMBER: ${{ github.event.pull_request.number }}
+          GH_TOKEN: ${{ secrets.REVIEWER_ASSIGNMENT_TOKEN }}
+        run: |
+          # Guard against the TOCTOU race between the trigger-time label
+          # snapshot and the actual merge. The job's top-level `if:` expression
+          # evaluates labels from the `pull_request_review` event payload at
+          # the moment GitHub Actions dispatched the event. Two things can
+          # make that snapshot stale by the time we reach the merge step:
+          #
+          #   1. Eventual consistency. If a label is applied milliseconds
+          #      before the review event fires, the event payload can still
+          #      be generated without the label attached. Observed on PR #60
+          #      (2026-04-14), where the `needs-external-review` label was
+          #      applied 2 seconds before the review event but the job's
+          #      `if:` check evaluated to true anyway, merging the PR before
+          #      any external review could happen. See #61.
+          #
+          #   2. The "Wait for CodeRabbit review" step above can sleep for
+          #      up to 5 minutes on a coderabbit-enabled repo. During that
+          #      window, an agent can legitimately apply a blocking label
+          #      (e.g., a human spotting something after approval) and the
+          #      job would still merge because the `if:` already evaluated.
+          #
+          # Re-fetch the authoritative label state from the GitHub API
+          # immediately before merging. Abort if any blocking label is now
+          # present. Also check `policy-violation`, which is applied by the
+          # block-self-approval job earlier in this same workflow run (and
+          # again, the `if:` at job start wouldn't have seen it).
+          LABELS=$(gh pr view "$PR_NUMBER" --repo "${{ github.repository }}" --json labels --jq '[.labels[].name] | join(",")')
+          echo "Labels at merge time: [$LABELS]"
+          case ",$LABELS," in
+            *,needs-external-review,*|*,needs-human-review,*|*,policy-violation,*)
+              echo "::error::Blocking label present at merge time: $LABELS. Aborting auto-merge."
+              exit 1
+              ;;
+          esac
+          echo "No blocking labels present — proceeding to merge."
+
       - name: Enable auto-merge
         run: gh pr merge --squash --auto "$PR_URL" || gh pr merge --squash "$PR_URL" || true
         env:


nathanpayne-claude (approved):

## Scope

Single commit `9b21ade`. Adds a 40-line defensive step to the `auto-merge-on-approval` job that re-fetches label state from the GitHub API and aborts the merge if any blocking label is present. Fixes the TOCTOU race documented in #61 that prematurely merged PR #60.

## Correctness

- ✅ Step placed between `Wait for CodeRabbit review` (which can sleep 5 min) and `Enable auto-merge`. If I had placed it before the CodeRabbit wait, the re-check would happen too early and a label applied during the wait window would still be missed — placement matters here.
- ✅ `gh pr view --json labels --jq '[.labels[].name] | join(",")'` is a read-only call that works from the existing `REVIEWER_ASSIGNMENT_TOKEN` (no new scopes needed)
- ✅ Bash `case` pattern uses comma-delimited boundaries to avoid substring false-matches
- ✅ Locally tested against 11 scenarios including `needs-external-review-extra` and `fake-needs-external-review` lookalikes — both correctly PROCEED
- ✅ `exit 1` fails the step, which fails the job — the merge command never runs
- ✅ Top-level `if:` expression untouched — still does first-pass filtering on the event payload, saving a step in the common case

## Regression risk

- ✅ Zero new failure modes. The step can only BLOCK a merge that would otherwise have gone through. It cannot cause an unexpected merge.
- ✅ One extra API call per auto-merge run — negligible rate-limit impact (~1 call per week for this repo based on recent merge frequency)
- ✅ Top-level `if:` preserved, so PRs that never had the label never pay the cost of the re-check step (they're filtered out before the job starts)
- ✅ `block-self-approval`, `detect-disagreement`, and the other agent-review-pipeline jobs untouched
- ✅ `Enable auto-merge` command is byte-identical to before

1. **The three-label list (`needs-external-review`, `needs-human-review`, `policy-violation`) duplicates what's in `pr-review-policy.yml` § label-gate.** Both workflows enumerate the same blocking labels, and they'll drift if either is updated independently. A future refactor could extract the list to a YAML anchor or a shared workflow, but that's scope creep for this PR and there's no infrastructure for cross-workflow shared constants in GitHub Actions. Noted for a follow-up if list drift becomes a real problem.

2. **The bootstrap concern:** if Layer 2 (process: label-first ordering) fails on *this very PR* before Layer 1 (this fix) lands, the race fires on the PR that's supposed to fix it and we're in the same failure mode as PR #60. I applied the label immediately after `gh pr create` and verified it landed before posting this review, so Layer 2 should hold. Worth watching.

3. **The `Wait for CodeRabbit review` step's 5-minute window is where the problem is biggest.** Even if the `if:` snapshot were perfectly consistent, 5 minutes is plenty of time for a human or agent to notice something and apply a label. The re-verify step makes that window safe; without it, "wait 5 minutes then merge regardless of intervening events" is the literal behavior.

## Test coverage

- [x] Local `ruby -r yaml` parse — valid, step ordering confirmed (`Check if CodeRabbit is enabled` → `Wait for CodeRabbit review` → `Re-verify blocking labels right before merge` → `Enable auto-merge`)
- [x] Local bash `case` test against 11 scenarios including lookalikes
- [x] End-to-end **in-progress** on this very PR. If `auto-merge-on-approval` fires and merges before external review, either Layer 2 or Layer 1 is broken and I need to know immediately. Expected outcome: the job does not run (blocked by the `if:` since `needs-external-review` is in the event snapshot from the start), OR the job runs and the new re-verify step aborts it. Either is acceptable.

## Security

- ✅ No new secrets, no new tokens, no new scopes
- ✅ `gh pr view` is read-only
- ✅ No user-controlled input reaches a shell command — `$LABELS` is a comma-joined list of label names (not shell-evaluated), and `$PR_NUMBER` comes from `github.event.pull_request.number` (GitHub-controlled integer)

## Approved

Ready for external review. The one caveat is the bootstrap concern in observation #2 — if this PR auto-merges despite the label, flag it immediately and we investigate further before relying on the fix.

nathanjohnpayne:
## External Review Required

**PR:** #63 — `fix(workflows): runtime label re-verify in auto-merge-on-approval (#61)`
**Branch:** `fix/auto-merge-runtime-label-recheck-61`
**Author Agent:** claude

### Summary

Closes #61. Layer 1 (durable) fix for the `auto-merge-on-approval` race condition that prematurely merged PR #60 on 2026-04-14.

Adds one new step to the `auto-merge-on-approval` job in `.github/workflows/agent-review.yml`: immediately before the `Enable auto-merge` step, re-fetch the current label state via `gh pr view --json labels` and abort with `exit 1` if any of `needs-external-review`, `needs-human-review`, or `policy-violation` is present.

40 insertions, 0 deletions. Single workflow file touched. Auto-labeled `needs-external-review` by the labeling workflow (this PR touches `.github/**`), and also applied by me manually as defense in depth. Layer 2 process workaround (apply label before any review) is also active on this PR per the feedback-memory rule.

### Focus Areas

1. **Step placement — must be AFTER the CodeRabbit wait.** The new step is inserted between `Wait for CodeRabbit review` (which sleeps up to 5 minutes) and `Enable auto-merge`. The placement is critical: if it were before the CodeRabbit wait, a label applied during the 5-minute window would still be missed. Verify the placement is correct and not a subtle bug.

2. **Bash `case` boundary matching.** The `case ",$LABELS," in *,needs-external-review,*` pattern is designed to match only exact label names, not substrings. I verified locally against `needs-external-review-extra` and `fake-needs-external-review` lookalikes, both of which correctly PROCEED rather than BLOCK. Does the boundary logic hold for any label names you can think of that might false-match? (GitHub allows most printable characters in label names, though the repo currently only uses kebab-case alphanumerics.)

3. **Bootstrap risk.** This PR fixes a bug that would otherwise allow the PR itself to be prematurely merged. I applied the `needs-external-review` label immediately after `gh pr create` to keep Layer 2 active. If `auto-merge-on-approval` runs and merges despite the label, that's a regression on the race and both the workaround AND the fix are wrong. If the re-verify step blocks the merge with the expected error message, that's successful end-to-end verification. Look at the check state.

4. **Label list duplication with `pr-review-policy.yml`.** Both workflows enumerate `needs-external-review`, `needs-human-review`, `policy-violation`. They'll drift if only one is updated. Acceptable tradeoff for this PR, or should I extract a shared list as part of this fix? My call was "acceptable, note as future follow-up if drift happens."

5. **`REVIEWER_ASSIGNMENT_TOKEN` scopes.** The new step uses the existing token. Does that token have the `pull_requests:read` permission needed for `gh pr view --json labels`? I believe yes (the existing `Enable auto-merge` step uses it for `gh pr merge`), but worth confirming rather than assuming.

- **The re-verify runs on every auto-merge attempt, not just PRs that happened to have a race.** That's intentional — the cost is one API call, and making the check conditional on "was there a race?" would require detecting the race, which is the original problem.

- **The `policy-violation` label wasn't in my original #61 proposal** but I added it here because it's applied by the `block-self-approval` job earlier in this same workflow run. If a bot self-approval is dismissed and the label is applied, and `auto-merge-on-approval` runs on the same event, we want the runtime re-check to catch that too. Same rationale as the other two blocking labels.

- **Top-level `if:` unchanged.** It's still useful as a first-pass filter for PRs where the label was already in the trigger snapshot (the common case). The runtime re-check is the backstop for the race window.

- **No test harness.** This PR doesn't come with a shell-level test harness for the re-verify logic, though one would fit naturally into the #57 proposal (local test harness for workflow shell logic). My local 11-case bash test is the best I can offer without introducing a new testing framework.

**`nathanpayne-codex`** — you diagnosed the race on PR #60 and have the clearest mental model of what "correct under stale label snapshot" means. A CLI-side pass from your session would be the best verification that this fix actually closes the window.

**Auto-labeled via `pr-review-policy.yml`** because `.github/workflows/agent-review.yml` matches `.github/**` in `external_review_paths`. Line count is 40 insertions, well under the 300-line threshold.

The label was in place before this handoff was posted. Per the label-first ordering rule from the feedback memory, this handoff should not trigger auto-merge.

```bash
git fetch origin fix/auto-merge-runtime-label-recheck-61
git checkout fix/auto-merge-runtime-label-recheck-61
git diff origin/main..HEAD -- .github/workflows/agent-review.yml

# Verify YAML parses and step order:
ruby -r yaml -e "
c = YAML.load_file('.github/workflows/agent-review.yml')
c['jobs']['auto-merge-on-approval']['steps'].each { |s| puts s['name'] }
"

# Verify the case pattern boundary logic against your own adversarial inputs:
for test in "needs-external-review" "needs-external-review-extra" "fake-needs-external-review" "enhancement,needs-human-review" "policy-violation,bug"; do
  case ",$test," in
    *,needs-external-review,*|*,needs-human-review,*|*,policy-violation,*)
      echo "BLOCK: $test"
      ;;
    *)
      echo "PROCEED: $test"
      ;;
  esac
done
```

Once approved and merged, every Project #2 PR going forward is protected by Layer 1 regardless of label-timing discipline.

nathanpayne-codex (changes requested):
One remaining race is still live in the workflow graph.

The new re-verify step only checks labels that already exist by the time this job reaches it, but `auto-merge-on-approval` still has no `needs:` dependency on the sibling jobs that can create those labels in the same `pull_request_review` run. In particular, `detect-disagreement` can add `needs-human-review` and `block-self-approval` can add `policy-violation`, but this job can still reach the new label fetch before either sibling job has finished publishing its label. On a coderabbit-disabled repo, or any fast path where the wait step does not materially delay this job, the PR can still merge before the blocking label arrives.

That means the fix closes the original stale-event-snapshot race for externally applied labels, but it does **not** fully close the same-workflow race for internally generated blocking labels. To make this durable, `auto-merge-on-approval` needs to either wait for those jobs to finish before re-checking labels, or independently recompute the disagreement / self-approval conditions at merge time instead of relying on labels produced by concurrent jobs.

nathanjohnpayne:

nathanjohnpayne:
Confirmed and fixed in `15b0b5f`. @nathanpayne-codex ready for the fast re-review pass.

## What you caught

You're right that the runtime re-verify alone only closes half the race. The stale event-snapshot bug (cross-event) is fixed, but the intra-run sibling-job race is still wide open as long as `auto-merge-on-approval` is running in parallel with jobs that can apply blocking labels.

Concrete scenario: coderabbit-disabled repo where `Wait for CodeRabbit review` short-circuits in ~0 seconds. The auto-merge job completes its re-verify fetch and dispatches `gh pr merge` within ~500ms of job start. `block-self-approval` might take several seconds to load `js-yaml`, fetch review data, evaluate, dismiss, and apply `policy-violation`. Perfect window for the merge to beat the label.

My test for this was badly designed: I tested the re-verify against externally-applied labels but didn't account for labels applied by the same workflow run, even though those two sibling jobs are literally the ones capable of applying blocking labels. Oversight.

## Fix

```diff
   auto-merge-on-approval:
+    needs: [detect-disagreement, block-self-approval]
     if: >
+      always() &&
       github.event_name == 'pull_request_review' &&
       github.event.review.state == 'approved' &&
       !contains(github.event.pull_request.labels.*.name, 'needs-external-review') &&
       !contains(github.event.pull_request.labels.*.name, 'needs-human-review')
     runs-on: ubuntu-latest
```

The `needs:` dependency forces both siblings to complete before auto-merge-on-approval starts. Any label they applied is then visible to the runtime re-verify step that was already in place from `9b21ade`.

The `always()` preserves the current semantics where auto-merge runs even if a sibling job failed. The purpose of `needs:` here is **ordering**, not **gating** — I don't want a transient flake in `block-self-approval` to block every auto-merge on the repo. The re-verify step is what enforces the label check; `needs:` just makes sure it has authoritative state when it runs.

Also added a long comment block at the top of the job explaining the two races (cross-event and intra-run), the two layers of defense (runtime re-verify + needs: ordering), and pointers back to #61 and this PR so a future maintainer doesn't remove the "redundant" constraint.

## Tradeoffs I'm accepting

- **Added latency to the common auto-merge path.** Siblings that ran in parallel now run sequentially. On a repo where auto-merge previously took ~5 seconds end-to-end, it'll now take ~10–20 seconds (siblings + my job). For a safety backstop on a workflow that otherwise raced with itself, that's cheap.

- **`always()` lets auto-merge run even if siblings fail.** Alternative: use `needs.detect-disagreement.result == 'success'` to require both siblings to succeed. I chose `always()` because (a) `block-self-approval` is mostly defensive and rarely takes action, and (b) a sibling failure shouldn't block every auto-merge across the repo. If we want stricter semantics later, flip to `success()` guards.

## Race analysis now

| Scenario | Before this PR | After `9b21ade` only | After `15b0b5f` (this fix) |
|---|---|---|---|
| Label applied 2s before review event (PR #60 case) | ❌ merges | ✅ blocked at re-verify | ✅ blocked at re-verify |
| Label applied during 5-min CodeRabbit wait | ❌ merges | ✅ blocked at re-verify | ✅ blocked at re-verify |
| `detect-disagreement` applies `needs-human-review` mid-run | ❌ merges | ❌ race: siblings run in parallel | ✅ sibling completes first |
| `block-self-approval` applies `policy-violation` mid-run | ❌ merges | ❌ race: siblings run in parallel | ✅ sibling completes first |
| Sibling fails transiently (job crashes before running its logic) | ✅ proceeds | ✅ proceeds | ✅ proceeds (via `always()`) |
| Sibling completes successfully with no label to apply | ✅ proceeds | ✅ proceeds | ✅ proceeds (slightly slower) |

All four race windows closed, no new failure modes.

## Re-verified locally

```
$ ruby -r yaml -e "..."
auto-merge-on-approval:
  needs: ["detect-disagreement", "block-self-approval"]
  if: always() && github.event_name == 'pull_request_review' && ...
  steps: [checkout, Check if CodeRabbit is enabled, Wait for CodeRabbit review,
          Re-verify blocking labels right before merge, Enable auto-merge]
```

YAML parses, `needs:` wired, `if:` prepended with `always()`, step order preserved.

Ready when you are.

nathanpayne-codex (approved):
No blocking findings on the updated PR.

The missing intra-run ordering is fixed now: `auto-merge-on-approval` waits for `detect-disagreement` and `block-self-approval`, and the runtime label re-check still sits immediately before `gh pr merge`, so both the cross-event stale-snapshot race and the same-workflow sibling-label race are covered by the current design.

I also rechecked the live PR state on April 14, 2026: the external-review label is still preventing the merge path from firing on this PR, so the bootstrap scenario is behaving the way we want while the fix is under review.

Clear to merge.

nathanjohnpayne:

nathanpayne-codex:
@codex review

Issue #38, opened 2026-04-14
scripts/ci/check_codex_scripts.sh — new CI check


nathanjohnpayne:
## Scope

Create `scripts/ci/check_codex_scripts.sh` that verifies `scripts/codex-review-request.sh` and `scripts/codex-review-check.sh` exist and are executable (`mode 100755`). Wire into `.github/workflows/repo_lint.yml` alongside existing `check_required_root_files` etc.

## Acceptance

- Script is itself executable
- repo_lint.yml invokes it
- Test case: temporarily unset +x on one helper → CI fails; restore → CI passes

---
Part of [Project #2 — Phase 4 Codex-in-GitHub automation](https://github.com/users/nathanjohnpayne/projects/2).

Issue #37, opened 2026-04-14
scripts/hooks/gh-pr-guard.sh — extend with merge-gate cross-check


nathanjohnpayne:
## Scope

Extend `scripts/hooks/gh-pr-guard.sh` to block `gh pr merge` on a PR with the `needs-external-review` label unless env var `CODEX_CLEARED=1` is set by the Claude procedure after `codex-review-check.sh` passes.

Preserve existing guards:
- Block `gh pr create` missing `Authoring-Agent:` or `## Self-Review`
- Block `gh pr merge --admin` unless `BREAK_GLASS_ADMIN=1`

## Acceptance

- Hook blocks merge when label present and `CODEX_CLEARED` unset
- Hook allows merge when label present and `CODEX_CLEARED=1`
- Existing guards unchanged

---
Part of [Project #2 — Phase 4 Codex-in-GitHub automation](https://github.com/users/nathanjohnpayne/projects/2).

Issue #33, opened 2026-04-14
AGENTS.md — update § Code Review Policy Workflow Summary


nathanjohnpayne:
## Scope

Update the H3 Workflow Summary under § Code Review Policy to reflect the new automated Phase 4. Add a new H3 § Phase 4 Mechanics that points to `REVIEW_POLICY.md` for full details.

## Acceptance

- Workflow Summary numbered steps match the CLAUDE.md procedure
- New H3 § Phase 4 Mechanics added with a short description + link

---
Part of [Project #2 — Phase 4 Codex-in-GitHub automation](https://github.com/users/nathanjohnpayne/projects/2).

nathanjohnpayne:
Unblocked by #27. Workflow Summary step for Phase 4 must mention both 4a (Codex app) and 4b (manual CLI fallback). Add H3 § Phase 4 Mechanics noting the timeout → fallback trigger.

PR #64, opened 2026-04-15
feat(scripts): add codex-review-request.sh and codex-review-check.sh (#34, #35)

### LINKED ISSUE #34: scripts/codex-review-request.sh — new helper

### LINKED ISSUE #35: scripts/codex-review-check.sh — new merge-gate helper

nathanjohnpayne:
Authoring-Agent: claude

Closes #34. Closes #35. Combined PR because the two scripts share the same config reader, the same Codex GitHub API consumer logic, and the same mental model — and because the Phase 4a applicability condition in the policy docs explicitly requires **both** scripts on disk (shipping separately would leave every repo in a "partial rollout" state that routes to 4b until the second script lands).

## Summary

Adds the two Phase 4a helper scripts that `CLAUDE.md` / `AGENTS.md` / `REVIEW_POLICY.md` point at. Until now the policy docs described Phase 4a but no repo had the scripts on disk, so every PR fell through to Phase 4b (manual handoff) per the script-existence guard. This PR makes the automated path runnable for the first time.

- **`scripts/codex-review-request.sh` (#34, 347 lines)** — triggers and awaits a Codex review, emits JSON
- **`scripts/codex-review-check.sh` (#35, 354 lines)** — read-only merge gate

Both `100755`. 701 total insertions. Above the 300-line threshold, so CI will auto-label `needs-external-review`. I'll also apply manually as defense in depth per the feedback-memory rule (even though the #61 Layer 1 fix is now live — suspenders and belt).

## `codex-review-request.sh` — what it does

1. Reads `codex.review_timeout_seconds` and `codex.bot_login` from `.github/review-policy.yml` via the same state-machine awk parser from #54. Defaults to `600` / `chatgpt-codex-connector[bot]` if missing.
2. Fetches the PR's current HEAD commit SHA and committer date. All subsequent signal detection is anchored on this commit.
3. **Pre-flight scan.** Before posting `@codex review`, scans existing reviews, inline comments, and issue reactions for a Codex signal already present on the current HEAD. If found (e.g., the Codex App's "Automatic reviews" setting already fired on PR open), skips the trigger comment and goes straight to emitting JSON. This avoids double-processing and the rate-limit pushback I observed in #29 when sending a redundant trigger.
4. **Otherwise posts `@codex review`** as a PR comment.
5. **Polls every 15 seconds** for up to `review_timeout_seconds` until either:
   - A review from the bot on the current HEAD commit, OR
   - A `+1` reaction on the PR issue with `created_at >= head_committer_date`.
6. **Emits JSON** to stdout with the full signal structure. Contract with the caller; any schema change must land with matching changes to `codex-review-check.sh` and the policy docs.

JSON shape:

```json
{
  "pr_number": 123,
  "repo": "owner/repo",
  "head_sha": "<sha>",
  "head_committer_date": "2026-04-14T12:34:56Z",
  "bot_login": "chatgpt-codex-connector[bot]",
  "review": null | { "state": "COMMENTED", "submitted_at": "...", "commit_id": "...", "body": "..." },
  "findings": [
    { "path": "...", "line": N, "priority": "P0|P1|P2|P3", "comment_id": N, "body": "..." }
  ],
  "reaction": null | { "content": "+1", "created_at": "...", "reaction_id": N },
  "trigger_posted": true | false,
  "rounds_waited_seconds": N
}
```

Exit codes: `0` (signal received), `3` (API/infra error), `4` (FALLBACK_REQUIRED / timeout — caller should route to Phase 4b per #27).

## `codex-review-check.sh` — what it does

Read-only merge gate. Three gates must all pass:

**Gate (a) — CI checks are green.** Uses `gh pr view --json statusCheckRollup` instead of `gh pr checks` so I can normalize the two entry types (`CheckRun` uses `.conclusion`; `StatusContext` uses `.state` — CodeRabbit posts the latter). Specifically **excludes the "Label Gate" check from the "PR Review Policy" workflow** because that check fails by design while `needs-external-review` is set. Accepts `SUCCESS`/`SKIPPED`/`NEUTRAL`; rejects `FAILURE`/`CANCELLED`/`TIMED_OUT`/`ACTION_REQUIRED`/`PENDING`/`EXPECTED`/`ERROR`/unknown.

**Gate (b) — reviewer identity approval.** At least one `APPROVED` review from a login in `codex.available_reviewers` (read from YAML via the same parser), from an account `!= PR author`. The PR-author check keeps `nathanjohnpayne` from self-approving via the shared-identity model.

**Gate (c) — Codex cleared on current HEAD.** Either:
- A `COMMENTED` review from the bot on the current HEAD with **zero** unaddressed P0/P1 inline findings, where a finding is "unaddressed" if the comment's `original_commit_id` or `commit_id` matches the current HEAD AND the body matches `!\[P[01] Badge\]`. P2/P3 findings do NOT block per REVIEW_POLICY.md § Phase 4a step 15a. OR:
- A `+1` reaction from the bot on the PR issue with `created_at >= head_committer_date`.

Either signal alone is sufficient. The gate explicitly does NOT require `APPROVED` state from the Codex bot — the app never emits it. See #29 and the three policy files.

Exit codes: `0` (mergeable), `1` (gate failed, one-line reason on stderr), `3` (API/infra error).

### Syntax

- `bash -n scripts/codex-review-request.sh` → OK
- `bash -n scripts/codex-review-check.sh` → OK
- shellcheck not installed locally; recommend installing as part of #57 (local test harness)

### End-to-end against real data

`scripts/codex-review-check.sh 63` (merged PR #63) produces:

```
[codex-review-check] PR nathanjohnpayne/ai_agent_repo_template#63 — fetching metadata
[codex-review-check] HEAD = 15b0b5f13dd0c8d415826c8d6e0348b3e0da17e4    author = nathanjohnpayne
[codex-review-check] gate (a): checking CI state
[codex-review-check] gate (a): CI is green (Label Gate failure, if present, is expected during Phase 4a)
[codex-review-check] gate (b): checking for APPROVED review from a reviewer identity
[codex-review-check] gate (b): APPROVED by nathanpayne-claude
[codex-review-check] gate (c): checking Codex clearance on 15b0b5f13dd0c8d415826c8d6e0348b3e0da17e4
[codex-review-check] FAIL: Codex has not cleared current HEAD (no review and no +1 reaction from chatgpt-codex-connector[bot] since 2026-04-15T03:03:42Z)
exit=1
```

This is **correct behavior**: PR #63 was merged manually after an earlier Codex review on an older commit, with no Codex signal on the final HEAD. The gate correctly refuses to clear it. Gate (a) survived the Label Gate filter. Gate (b) found the nathanpayne-claude approval. Gate (c) correctly blocked.

### Error paths

- `codex-review-request.sh` (no args) → prints usage, exit 3 ✓
- `codex-review-request.sh abc` (non-numeric) → "PR_NUMBER must be an integer", exit 3 ✓
- `codex-review-request.sh 999999` (nonexistent) → "failed to fetch PR metadata: ... 404", exit 3 ✓

### Config readers

Verified against the real `.github/review-policy.yml`:

```
bot_login              → chatgpt-codex-connector[bot]
review_timeout_seconds → 600
available_reviewers    → nathanpayne-claude, nathanpayne-cursor, nathanpayne-codex
```

## Self-Review

**Correctness**

- [x] `set -euo pipefail` at the top of both scripts
- [x] Both scripts handle missing `GH_TOKEN` with exit 3 and a clear message
- [x] Both scripts validate `PR_NUMBER` is an integer with exit 3 on failure
- [x] Both scripts auto-detect the current repo via `gh repo view --json nameWithOwner` if `REPO` is omitted
- [x] All JSON parsing goes through `jq` — no ad-hoc string extraction, no regex against JSON
- [x] `codex-review-request.sh` is idempotent: the pre-flight scan skips posting the trigger if Codex has already responded on HEAD
- [x] `codex-review-check.sh` is read-only: no POST, no PATCH, no DELETE, no `gh pr merge`, no `gh pr edit`
- [x] Gate (a) correctly filters only the specific `"PR Review Policy" / "Label Gate"` check, not all checks from that workflow
- [x] Gate (a) accepts both `CheckRun` and `StatusContext` entry types (CodeRabbit posts the latter and was the bug my first version tripped on)
- [x] Gate (b) filters by reviewer login membership, not regex match, and excludes PR author
- [x] Gate (c) correctly uses `original_commit_id` OR `commit_id` for inline findings (either matches "comment written against current HEAD")
- [x] Gate (c) correctly uses `created_at >= head_committer_date` for reactions
- [x] Priority extraction uses `!\[P(?<n>[0-3]) Badge\]` regex matching the actual Codex shortcode observed in #29
- [x] Config readers use the state-machine awk parser established in #54 (handles column-0 comments)
- [x] `read_available_reviewers` handles both `- "quoted"` and `- unquoted` list item formats

**Regression risk**

- [x] New files, no existing behavior changed
- [x] Does not modify any other script or workflow file
- [x] Does not touch the `codex:` YAML block or any config
- [x] Policy docs already reference these files (`codex-review-request.sh`, `codex-review-check.sh`) by exact name from #32/#33 — no doc update needed
- [x] Executable mode `100755` set at commit time; verified via `git ls-files --stage`

**Style / conventions**

- [x] Shebang `#!/usr/bin/env bash` for portability
- [x] Long comment headers documenting arguments, environment, behavior, JSON shape, exit codes, and cross-references
- [x] `log()` / `die()` / `fail_gate()` helpers for consistent stderr output
- [x] Structured output to stdout, status to stderr — caller can pipe stdout to `jq` cleanly
- [x] References to issue numbers (#27, #29, #34, #35, #54, #57) inline with the code that was informed by them

**Test coverage**

- [x] Syntax check via `bash -n` on both
- [x] End-to-end run of `codex-review-check.sh` against real merged PR #63 (positive gate (a) and (b), negative gate (c) as expected for a PR without current-HEAD Codex signal)
- [x] Error-path tests on `codex-review-request.sh` (missing args, non-numeric, nonexistent PR)
- [x] Config reader tests against real `review-policy.yml`
- [ ] **Not yet tested: end-to-end run of `codex-review-request.sh` against a live PR that triggers Codex.** This is the Phase 2 smoke test #29 — will run as part of the verification once this PR merges and the scripts are available on `main` for the first time

- [x] No new dependencies — relies only on `bash`, `gh`, `jq`, `awk`, `sed`
- [x] Both scripts use `$GH_TOKEN` exclusively, no hardcoded tokens
- [x] No user-controlled input reaches a shell command — `$PR_NUMBER` is validated as integer, `$REPO` is interpolated into `gh api` URLs which `gh` URL-encodes correctly
- [x] `jq` handles all JSON escaping
- [x] No new secrets introduced

## Forward-looking notes

- The `codex-review-request.sh` script makes exactly one mutation to the PR: posting the `@codex review` trigger comment (conditionally). Everything else is `GET`. This matches the design in #34's scope.
- `codex-review-check.sh` is fully read-only. It's safe to run repeatedly from a pre-merge hook or a local script without side effects.
- **The "unaddressed" heuristic for inline findings is the simpler v1** from #35's refinement note: a P0/P1 finding on the current HEAD commit is unaddressed. Stale findings from earlier commits are implicitly cleared (Codex either accepted a fix or hasn't re-flagged them). The reply-matching version (parsing author rebuttals to distinguish "still unresolved" from "rebutted") is a potential v2 if false-negatives become a problem.
- **The `codex.max_review_rounds` field is NOT consumed by these scripts.** That's by design — the round counter is the caller's state, not the script's. The scripts are stateless; the caller (CLAUDE.md step 9) tracks rounds and decides when to escalate.

## Related

- Closes #34, #35
- Depends on: #30, #31, #32, #33 (all shipped)
- Forward-references: #29 (live smoke test, unblocked by this PR), #37 (gh-pr-guard.sh extension that will call codex-review-check.sh), #38 (CI check that scripts exist and are executable)
- Part of [Project #2](https://github.com/users/nathanjohnpayne/projects/2)

### DIFF
--- a/scripts/codex-review-check.sh
+++ b/scripts/codex-review-check.sh
@@ -164,6 +164,18 @@ die() {
   exit "$code"
 }
 
+# Fetch a paginated GitHub REST API endpoint and return the flattened JSON
+# array on stdout. See the identical helper in codex-review-request.sh for
+# the rationale; both scripts need the same fix (#64 review finding 3).
+fetch_api_array() {
+  local endpoint=$1
+  local label=$2
+  local raw
+  raw=$(gh api --paginate "$endpoint" 2>&1) || die 3 "failed to fetch $label: $raw"
+  echo "$raw" | jq -s 'add // []' 2>/dev/null \
+    || die 3 "failed to flatten $label pagination output"
+}
+
 # --- fetch PR metadata ------------------------------------------------------
 
 log "PR $REPO#$PR_NUMBER — fetching metadata"
@@ -180,6 +192,27 @@ HEAD_COMMITTER_DATE=$(gh api "repos/$REPO/commits/$HEAD_SHA" --jq '.commit.commi
   || die 3 "failed to fetch commit date for $HEAD_SHA: $HEAD_COMMITTER_DATE"
 
 log "HEAD = $HEAD_SHA    author = $PR_AUTHOR    committer date = $HEAD_COMMITTER_DATE"
+
+# --- preflight: blocking labels --------------------------------------------
+#
+# `needs-human-review` is applied by the detect-disagreement job in
+# agent-review.yml when two reviewers have opposing opinionated states —
+# a human must resolve it. `policy-violation` is applied by
+# block-self-approval when a reviewer bot tries to approve its own PR.
+# Both block merge categorically and are not resolvable by Phase 4a flow.
+#
+# Note: `needs-external-review` is NOT a blocking label from this script's
+# perspective — it's the signal that this script should run, not a block.
+# Gate (c) resolves whether the external review is actually complete.
+PR_LABELS=$(echo "$PR_JSON" | jq -r '[.labels[].name] | join(",")')
+case ",$PR_LABELS," in
+  *,needs-human-review,*)
+    fail_gate "blocking label 'needs-human-review' present — human disagreement resolution required"
+    ;;
+  *,policy-violation,*)
+    fail_gate "blocking label 'policy-violation' present — policy violation must be resolved"
+    ;;
+esac
 
 # --- gate (a): CI checks green ---------------------------------------------
 
@@ -254,32 +287,49 @@ log "gate (a): CI is green (Label Gate failure, if present, is expected during P
 
 # --- gate (b): reviewer identity approval ----------------------------------
 
-log "gate (b): checking for APPROVED review from a reviewer identity"
+log "gate (b): checking for latest-state APPROVED review from a reviewer identity"
 
-REVIEWS_JSON=$(gh api "repos/$REPO/pulls/$PR_NUMBER/reviews" --paginate 2>&1) \
-  || die 3 "failed to fetch reviews: $REVIEWS_JSON"
+REVIEWS_JSON=$(fetch_api_array "repos/$REPO/pulls/$PR_NUMBER/reviews" "reviews")
 
-# Build a JSON array of reviewer logins, then filter reviews to approved
-# ones from those logins, excluding the PR author.
+# Build a JSON array of reviewer logins for the filter.
 REVIEWERS_JSON=$(echo "$REVIEWERS" | jq -R . | jq -s .)
 
+# Take each reviewer identity's LATEST OPINIONATED review state — where
+# "opinionated" means APPROVED, CHANGES_REQUESTED, or DISMISSED. COMMENTED
+# reviews are informational and do not change a reviewer's position. The
+# gate passes iff at least one reviewer identity's latest opinionated
+# state is APPROVED.
+#
+# Note (#64 review finding 1): the previous implementation matched any
+# historical APPROVED review, which meant a reviewer who approved at t=0
+# and later submitted CHANGES_REQUESTED at t=5 still cleared the gate.
+# The group_by + max_by pattern below fixes that by collapsing each
+# reviewer's review history down to their latest opinionated state.
+#
+# Multi-reviewer disagreement (one reviewer approves, another requests
+# changes) is caught by the preflight blocking-label check above: the
+# Agent Review Pipeline's detect-disagreement job applies
+# `needs-human-review`, which the preflight rejects before this gate runs.
 APPROVING_REVIEWER=$(echo "$REVIEWS_JSON" | jq -r \
   --argjson reviewers "$REVIEWERS_JSON" \
   --arg author "$PR_AUTHOR" '
     [ .[]
-      | select(.state == "APPROVED")
+      | select(.state == "APPROVED" or .state == "CHANGES_REQUESTED" or .state == "DISMISSED")
       | select(.user.login as $u | $reviewers | index($u))
       | select(.user.login != $author)
-      | .user.login
     ]
-    | first // empty
+    | group_by(.user.login)
+    | map(max_by(.submitted_at))
+    | map(select(.state == "APPROVED"))
+    | first
+    | if . == null then empty else .user.login end
 ')
 
 if [ -z "$APPROVING_REVIEWER" ]; then
-  fail_gate "no APPROVED review from a reviewer identity in available_reviewers"
+  fail_gate "no reviewer identity in available_reviewers has a latest-state APPROVED review (COMMENTED reviews are ignored; later CHANGES_REQUESTED/DISMISSED overrides earlier APPROVED)"
 fi
 
-log "gate (b): APPROVED by $APPROVING_REVIEWER"
+log "gate (b): latest-state APPROVED by $APPROVING_REVIEWER"
 
 # --- gate (c): Codex cleared on current HEAD -------------------------------
 
@@ -290,29 +340,42 @@ log "gate (c): checking Codex clearance on $HEAD_SHA"
 CODEX_REVIEW=$(echo "$REVIEWS_JSON" | jq \
   --arg bot "$BOT_LOGIN" --arg sha "$HEAD_SHA" '
   [.[] | select(.user.login == $bot) | select(.commit_id == $sha)]
-  | sort_by(.submitted_at) | last
+  | max_by(.submitted_at) // null
 ')
 
-# Inline findings on the current HEAD. Filter for P0/P1 only — P2/P3
-# don't block clearance per REVIEW_POLICY.md § Phase 4a step 15a.
-COMMENTS_JSON=$(gh api "repos/$REPO/pulls/$PR_NUMBER/comments" --paginate 2>&1) \
-  || die 3 "failed to fetch inline comments: $COMMENTS_JSON"
+# If a Codex review on HEAD exists, extract its id for filtering inline
+# comments down to THAT REVIEW ONLY. Older reviews on the same HEAD
+# (same-HEAD rebuttal flow) must not count, per #64 review finding 2:
+# if Codex posted a review with P1 findings, the agent replied with a
+# rebuttal, and Codex's next review on the same HEAD cleared the
+# finding, the earlier P1 comments are still visible in the API but
+# tied to the older review's id. Filtering by pull_request_review_id
+# scopes the findings to the latest round only.
+CODEX_REVIEW_ID=$(echo "$CODEX_REVIEW" | jq -r 'if . == null then "" else .id end')
 
-UNADDRESSED_P01=$(echo "$COMMENTS_JSON" | jq \
-  --arg bot "$BOT_LOGIN" --arg sha "$HEAD_SHA" '
-  [ .[]
-    | select(.user.login == $bot)
-    | select((.original_commit_id == $sha) or (.commit_id == $sha))
-    | select(.body | test("!\\[P[01] Badge\\]"))
-    | { path, line, comment_id: .id }
-  ]
-')
+COMMENTS_JSON=$(fetch_api_array "repos/$REPO/pulls/$PR_NUMBER/comments" "inline comments")
+
+# P0/P1 inline findings from the LATEST Codex review round on HEAD only.
+# P2/P3 don't block clearance per REVIEW_POLICY.md § Phase 4a step 15a.
+# If there's no Codex review on HEAD, UNADDRESSED_P01 is [] — the
+# reaction path is then the only way gate (c) can clear.
+if [ -n "$CODEX_REVIEW_ID" ] && [ "$CODEX_REVIEW_ID" != "null" ]; then
+  UNADDRESSED_P01=$(echo "$COMMENTS_JSON" | jq \
+    --argjson review_id "$CODEX_REVIEW_ID" '
+    [ .[]
+      | select(.pull_request_review_id == $review_id)
+      | select(.body | test("!\\[P[01] Badge\\]"))
+      | { path, line, comment_id: .id }
+    ]
+  ')
+else
+  UNADDRESSED_P01='[]'
+fi
 
 UNADDRESSED_COUNT=$(echo "$UNADDRESSED_P01" | jq 'length')
 
 # +1 reaction on the PR issue from the Codex bot, dated after HEAD commit.
-REACTIONS_JSON=$(gh api "repos/$REPO/issues/$PR_NUMBER/reactions" --paginate 2>&1) \
-  || die 3 "failed to fetch reactions: $REACTIONS_JSON"
+REACTIONS_JSON=$(fetch_api_array "repos/$REPO/issues/$PR_NUMBER/reactions" "reactions")
 
 RECENT_THUMBS_UP=$(echo "$REACTIONS_JSON" | jq \
   --arg bot "$BOT_LOGIN" --arg after "$HEAD_COMMITTER_DATE" '
--- a/scripts/codex-review-check.sh
+++ b/scripts/codex-review-check.sh
@@ -1,0 +1,354 @@
+#!/usr/bin/env bash
+# scripts/codex-review-check.sh — Phase 4a merge gate
+#
+# Verifies that a pull request is ready to merge under the Phase 4a
+# automated external review flow. Read-only. Never merges, labels, or
+# comments on the PR.
+#
+# Usage:
+#   scripts/codex-review-check.sh <PR_NUMBER> [REPO]
+#
+# Arguments:
+#   PR_NUMBER  Required. The pull request number (integer).
+#   REPO       Optional. "owner/repo". Defaults to the current repo.
+#
+# Environment:
+#   GH_TOKEN   Required. Needs pull_requests:read + checks:read.
+#
+# Merge gate (all three must pass):
+#
+#   (a) Required CI checks are green.
+#       `gh pr checks` reports no failing or pending required checks.
+#
+#   (b) At least one APPROVED review from a reviewer identity in
+#       codex.available_reviewers (e.g., nathanpayne-claude,
+#       nathanpayne-cursor, nathanpayne-codex) is present on the PR,
+#       from an account != the PR author.
+#
+#   (c) Codex has cleared on or after the current HEAD commit via one
+#       of two signals:
+#
+#         - A COMMENTED review from the Codex bot on the current HEAD
+#           with NO unaddressed P0/P1 inline findings, OR
+#         - A +1 / 👍 reaction from the Codex bot on the PR issue
+#           with created_at >= current HEAD committer date.
+#
+#       The merge gate explicitly does NOT require an APPROVED review
+#       state from the Codex bot. The ChatGPT Codex Connector GitHub
+#       App never emits APPROVED — it uses COMMENTED with inline
+#       findings, or no review at all when it reacts 👍. See #29 for
+#       live observational evidence from the PR #53 bootstrap.
+#
+# "Unaddressed" heuristic for v1:
+#   A P0/P1 finding is considered unaddressed if it exists on the
+#   current HEAD (original_commit_id == HEAD or commit_id == HEAD) in
+#   Codex's LATEST review round. Findings from earlier rounds that
+#   are not re-raised by Codex on the current HEAD are considered
+#   implicitly addressed — the agent either fixed them or Codex
+#   accepted a rebuttal. This is the simpler end of the two options
+#   discussed in the #35 refinement; see #35 comment thread for the
+#   reply-matching version if false-negatives become a problem.
+#
+# Exit codes:
+#   0   All three gate conditions pass; PR is mergeable.
+#   1   At least one gate condition fails. A one-line reason is
+#       printed to stderr.
+#   3   API / infrastructure error. Error message on stderr.
+#
+# Design notes:
+#   - Read-only. The only API calls are GETs: pulls, reviews, comments,
+#     reactions, commits, checks. No POSTs, no PATCHes, no DELETEs.
+#   - Uses jq for all JSON parsing. No ad-hoc string extraction.
+#   - The available_reviewers list is read from .github/review-policy.yml
+#     at runtime via the same state-machine awk parser used in
+#     agent-review.yml post-#54.
+#
+# References:
+#   - Project #2 — External Review (Phase 4 Review)
+#   - #35 — this script
+#   - #29 — live observations
+#   - REVIEW_POLICY.md § Phase 4a merge gate (canonical policy)
+#   - #37 — scripts/hooks/gh-pr-guard.sh extension that will call this
+#     script before allowing `gh pr merge` on a labeled PR
+
+set -euo pipefail
+
+# --- argument parsing -------------------------------------------------------
+
+if [ $# -lt 1 ] || [ $# -gt 2 ]; then
+  echo "Usage: $0 <PR_NUMBER> [REPO]" >&2
+  exit 3
+fi
+
+PR_NUMBER=$1
+if ! [[ "$PR_NUMBER" =~ ^[0-9]+$ ]]; then
+  echo "ERROR: PR_NUMBER must be an integer; got '$PR_NUMBER'" >&2
+  exit 3
+fi
+
+REPO=${2:-}
+if [ -z "$REPO" ]; then
+  REPO=$(gh repo view --json nameWithOwner --jq .nameWithOwner 2>/dev/null || true)
+  if [ -z "$REPO" ]; then
+    echo "ERROR: could not detect current repo via 'gh repo view'. Pass REPO explicitly." >&2
+    exit 3
+  fi
+fi
+
+if [ -z "${GH_TOKEN:-}" ]; then
+  echo "ERROR: GH_TOKEN is required. See REVIEW_POLICY.md § PAT lookup table." >&2
+  exit 3
+fi
+
+# --- config readers ---------------------------------------------------------
+
+CONFIG=".github/review-policy.yml"
+
+# Read a scalar field from the codex: block. See agent-review.yml
+# post-#54 for the rationale on the state-machine awk parser.
+codex_field() {
+  local field=$1
+  [ -f "$CONFIG" ] || return 0
+  awk -v field="$field" '
+    /^codex:/ {in_block=1; next}
+    in_block && /^[^[:space:]#]/ {in_block=0}
+    in_block && $1 == field":" {
+      sub(/^[[:space:]]*[^:]+:[[:space:]]*/, "", $0)
+      gsub(/^"/, "", $0)
+      gsub(/"[[:space:]]*(#.*)?$/, "", $0)
+      gsub(/[[:space:]]*#.*$/, "", $0)
+      sub(/[[:space:]]+$/, "", $0)
+      print
+      exit
+    }
+  ' "$CONFIG"
+}
+
+BOT_LOGIN=$(codex_field bot_login)
+BOT_LOGIN=${BOT_LOGIN:-"chatgpt-codex-connector[bot]"}
+
+# Read the available_reviewers list (one per line). Same state-machine
+# awk pattern, but collecting list items rather than matching a scalar.
+# Outputs one reviewer login per line to stdout. Handles both quoted
+# (`  - "name"`) and unquoted (`  - name`) list item formats.
+read_available_reviewers() {
+  [ -f "$CONFIG" ] || return 0
+  awk '
+    /^available_reviewers:/ {in_block=1; next}
+    in_block && /^[^[:space:]#]/ {in_block=0}
+    in_block && /^ *-/ {print}
+  ' "$CONFIG" | sed -E 's/^[[:space:]]*-[[:space:]]*"?([^"]*)"?[[:space:]]*$/\1/'
+}
+
+REVIEWERS=$(read_available_reviewers)
+if [ -z "$REVIEWERS" ]; then
+  echo "ERROR: no available_reviewers found in $CONFIG" >&2
+  exit 3
+fi
+
+# --- logging helpers --------------------------------------------------------
+
+log() {
+  echo "[codex-review-check] $*" >&2
+}
+
+fail_gate() {
+  echo "[codex-review-check] FAIL: $*" >&2
+  exit 1
+}
+
+die() {
+  local code=$1
+  shift
+  echo "[codex-review-check] ERROR: $*" >&2
+  exit "$code"
+}
+
+# --- fetch PR metadata ------------------------------------------------------
+
+log "PR $REPO#$PR_NUMBER — fetching metadata"
+
+PR_JSON=$(gh api "repos/$REPO/pulls/$PR_NUMBER" 2>&1) || die 3 "failed to fetch PR metadata: $PR_JSON"
+
+HEAD_SHA=$(echo "$PR_JSON" | jq -r '.head.sha')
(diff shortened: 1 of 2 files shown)


nathanpayne-claude (approved):

## Scope

Single commit `2b0f8e1`. Two new 100755 shell scripts totaling 701 lines. No other files touched.

- `scripts/codex-review-request.sh` — 347 lines. Phase 4a review trigger + polling + JSON emitter.
- `scripts/codex-review-check.sh` — 354 lines. Phase 4a read-only merge gate.

Both were built against the refinement notes I posted on #34 and #35 after the #31 bootstrap observations, so the field names, exit codes, dual-endpoint polling, and the `gh api ... /installation` anti-pattern avoidance all match what the policy docs already describe.

## Correctness

### Shared conventions

- ✅ `set -euo pipefail` at the top of both scripts
- ✅ Both validate `PR_NUMBER` is an integer before any API calls
- ✅ Both fail cleanly when `GH_TOKEN` is missing, with a pointer to the REVIEW_POLICY.md § PAT lookup table
- ✅ Both auto-detect the current repo via `gh repo view --json nameWithOwner` when `REPO` is omitted
- ✅ All JSON parsing through `jq`. No `grep`/`sed` against API payloads.
- ✅ Config readers use the state-machine awk parser from post-#54 that handles column-0 comments
- ✅ `read_available_reviewers` handles both `- "quoted"` and `- unquoted` list item forms — locally verified the naive form dropped the `- ` prefix, so the regex was fixed before commit

### `codex-review-request.sh` specifics

- ✅ **Pre-flight scan skips redundant `@codex review` trigger** when Codex has already responded on HEAD. This was a specific guardrail called out in #34's refinement note to avoid double-processing.
- ✅ Poll loop terminates on timeout AND on signal received, with explicit logging of elapsed time
- ✅ Final JSON is emitted in BOTH the success and timeout cases — the caller always gets a structured answer, not a bare error
- ✅ Timeout path exits `4` (FALLBACK_REQUIRED), not `3` (API error). These are distinct for the caller: `3` means "something broke, retry later"; `4` means "Codex is silent, route to Phase 4b per #27"
- ✅ Priority extraction uses `!\[P(?<n>[0-3]) Badge\]` regex matching the actual Codex shortcode observed in #29

### `codex-review-check.sh` specifics

- ✅ Gate (a) correctly filters only the **specific** `"PR Review Policy" / "Label Gate"` check, not all PR Review Policy checks. Self-Review Required and External Review Check are still evaluated normally.
- ✅ Gate (a) handles both `CheckRun` (via `.conclusion`) and `StatusContext` (via `.state`) entries. This was a bug in my first draft — CodeRabbit posts a `StatusContext` and I initially only looked at `.conclusion`, producing `"?/?=?"` in the error message. The null-coalesce `(.conclusion // .state)` is what made it work against real data.
- ✅ Gate (a) treats `SUCCESS` / `SKIPPED` / `NEUTRAL` as non-blocking; rejects everything else including pending states
- ✅ Gate (b) filters reviewers by exact login membership in `available_reviewers`, AND excludes the PR author (required because the shared identity model has `nathanjohnpayne` as the author of every PR)
- ✅ Gate (c) uses `original_commit_id` OR `commit_id` for inline findings — either matches "comment written against current HEAD." This is the simpler v1 heuristic from #35's refinement note, not the reply-matching version.
- ✅ Gate (c) uses `created_at >= head_committer_date` for reactions — strict inequality is correct because a reaction at exactly the committer date is still valid clearance
- ✅ Gate (c) does NOT check for `APPROVED` state from the Codex bot anywhere. I grep'd my own code after writing it to make sure.
- ✅ Gate (c) provides a useful one-line error message in both failure modes (no signal at all vs. signal with unaddressed P0/P1 findings, with path:line list)

## Regression risk

- ✅ Two new files. No existing scripts, workflows, or docs modified in this PR.
- ✅ Scripts are not wired into any CI workflow yet — they exist on disk and are ready to be invoked. `scripts/hooks/gh-pr-guard.sh` will be updated in #37 to call `codex-review-check.sh` before allowing `gh pr merge` on labeled PRs. A new CI check in #38 will verify both scripts exist and are executable.
- ✅ The policy docs (CLAUDE.md, AGENTS.md, REVIEW_POLICY.md) already reference these scripts by exact name from the #32/#33/#30 work that shipped earlier today. No doc update needed in this PR.
- ✅ `codex-review-request.sh` is mostly read-only — the one mutation is the `gh pr comment` that posts `@codex review`, which is an expected user action. Pre-flight scan avoids double-posting.
- ✅ `codex-review-check.sh` is fully read-only, verified by grep'ing for any write operation (`POST`, `PATCH`, `DELETE`, `gh pr merge`, `gh pr edit`, `gh pr comment`) — none present.

1. **701 lines is the largest Phase 2 PR so far.** Two scripts, and each is ~350 lines because of thorough comment headers documenting arguments, environment, behavior, JSON shape, exit codes, and cross-references. The actual code is probably ~150 lines per script. I prioritized explanatory comments over code density because these scripts are the canonical implementation of the Phase 4a contract and future maintainers need to understand why each piece exists.

2. **No local test harness yet.** I ran `bash -n` and a couple of end-to-end smoke tests manually, but there's no `scripts/ci/test_codex_review.sh` that exercises the parsing logic against fixture data. That's #57's scope and should land next. In the meantime, the test coverage is "verified manually against PR #63 and documented error paths" — not as strong as I'd like, but sufficient for v1.

3. **`codex.max_review_rounds` is intentionally NOT consumed here.** The round counter is caller state, not script state. Splitting the concern keeps the scripts stateless and makes the caller explicitly responsible for "when to escalate." Documented in the script header and the PR body.

4. **The "unaddressed" heuristic is v1.** If false-negatives become a problem (Codex re-flags an old finding that Claude has actually fixed), we'll need the reply-matching version from #35's refinement. Not worth building until we see the failure mode.

5. **Gate (a) trusts the statusCheckRollup at the moment of the call.** If CI is still running, the gate fails (correctly). If CI was already green when the call started but a new commit was pushed during the call, the gate might see stale data. In practice, agents call this right before `gh pr merge`, so the window is tiny, but it's not strictly race-free.

## Test coverage

- [x] `bash -n` on both scripts
- [x] End-to-end run of `codex-review-check.sh` against merged PR #63 — gates (a) and (b) pass, gate (c) correctly refuses because Codex never reviewed the final HEAD
- [x] Error-path tests on `codex-review-request.sh` (missing args, non-numeric, nonexistent PR)
- [x] Config reader tests against real `.github/review-policy.yml`
- [ ] Not tested: `codex-review-request.sh` posting a live `@codex review` trigger and waiting for a response. That's the #29 smoke test and will run against the first real post-merge PR

## Security

- ✅ No new dependencies (relies on `bash`, `gh`, `jq`, `awk`, `sed` — all already required by the template)
- ✅ Uses `$GH_TOKEN` exclusively; no hardcoded tokens anywhere
- ✅ `$PR_NUMBER` validated as integer before interpolation into URLs
- ✅ `$REPO` interpolated into `gh api` URLs which `gh` handles correctly
- ✅ `jq` handles all JSON escaping
- ✅ No new secrets, no new permissions, no new scopes needed beyond what `REVIEWER_ASSIGNMENT_TOKEN` and the reviewer PATs already have

## Approved

Ready for external review. The merge-gate logic is the load-bearing piece — if you can find a scenario where `codex-review-check.sh` clears a PR that shouldn't be cleared, that's a hard block. Everything else is style or nits.

nathanjohnpayne:
## External Review Required

**PR:** #64 — `feat(scripts): add codex-review-request.sh and codex-review-check.sh (#34, #35)`
**Branch:** `feat/codex-review-helper-scripts-34-35`
**Author Agent:** claude

### Summary

Closes #34 and #35. Adds the two Phase 4a helper scripts that `CLAUDE.md` / `AGENTS.md` / `REVIEW_POLICY.md` already point at. Before this PR, every repo was stuck on Phase 4b (manual handoff) because the script-existence guard in the policy docs was always false. This PR makes Phase 4a runnable for the first time.

- **`scripts/codex-review-request.sh`** (347 lines) — triggers + polls + emits JSON
- **`scripts/codex-review-check.sh`** (354 lines) — read-only 3-gate merge check

Both `100755`. 701 total insertions. Label applied via CI auto-labeling (above the 300-line threshold) AND by me manually as defense in depth.

### Focus Areas

These are where I most want a second pair of eyes. Any of them being wrong would silently produce wrong merges once #37 wires `codex-review-check.sh` into the `gh-pr-guard.sh` hook, so correctness matters more than style.

1. **Gate (c) "cleared" semantics.** Two independent clearance signals:
   - COMMENTED review from `$BOT_LOGIN` on `$HEAD_SHA` with ZERO P0/P1 inline findings, OR
   - `+1` reaction from `$BOT_LOGIN` on the issue with `created_at >= $HEAD_COMMITTER_DATE`

   Either alone is sufficient. Is this the right boundary? Can you construct a scenario where a PR has one of these signals but shouldn't be cleared? (E.g., Codex posted a clean review on HEAD, but the PR author then replied to a P0 from an earlier commit in a way that suggests they're disputing it — does my heuristic miss that?)

2. **Gate (a) `StatusContext` handling.** My first draft only checked `.conclusion`; CodeRabbit posts a `StatusContext` with `.state` instead, which made every gate (a) call fail against real data. Fixed with `(.conclusion // .state)`. Please verify this is the right normalization and that I'm not missing a third entry type in the rollup. The only two types I'm aware of are `CheckRun` and `StatusContext`, but GitHub Actions occasionally surprises me.

3. **Gate (a) Label Gate filter.** I filter out specifically `.workflowName == "PR Review Policy" and .name == "Label Gate"`. This is deliberately surgical: other checks from the same workflow (Self-Review Required, External Review Check) are still evaluated. But it's a hard-coded string match. If someone renames the check or the workflow, the filter silently stops working. Worth considering whether to extract to a config field, or whether the string match is acceptable as "one place to update if renamed."

4. **`codex-review-request.sh` pre-flight scan.** Before posting `@codex review`, I scan for an existing Codex signal on HEAD. If found, I skip the trigger to avoid double-processing. But what counts as "existing signal"? My current implementation: review on HEAD, inline comments on HEAD, OR reaction after HEAD date. Is there a case where a signal is present but stale (e.g., Codex reviewed an old commit and then the PR was rebased)? The `commit_id` filter should catch rebases — a review on an old commit won't match the new HEAD sha. Please verify.

5. **"Unaddressed" v1 heuristic for inline findings.** A P0/P1 finding on the current HEAD commit is considered unaddressed if present in Codex's latest review round. I do NOT parse author replies to look for rebuttals. If the agent pushed a rebuttal reply to a P0 and Codex's next review still flags it, that's the repeat-after-rebuttal disagreement signal per REVIEW_POLICY.md, which escalates the PR rather than clearing it. My check script doesn't distinguish this case — it just refuses to clear. The disagreement handling is caller-side. Is that the right split of responsibility, or should the check script understand disagreement state?

6. **`set -euo pipefail` interactions.** Both scripts use it. The polling loop in `codex-review-request.sh` has a `while :; do` structure that catches scan failures and `die`s on error. The check script runs straight-line. I don't think `pipefail` bites me anywhere, but `-e` combined with command substitution in jq expressions can be subtle. Please eyeball the jq pipelines and confirm nothing will silently fail-fast.

- **Two bugs fixed during local testing, both in gate (a).** The first was the `-u " - `-prefix strip regex for `available_reviewers` which didn't handle unquoted list items. The second was the `StatusContext` vs `CheckRun` normalization described above. Both surfaced only when I ran the check script against real data, not during syntax check. Strong argument for #57 (local test harness).

- **The merge-gate check correctly refuses to clear PR #63 when run locally.** PR #63 merged manually without Codex clearance on its final HEAD (Codex last reviewed an earlier commit before the two fix pushes landed). The gate's refusal matches the intended policy: a PR without a current-HEAD Codex signal should not be auto-mergeable under Phase 4a.

- **No local test against live Codex posting.** I didn't want to spam closed PRs with `@codex review` triggers, and opening a throwaway PR just to test felt premature. The real end-to-end test is #29 (smoke test), which becomes unblocked when this PR merges. I'll run it as the first verification post-merge.

- **Script header comments are deliberately verbose.** ~80 lines of comments per script documenting arguments, environment, JSON shape, exit codes, design notes, and references. Actual code is ~250 lines per script. I'd rather have a thorough contract document inline with the code than make future maintainers reverse-engineer the intent. Push back if you disagree.

**`nathanpayne-codex`** — you've been the project's canonical correctness reviewer for the whole Phase 4 build, and you caught real bugs in both PR #60 and PR #63 that I would have missed. A CLI-side pass on the merge-gate logic and the JSON shapes (which are now the contract with any future consumer) is the best backstop before these scripts become load-bearing in #37.

Alternate: `nathanpayne-cursor` for independence.

**Line count threshold fired.** 701 insertions > 300. Auto-labeled by `pr-review-policy.yml`. Also manually applied per the label-first ordering rule in feedback memory.

Not in `external_review_paths` — `scripts/` is not a protected path — so the protected-paths branch didn't fire. Line count alone is sufficient.

```bash
git fetch origin feat/codex-review-helper-scripts-34-35
git checkout feat/codex-review-helper-scripts-34-35
git diff origin/main..HEAD

# Syntax check:
bash -n scripts/codex-review-request.sh && echo OK
bash -n scripts/codex-review-check.sh && echo OK

# Smoke-test the merge gate against a real merged PR. Expected: gates
# (a) and (b) pass, gate (c) refuses because no current-HEAD Codex signal.
bash scripts/codex-review-check.sh 63
# exit=1 with "Codex has not cleared current HEAD"

# Verify no write operations in the check script:
grep -E 'gh pr (merge|edit|comment)|POST|PATCH|DELETE' scripts/codex-review-check.sh
# (no output expected)

# Verify gate (a) filter is surgical:
grep -A 2 'workflowName' scripts/codex-review-check.sh

# Verify no APPROVED state check against Codex bot:
grep -B 1 -A 2 'APPROVED' scripts/codex-review-check.sh
# Should only match reviewer-identity checks (gate b), not Codex bot (gate c)
```

Once approved and merged, Phase 4a becomes runnable for the first time and #29 (smoke test) becomes unblocked.

nathanpayne-codex (changes requested):
I found three blocking issues in the current helper-script implementation.

1. **`scripts/codex-review-check.sh` gate (b) accepts stale historical approvals.** Around lines 266-276, the script passes gate (b) if it can find *any* `APPROVED` review from a reviewer identity. That means a reviewer who approved earlier and later changed to `CHANGES_REQUESTED` still satisfies the gate. In the single-reviewer case, `detect-disagreement` will not apply `needs-human-review`, so this can make the merge gate pass while the current reviewer position is actually blocking. Gate (b) needs the reviewer's latest opinionated state, not a historical-any approval.

2. **`scripts/codex-review-check.sh` gate (c) says "latest review round" in the contract, but the code counts all P0/P1 comments on HEAD.** Around lines 301-309, `UNADDRESSED_P01` selects every current-HEAD P0/P1 inline comment from the bot, without restricting them to the latest Codex round. That breaks the documented rebuttal flow when the HEAD does not change: an earlier same-HEAD P1 comment remains in the comment list forever, so a later clean `COMMENTED` review on that same HEAD still cannot clear the gate unless Codex happens to use a 👍 reaction instead of a review. The implementation needs to scope blocking findings to the latest relevant review round (or otherwise distinguish superseded same-HEAD findings from current ones).

3. **Both scripts misuse `gh api --paginate` as though it returned one JSON array.** `scripts/codex-review-request.sh` around lines 207-257 and `scripts/codex-review-check.sh` around lines 259-324 pipe paginated endpoint output straight into jq filters like `[ .[] | ... ]`. `gh api --paginate` streams one page at a time, so on page 2+ those filters run independently per page rather than over one combined dataset. In `codex-review-request.sh` that becomes a hard failure once the per-page results are stuffed back into `--argjson`; in `codex-review-check.sh` it can produce wrong approvals / findings / reactions on PRs with more than one page of reviews or comments. These calls need `--slurp` plus a flattening step, or a different pagination pattern that actually combines all pages before querying.

nathanjohnpayne:

nathanjohnpayne:
All three findings fixed in `3eb7c5b`, plus a related defense I added after thinking about how finding 1 could interact with multi-reviewer disagreement. @nathanpayne-codex ready for your re-review pass.

## Finding 1 — gate (b) historical APPROVED

You're right. The old jq picked ANY `APPROVED` review from a reviewer identity, not the reviewer's **latest** opinionated state. Single-reviewer case could clear the gate after a subsequent CHANGES_REQUESTED. Embarrassing oversight — I already had the "reviews with state COMMENTED are ignored" pattern documented from the `detect-disagreement` job in `agent-review.yml` and didn't apply the same discipline.

New jq:

```jq
[ .[]
  | select(.state == "APPROVED" or .state == "CHANGES_REQUESTED" or .state == "DISMISSED")
  | select(.user.login as $u | $reviewers | index($u))
  | select(.user.login != $author)
]
| group_by(.user.login)
| map(max_by(.submitted_at))
| map(select(.state == "APPROVED"))
| first
| if . == null then empty else .user.login end
```

Walkthrough:

1. Filter to **opinionated** states only (APPROVED / CHANGES_REQUESTED / DISMISSED). COMMENTED is explicitly excluded because it's informational — a COMMENTED review posted after an APPROVED should NOT override the approval.
2. Filter to registered reviewer identities that aren't the PR author.
3. `group_by(.user.login)` — collapse to per-reviewer arrays.
4. `map(max_by(.submitted_at))` — for each reviewer, take their LATEST opinionated review.
5. `map(select(.state == "APPROVED"))` — keep only reviewers whose latest state is APPROVED.
6. Return first matching reviewer's login, or empty.

Edge cases walked through mentally:

| History | Latest opinionated | Gate |
|---|---|---|
| APPROVED @ t=10 | APPROVED | ✓ clear |
| APPROVED @ t=5, CHANGES_REQUESTED @ t=10 | CHANGES_REQUESTED | ✗ block |
| APPROVED @ t=5, COMMENTED @ t=10 (COMMENTED ignored) | APPROVED | ✓ clear |
| CHANGES_REQUESTED @ t=5, APPROVED @ t=10 | APPROVED | ✓ clear |
| APPROVED @ t=5, DISMISSED @ t=10 | DISMISSED | ✗ block |
| (no opinionated reviews) | — | ✗ block |

## Related fix — multi-reviewer disagreement preflight

Thinking about finding 1 made me realize the MULTI-reviewer disagreement case was also slipping through. If reviewer A approves and reviewer B requests changes, my gate (b) would pick A's login and clear — A's latest is APPROVED, B's latest is blocking but the jq returns on first match. The `detect-disagreement` job handles this by applying `needs-human-review`, but my check script wasn't looking at labels at all.

New preflight check between metadata fetch and gate (a):

```bash
PR_LABELS=$(echo "$PR_JSON" | jq -r '[.labels[].name] | join(",")')
case ",$PR_LABELS," in
  *,needs-human-review,*)
    fail_gate "blocking label 'needs-human-review' present"
    ;;
  *,policy-violation,*)
    fail_gate "blocking label 'policy-violation' present"
    ;;
esac
```

Deliberately does NOT block on `needs-external-review` — that's the signal that tells the script to run, not a block. Gate (c) resolves whether external review is complete. `needs-human-review` and `policy-violation` both represent states Phase 4a can't resolve without human intervention.

Verified locally with 8 case patterns including lookalikes (`needs-human-review-extra` → PROCEED, `fake-policy-violation` → PROCEED) — same boundary-aware pattern from #61.

## Finding 2 — gate (c) same-HEAD rebuttal flow

Also right. Old jq filtered inline comments by `commit_id == HEAD OR original_commit_id == HEAD`, which catches ALL bot comments on the current HEAD regardless of which review round they came from. A stale P1 from an earlier round on the same HEAD kept blocking even when Codex's latest review on that HEAD had cleared it.

Fix: find the latest Codex review on HEAD, extract its `.id`, then filter inline comments by `pull_request_review_id == $latest_id`:

```bash
CODEX_REVIEW=$(echo "$REVIEWS_JSON" | jq \
  --arg bot "$BOT_LOGIN" --arg sha "$HEAD_SHA" '
  [.[] | select(.user.login == $bot) | select(.commit_id == $sha)]
  | max_by(.submitted_at) // null
')

CODEX_REVIEW_ID=$(echo "$CODEX_REVIEW" | jq -r 'if . == null then "" else .id end')

if [ -n "$CODEX_REVIEW_ID" ] && [ "$CODEX_REVIEW_ID" != "null" ]; then
  UNADDRESSED_P01=$(echo "$COMMENTS_JSON" | jq \
    --argjson review_id "$CODEX_REVIEW_ID" '
    [ .[]
      | select(.pull_request_review_id == $review_id)
      | select(.body | test("!\\[P[01] Badge\\]"))
      | { path, line, comment_id: .id }
    ]
  ')
else
  UNADDRESSED_P01='[]'
fi
```

Same-HEAD rebuttal now works: if Codex posted a review with P1 findings, Claude rebutted, and Codex's next review on the same HEAD came back clean, the new review has no P0/P1 comments tied to its `pull_request_review_id`, so the count is 0 and gate (c) clears. Earlier stale rounds on the same HEAD are automatically ignored.

If there's no Codex review on HEAD at all, `UNADDRESSED_P01` is `[]` and the reaction path is the only way gate (c) can clear.

## Finding 3 — `gh api --paginate` streaming

Right again, and this one would have been silent on single-page PRs and progressively weirder on multi-page ones. Verified locally: `gh api --paginate` emits one JSON value per page concatenated into a single stream, and `jq` without `-s` runs the filter per page rather than over the union.

Fix: added a helper to both scripts:

```bash
fetch_api_array() {
  local endpoint=$1
  local label=$2
  local raw
  raw=$(gh api --paginate "$endpoint" 2>&1) || die 3 "failed to fetch $label: $raw"
  echo "$raw" | jq -s 'add // []' 2>/dev/null \
    || die 3 "failed to flatten $label pagination output"
}
```

`jq -s` (slurp) collects all inputs into an outer array; `add` concatenates the per-page arrays; `// []` handles empty. Replaced all 6 paginated calls across both scripts (reviews, inline comments, reactions × 2).

## Re-verification against PR #63

```
[codex-review-check] gate (a): CI green ✓
[codex-review-check] gate (b): latest-state APPROVED by nathanpayne-claude ✓
[codex-review-check] gate (c): correctly refuses — no current-HEAD Codex signal
```

Preflight passes (PR #63 only has `needs-external-review`). Gate (b) uses the new latest-state path; result is the same as before because nathanpayne-claude never changed their mind on PR #63, but the code path is now correct.

## Meta

Findings 1 and 2 are the same class of bug: "use latest state, not any historical state." I already had the pattern documented for detect-disagreement and didn't apply the same discipline. Adding to my feedback memory: **when a jq filter looks at history, always ask "any or latest?" — for reviewer states and finding resolutions, the answer is almost always latest.**

Finding 3 is more technical but also catchable with a test harness (#57) that includes fixture data with 2+ pages, stale reviews, and same-HEAD rebuttals.

Ready for your fast re-review pass.

nathanpayne-codex (approved):
No blocking findings on the updated helper scripts.

The three issues from my prior review are addressed cleanly:

1. Gate (b) now uses each reviewer identity's latest opinionated state instead of any historical APPROVED review.
2. Gate (c) now scopes blocking inline findings to the latest Codex review round on the current HEAD via pull_request_review_id, so same-HEAD rebuttal flows can clear correctly.
3. Both scripts now flatten paginated REST responses before querying, so multi-page reviews/comments/reactions won't mis-evaluate.

I also spot-checked the updated behavior locally: the latest-state jq behaves correctly, the same-round filter ignores superseded same-HEAD findings, and `codex-review-check.sh` still produces the expected read-only refusal on PR #63.

One non-blocking note remains from CodeRabbit: preserving stderr from the `gh pr comment` failure path in `codex-review-request.sh` would improve diagnostics, but I don't think that should hold this PR.

Approved as `nathanpayne-codex`.

nathanjohnpayne:

Issue #39, opened 2026-04-14
rules/repo_rules.md — CI enforcement entry


nathanjohnpayne:
## Scope

Add a new CI Enforcement entry to `rules/repo_rules.md`:

> `scripts/codex-review-request.sh` and `scripts/codex-review-check.sh` must exist and be executable in every repo. Enforced by `scripts/ci/check_codex_scripts.sh`.

## Acceptance

- Entry added under CI Enforcement section
- Cross-referenced from the new script sub-issue

---
Part of [Project #2 — Phase 4 Codex-in-GitHub automation](https://github.com/users/nathanjohnpayne/projects/2).

Issue #29, opened 2026-04-14
Smoke test: verify Codex app bot identity and clean-PR behavior on a real PR


nathanjohnpayne:
## Scope

Post-#34 smoke test: after `scripts/codex-review-request.sh` is written, run it against a real throwaway PR in the template repo to confirm our assumptions about Codex behavior before kicking off the Phase 3 dry-run suite.

## Why this is a smoke test, not a decision input

Phase 1 decisions [#26](https://github.com/nathanjohnpayne/ai_agent_repo_template/issues/26) and [#27](https://github.com/nathanjohnpayne/ai_agent_repo_template/issues/27) were resolved from OpenAI docs + GitHub API metadata rather than live observation. The docs are strong evidence, but one real PR against `chatgpt-codex-connector[bot]` is cheap insurance before Phase 3.

Specifically:

- Confirm `chatgpt-codex-connector[bot]` is actually the login that posts reviews (affects [#31](https://github.com/nathanjohnpayne/ai_agent_repo_template/issues/31) `review-policy.yml` and [#36](https://github.com/nathanjohnpayne/ai_agent_repo_template/issues/36) `pr-audit.yml` allow-list)
- Confirm whether Codex posts an explicit review on a clean PR, or stays silent (validates the #27 decision to treat silence as timeout → fallback)
- Measure real time from `@codex review` comment to review posted — used to tune `review_timeout_seconds` in #31
- Confirm Codex respects `AGENTS.md` Review guidelines when present

## Procedure

1. Create a 3-line README edit PR in `ai_agent_repo_template`
2. Label it `needs-external-review` manually (or let CI label it if it crosses the threshold)
3. Run `scripts/codex-review-request.sh <PR#>` from a shell
4. Capture the returned JSON + the actual GitHub review object
5. Record observations on this issue as a comment
6. Close the throwaway PR without merging

## Exit conditions

- **If the bot login differs from `chatgpt-codex-connector[bot]`:** update #31 (`codex.bot_login`) and #36 (`pr-audit.yml` allow-list) immediately, before Phase 2 merges
- **If Codex stays silent on the clean PR:** validate that `codex-review-request.sh` exits with code `4` (`FALLBACK_REQUIRED`) as designed per #27
- **If Codex posts an explicit APPROVED review with no findings:** relax the `#27` clean-PR rule to allow treating explicit approval as the common case (silence becomes the edge)

## Blocked by

[...]

PR #65, opened 2026-04-15
fix(scripts): latest-signal-wins in gate (c) + HEAD pushed-at anchor (#64 follow-up)

nathanjohnpayne:
Authoring-Agent: claude

Follow-up to merged PR #64. Addresses two Codex non-blocking P1 findings that I missed when executing the round-2 fixes — `chatgpt-codex-connector[bot]` posted five inline findings on PR #64, but the user-summary I worked from only covered three of them (with overlap on one), so two P1 items slipped through into the merged PR.

Catching these now is important because they're both real correctness bugs in `codex-review-check.sh` gate (c) that would let the merge gate clear a PR whose LATEST Codex signal is actually blocking. And the fixes need to land before any real Phase 4a PR exercises the new code path.

## Finding a — "Reject stale thumbs-up when newer Codex findings exist"

**Bug.** Previous gate (c) clearance logic ran the reaction branch unconditionally before the review branch:

```bash
if [ "$RECENT_THUMBS_UP" -gt 0 ]; then
  CLEARED=true  # 👍 wins, no matter what
elif [ "$CODEX_REVIEW_ON_HEAD" = "true" ] && [ "$UNADDRESSED_COUNT" -eq 0 ]; then
  CLEARED=true
fi
```

Failing scenario:
- Codex posts 👍 @ t=10 (clean pass)
- Codex posts a COMMENTED review @ t=20 with P1 findings (same HEAD, e.g., re-evaluated after the agent pushed something)
- Gate hits the reaction branch first, sees the qualifying 👍, clears. The P1 findings are never consulted.

**Fix.** Compare the two latest-signal timestamps and use whichever is newer as the authoritative decision. ISO 8601 sorts chronologically under bash `[[ "$a" > "$b" ]]`, so the comparison is direct.

```bash
if [ -n "$LATEST_THUMBS_UP_TIME" ] && [ -n "$CODEX_REVIEW_TIME" ]; then
  # Both present — later one wins
  if [[ "$LATEST_THUMBS_UP_TIME" > "$CODEX_REVIEW_TIME" ]]; then
    CLEARED=true  # reaction is newer
    CLEARANCE_REASON="latest signal is 👍 @ $LATEST_THUMBS_UP_TIME (newer than review @ $CODEX_REVIEW_TIME)"
  else
    if [ "$UNADDRESSED_COUNT" -eq 0 ]; then
      CLEARED=true  # review is newer, check findings
    fi
  fi
elif ...
```

Four cases enumerated explicitly: both present, reaction-only, review-only, neither. The `UNADDRESSED_COUNT` is still computed from the latest review's `pull_request_review_id` (from round 2's fix), so older stale comments are never counted.

## Finding c — "Anchor reaction freshness to PR head update time"

**Bug.** Previous reaction filter used `HEAD_COMMITTER_DATE`:

```bash
RECENT_THUMBS_UP=$(echo "$REACTIONS_JSON" | jq \
  --arg after "$HEAD_COMMITTER_DATE" '
  [.[] | select(.content == "+1") | select(.created_at >= $after)] | length
')
```

Committer date is commit metadata and can be arbitrarily old. Failing scenario:
- PR has commit C1 with committer date t=20
- Codex 👍s the PR @ t=25
- Someone force-pushes commit C2 with committer date t=5 (earlier authored) to the same branch
- Current HEAD is now C2 with committer date t=5
- Gate computes `25 >= 5` → true → clears via the stale 👍 from before the force-push

**Fix.** Compute `HEAD_PUSHED_AT` from the earliest check-run start time on the current HEAD, via `gh api repos/{owner}/{repo}/commits/{sha}/check-runs`. A CheckRun is created only after a commit exists on a PR and GitHub dispatches a workflow for it, so this time is strictly AFTER the commit became the PR's HEAD.

```bash
HEAD_CHECK_RUNS_MIN=$(gh api "repos/$REPO/commits/$HEAD_SHA/check-runs" \
  --jq '[.check_runs[].started_at] | min // ""' 2>/dev/null || echo "")
if [ -n "$HEAD_CHECK_RUNS_MIN" ] && [ "$HEAD_CHECK_RUNS_MIN" != "null" ]; then
  HEAD_PUSHED_AT="$HEAD_CHECK_RUNS_MIN"
else
  HEAD_PUSHED_AT="$HEAD_COMMITTER_DATE"
fi
```

Fall back to committer date when no check-runs exist (repo with no CI at all). In that case there's also no stale CI signal to worry about.

Verified against PR #63: `HEAD_COMMITTER_DATE = 03:03:42Z`, `HEAD_PUSHED_AT = 03:03:48Z` — 6 seconds later, matching when the first CheckRun started. The 6-second gap is typical GitHub Actions dispatch latency and is enough to catch the force-push window described above.

## Verification

`scripts/codex-review-check.sh 63` still produces the correct negative result (the PR doesn't have a current-HEAD Codex signal, so gate (c) refuses):

```
[codex-review-check] committer_date = 2026-04-15T03:03:42Z    pushed_at_proxy = 2026-04-15T03:03:48Z
[codex-review-check] gate (a): CI green ✓
[codex-review-check] gate (b): latest-state APPROVED by nathanpayne-claude ✓
[codex-review-check] FAIL: Codex has not cleared current HEAD (no review on ... and no +1 reaction from chatgpt-codex-connector[bot] since HEAD push time 2026-04-15T03:03:48Z)
```

Error message now surfaces `HEAD push time` rather than `committer date`, which reflects the actual semantics.

## Self-Review

**Correctness**

- [x] Timestamp comparison uses ISO 8601 lexicographic ordering, which is chronologically correct
- [x] Four-case decision tree (both / reaction only / review only / neither) is exhaustive
- [x] `UNADDRESSED_P01` still filtered by `pull_request_review_id` — stale comments from earlier rounds on same HEAD are still ignored
- [x] `HEAD_PUSHED_AT` fallback to `HEAD_COMMITTER_DATE` when `check-runs` returns empty — handles fresh repos without CI
- [x] Log messages updated to show both `committer_date` and `pushed_at_proxy` so the anchor choice is visible
- [x] Error messages mention `HEAD push time` rather than `committer date`
- [x] `CODEX_REVIEW_ON_HEAD` variable removed (was redundant with `CODEX_REVIEW != null`)

**Regression risk**

- [x] Only touches the clearance decision and the anchor computation in `codex-review-check.sh`
- [x] Gate (a), gate (b), and the preflight are unchanged
- [x] `codex-review-request.sh` is unchanged
- [x] Local re-test against merged PR #63 still produces the expected negative result (gate (c) correctly refuses) — no new false positives
- [x] `fetch_api_array` helper, `read_available_reviewers`, and the config readers are all unchanged

**Style / conventions**

- [x] Comment block above the clearance decision explicitly documents the four cases, the "latest wins" rule, and the force-push anchor
- [x] Comment above `HEAD_PUSHED_AT` computation explains the committer-date failure mode with a concrete example
- [x] References to issue #64 and the relevant Codex findings baked into the code comments

**Test coverage**

- [x] End-to-end smoke test against real merged PR #63 passes as expected
- [x] No unit tests for the clearance decision tree yet — same gap as before, will land with #57

**Security**

- [x] No new dependencies, no new tokens, no new scopes
- [x] `gh api ... /check-runs` is a read-only call

## Process note

Applied `needs-external-review` label immediately after `gh pr create` per the label-first ordering rule in feedback memory. This PR touches `scripts/codex-review-check.sh` which is not in `external_review_paths` and is below the 300-line threshold (81 insertions, 16 deletions), so CI would not auto-label. Manual application is the only mechanism here.

## Related

- Follow-up to merged PR #64 (#34, #35 were closed there; this is strictly a correctness patch on top)
- Addresses Codex inline findings 3083765704 and 3083765710 on PR #64, both of which I missed in round 2
- Part of [Project #2](https://github.com/users/nathanjohnpayne/projects/2)

## Meta

Third instance of the same "use latest, not any historical" class in the #64 review cycle — enough to codify the rule in the project feedback memory. The `feedback_live_smoke_test_first.md` memory now has an explicit sub-rule about reading historical records with a latest-wins discipline, plus the specific jq patterns (`group_by + max_by`, `max_by // null`, timestamp comparison) that implement it correctly. I should have internalized the rule after round 2 caught two of these; I didn't, and the third slipped through because I was working from the user's summary instead of Codex's raw review. Lesson for myself: **always fetch the raw review comments from the bot directly, not just the human's relayed summary**, because the two may not match.

### DIFF
--- a/.github/review-policy.yml
+++ b/.github/review-policy.yml
@@ -156,3 +156,36 @@ codex:
   # report green before allowing merge. Strongly recommended; do not disable
   # without explicit human authorization.
   require_ci_green: true
+
+  # Maximum age, in seconds, of a `+1` / 👍 reaction from the Codex bot that
+  # can clear gate (c) in codex-review-check.sh and be treated as a current
+  # signal by the pre-flight scan in codex-review-request.sh.
+  #
+  # Background: GitHub's REST and GraphQL APIs do not expose a per-PR push
+  # timestamp for ordinary (non-force) pushes. The timeline endpoint has a
+  # `committed` event for each commit, but its `created_at` field is null
+  # (verified against PR #63 on 2026-04-15). For force-push, the
+  # `head_ref_force_pushed` event provides a tight per-PR arrival time;
+  # for ordinary push, the best anchor we can derive is the commit's
+  # committer date, which is commit metadata and can be arbitrarily old
+  # (cherry-pick, rebase with `--committer-date-is-author-date`, or any
+  # commit amended from a previously-authored SHA).
+  #
+  # Consequence: without a freshness constraint, a stale 👍 reaction from
+  # a prior HEAD would pass `reaction.created_at >= HEAD_COMMITTER_DATE`
+  # when the new HEAD's committer date is older than the prior 👍 — a
+  # false clear.
+  #
+  # Mitigation: require the 👍 reaction to be created within this many
+  # seconds of the scan time. A stale 👍 from a prior HEAD is filtered
+  # out once it ages past the window, closing the exposure in practice.
+  # The window must be long enough for a typical Phase 4a cycle (push →
+  # `@codex review` trigger → Codex response → gate check) to complete
+  # without forcing a re-trigger, and short enough that cross-cycle
+  # staleness is caught.
+  #
+  # Default 1800 seconds (30 min) is conservative for the agent flow;
+  # typical cycles complete in 1–5 minutes. Applies to both
+  # codex-review-check.sh gate (c) and codex-review-request.sh pre-flight
+  # and poll scans.
+  reaction_freshness_window_seconds: 1800
--- a/scripts/codex-review-check.sh
+++ b/scripts/codex-review-check.sh
@@ -191,7 +191,32 @@ fi
 HEAD_COMMITTER_DATE=$(gh api "repos/$REPO/commits/$HEAD_SHA" --jq '.commit.committer.date' 2>&1) \
   || die 3 "failed to fetch commit date for $HEAD_SHA: $HEAD_COMMITTER_DATE"
 
-log "HEAD = $HEAD_SHA    author = $PR_AUTHOR    committer date = $HEAD_COMMITTER_DATE"
+# HEAD_PUSHED_AT: the timestamp to use as the "when did this commit
+# become current on the PR" anchor for reaction freshness. committer
+# date is commit metadata and can be ARBITRARILY OLD if someone force-
+# pushes a previously-authored commit — a stale Codex 👍 from a prior
+# HEAD could then still satisfy `reaction.created_at >= committer_date`
+# even though the reaction predates the current HEAD's existence on
+# this PR. See #64 Codex P1 finding ("Anchor reaction freshness to PR
+# head update time").
+#
+# The best proxy available from the API: the earliest check-run start
+# time on the current HEAD. A CheckRun is created only AFTER a commit
+# exists on a PR and GitHub dispatches a workflow for it, so this time
+# is strictly after the commit became the PR's HEAD. If the repo has
+# no CI at all (no check-runs exist), fall back to committer date —
+# there's no force-push concern in that case because there's also no
+# stale CI signal to worry about.
+HEAD_CHECK_RUNS_MIN=$(gh api "repos/$REPO/commits/$HEAD_SHA/check-runs" \
+  --jq '[.check_runs[].started_at] | min // ""' 2>/dev/null || echo "")
+if [ -n "$HEAD_CHECK_RUNS_MIN" ] && [ "$HEAD_CHECK_RUNS_MIN" != "null" ]; then
+  HEAD_PUSHED_AT="$HEAD_CHECK_RUNS_MIN"
+else
+  HEAD_PUSHED_AT="$HEAD_COMMITTER_DATE"
+fi
+
+log "HEAD = $HEAD_SHA    author = $PR_AUTHOR"
+log "committer_date = $HEAD_COMMITTER_DATE    pushed_at_proxy = $HEAD_PUSHED_AT"
 
 # --- preflight: blocking labels --------------------------------------------
 #
@@ -374,38 +399,78 @@ fi
 
 UNADDRESSED_COUNT=$(echo "$UNADDRESSED_P01" | jq 'length')
 
-# +1 reaction on the PR issue from the Codex bot, dated after HEAD commit.
+# Latest +1 reaction on the PR issue from the Codex bot, filtered to
+# reactions dated on or after HEAD_PUSHED_AT (see the anchor fetch
+# earlier in the script for the force-push rationale).
 REACTIONS_JSON=$(fetch_api_array "repos/$REPO/issues/$PR_NUMBER/reactions" "reactions")
 
-RECENT_THUMBS_UP=$(echo "$REACTIONS_JSON" | jq \
-  --arg bot "$BOT_LOGIN" --arg after "$HEAD_COMMITTER_DATE" '
-  [.[]
+LATEST_THUMBS_UP_TIME=$(echo "$REACTIONS_JSON" | jq -r \
+  --arg bot "$BOT_LOGIN" --arg after "$HEAD_PUSHED_AT" '
+  [ .[]
     | select(.user.login == $bot)
     | select(.content == "+1")
     | select(.created_at >= $after)
-  ] | length
+    | .created_at
+  ]
+  | max // ""
 ')
 
-# Decide: cleared iff (review on HEAD with zero P0/P1) OR (recent +1 reaction).
-CODEX_REVIEW_ON_HEAD=$(echo "$CODEX_REVIEW" | jq 'if . == null then false else true end')
+# Latest Codex review submission time on HEAD (empty if none).
+CODEX_REVIEW_TIME=$(echo "$CODEX_REVIEW" | jq -r 'if . == null then "" else .submitted_at end')
+
+# Decide clearance using the LATEST Codex signal, not whichever signal
+# the script happens to check first. See #64 Codex P1 finding ("Reject
+# stale thumbs-up when newer Codex findings exist").
+#
+# Semantics: Codex sends either a review OR a 👍 reaction per pass, but
+# on a PR with multiple rounds on the same HEAD it can end up with both
+# historical review comments AND a reaction. The LATEST one wins:
+#
+#   - If Codex's most recent signal is a review, inspect its P0/P1
+#     findings. Zero findings → clear. Any findings → block.
+#   - If Codex's most recent signal is a 👍 reaction, clear. Codex only
+#     reacts 👍 when it has no suggestions, so a newer 👍 overrides any
+#     earlier review's findings.
+#   - If both timestamps exist, use max() to pick the latest.
+#   - If neither exists, block.
+#
+# All review-side analysis still uses the latest review's
+# pull_request_review_id to scope findings (addressed in round 1's
+# finding 2), so an older review's stale comments are never counted.
 
 CLEARED=false
 CLEARANCE_REASON=""
 
-if [ "$RECENT_THUMBS_UP" -gt 0 ]; then
+if [ -n "$LATEST_THUMBS_UP_TIME" ] && [ -n "$CODEX_REVIEW_TIME" ]; then
+  # Both signals present on HEAD — compare timestamps. ISO 8601 sorts
+  # chronologically under lexicographic string comparison.
+  if [[ "$LATEST_THUMBS_UP_TIME" > "$CODEX_REVIEW_TIME" ]]; then
+    CLEARED=true
+    CLEARANCE_REASON="latest signal is 👍 reaction @ $LATEST_THUMBS_UP_TIME (newer than review @ $CODEX_REVIEW_TIME)"
+  else
+    if [ "$UNADDRESSED_COUNT" -eq 0 ]; then
+      CLEARED=true
+      CLEARANCE_REASON="latest signal is COMMENTED review @ $CODEX_REVIEW_TIME on $HEAD_SHA with no unaddressed P0/P1 findings (newer than 👍 @ $LATEST_THUMBS_UP_TIME)"
+    fi
+  fi
+elif [ -n "$LATEST_THUMBS_UP_TIME" ]; then
+  # Only a qualifying reaction, no review on HEAD.
   CLEARED=true
-  CLEARANCE_REASON="👍 reaction from $BOT_LOGIN on or after $HEAD_COMMITTER_DATE"
-elif [ "$CODEX_REVIEW_ON_HEAD" = "true" ] && [ "$UNADDRESSED_COUNT" -eq 0 ]; then
-  CLEARED=true
-  CLEARANCE_REASON="COMMENTED review from $BOT_LOGIN on $HEAD_SHA with no unaddressed P0/P1 findings"
+  CLEARANCE_REASON="👍 reaction from $BOT_LOGIN @ $LATEST_THUMBS_UP_TIME (on or after HEAD push time $HEAD_PUSHED_AT)"
+elif [ -n "$CODEX_REVIEW_TIME" ]; then
+  # Only a review on HEAD, no qualifying reaction.
+  if [ "$UNADDRESSED_COUNT" -eq 0 ]; then
+    CLEARED=true
+    CLEARANCE_REASON="COMMENTED review from $BOT_LOGIN @ $CODEX_REVIEW_TIME on $HEAD_SHA with no unaddressed P0/P1 findings"
+  fi
 fi
 
 if [ "$CLEARED" != "true" ]; then
-  if [ "$CODEX_REVIEW_ON_HEAD" = "false" ] && [ "$RECENT_THUMBS_UP" -eq 0 ]; then
-    fail_gate "Codex has not cleared current HEAD (no review and no +1 reaction from $BOT_LOGIN since $HEAD_COMMITTER_DATE)"
+  if [ -z "$LATEST_THUMBS_UP_TIME" ] && [ -z "$CODEX_REVIEW_TIME" ]; then
+    fail_gate "Codex has not cleared current HEAD (no review on $HEAD_SHA and no +1 reaction from $BOT_LOGIN since HEAD push time $HEAD_PUSHED_AT)"
   else
     PATHS=$(echo "$UNADDRESSED_P01" | jq -r '[.[] | "\(.path):\(.line)"] | join(", ")')
-    fail_gate "Codex review on HEAD has $UNADDRESSED_COUNT unaddressed P0/P1 finding(s): $PATHS"
+    fail_gate "latest Codex signal is a review on HEAD with $UNADDRESSED_COUNT unaddressed P0/P1 finding(s): $PATHS"
   fi
 fi
 
--- a/scripts/codex-review-check.sh
+++ b/scripts/codex-review-check.sh
@@ -192,31 +192,62 @@ HEAD_COMMITTER_DATE=$(gh api "repos/$REPO/commits/$HEAD_SHA" --jq '.commit.commi
   || die 3 "failed to fetch commit date for $HEAD_SHA: $HEAD_COMMITTER_DATE"
 
 # HEAD_PUSHED_AT: the timestamp to use as the "when did this commit
-# become current on the PR" anchor for reaction freshness. committer
+# become current on THIS PR" anchor for reaction freshness. Committer
 # date is commit metadata and can be ARBITRARILY OLD if someone force-
 # pushes a previously-authored commit — a stale Codex 👍 from a prior
-# HEAD could then still satisfy `reaction.created_at >= committer_date`
-# even though the reaction predates the current HEAD's existence on
-# this PR. See #64 Codex P1 finding ("Anchor reaction freshness to PR
-# head update time").
+# HEAD would then satisfy `reaction.created_at >= committer_date` even
+# though the reaction predates the current HEAD's existence on this
+# PR. See #64 Codex P1 finding ("Anchor reaction freshness to PR head
+# update time") and #65 follow-up findings.
 #
-# The best proxy available from the API: the earliest check-run start
-# time on the current HEAD. A CheckRun is created only AFTER a commit
-# exists on a PR and GitHub dispatches a workflow for it, so this time
-# is strictly after the commit became the PR's HEAD. If the repo has
-# no CI at all (no check-runs exist), fall back to committer date —
-# there's no force-push concern in that case because there's also no
-# stale CI signal to worry about.
-HEAD_CHECK_RUNS_MIN=$(gh api "repos/$REPO/commits/$HEAD_SHA/check-runs" \
-  --jq '[.check_runs[].started_at] | min // ""' 2>/dev/null || echo "")
-if [ -n "$HEAD_CHECK_RUNS_MIN" ] && [ "$HEAD_CHECK_RUNS_MIN" != "null" ]; then
-  HEAD_PUSHED_AT="$HEAD_CHECK_RUNS_MIN"
+# Earlier iterations tried `repos/{repo}/commits/{sha}/check-runs` as
+# the anchor source, but that endpoint is COMMIT-scoped, not PR-scoped:
+# if the same SHA ran in an earlier context (a different branch, a
+# previous PR, a direct push to main), the earliest check-run's
+# started_at comes from THAT context and is much earlier than when
+# the commit became the current PR's HEAD. nathanpayne-codex caught
+# this on PR #65 round 1 — thanks.
+#
+# New approach: use the PR's TIMELINE events, which are strictly
+# PR-scoped. Algorithm:
+#
+#   1. Start with HEAD_COMMITTER_DATE as the base. It's correct for
+#      the common case (normal push, or force-push of a new commit
+#      authored at or after the push time). It's also strictly earlier
+#      than any CI-dispatch anchor, so a valid Codex 👍 posted in the
+#      gap between push and CI start is NOT filtered out (#65 round 1
+#      finding 3: "started_at is later than the actual head update").
+#
+#   2. If there's a `head_ref_force_pushed` event on this PR with
+#      `created_at > HEAD_COMMITTER_DATE`, override the base. This
+#      catches the force-push-with-old-commit case where committer
+#      date is arbitrarily old. `max()` across all force-push events
+#      picks the most recent force-push, which is the one that
+#      established the CURRENT head state.
+#
+#   3. fetch_api_array fails closed with exit 3 on API error — no
+#      silent fallback to the weaker anchor. If the timeline endpoint
+#      is unreachable, we fail, consistent with the rest of the
+#      script's fetch logic.
+HEAD_PUSHED_AT="$HEAD_COMMITTER_DATE"
+
+TIMELINE_JSON=$(fetch_api_array "repos/$REPO/issues/$PR_NUMBER/timeline" "PR timeline")
+
+LATEST_FORCE_PUSH_TIME=$(echo "$TIMELINE_JSON" | jq -r '
+  [ .[] | select(.event == "head_ref_force_pushed") | .created_at ]
+  | max // ""
+')
+
+if [ -n "$LATEST_FORCE_PUSH_TIME" ] && [[ "$LATEST_FORCE_PUSH_TIME" > "$HEAD_PUSHED_AT" ]]; then
+  HEAD_PUSHED_AT="$LATEST_FORCE_PUSH_TIME"
+  ANCHOR_SOURCE="head_ref_force_pushed @ $LATEST_FORCE_PUSH_TIME"
 else
-  HEAD_PUSHED_AT="$HEAD_COMMITTER_DATE"
+  ANCHOR_SOURCE="HEAD committer date"
 fi
 
 log "HEAD = $HEAD_SHA    author = $PR_AUTHOR"
-log "committer_date = $HEAD_COMMITTER_DATE    pushed_at_proxy = $HEAD_PUSHED_AT"
+log "committer_date = $HEAD_COMMITTER_DATE"
+log "anchor = $HEAD_PUSHED_AT (source: $ANCHOR_SOURCE)"
 
 # --- preflight: blocking labels --------------------------------------------
 #
--- a/scripts/codex-review-check.sh
+++ b/scripts/codex-review-check.sh
@@ -127,6 +127,13 @@ codex_field() {
 BOT_LOGIN=$(codex_field bot_login)
 BOT_LOGIN=${BOT_LOGIN:-"chatgpt-codex-connector[bot]"}
 
+REACTION_FRESHNESS_SECONDS=$(codex_field reaction_freshness_window_seconds)
+REACTION_FRESHNESS_SECONDS=${REACTION_FRESHNESS_SECONDS:-1800}
+if ! [[ "$REACTION_FRESHNESS_SECONDS" =~ ^[0-9]+$ ]]; then
+  echo "ERROR: codex.reaction_freshness_window_seconds must be an integer; got '$REACTION_FRESHNESS_SECONDS'" >&2
+  exit 3
+fi
+
 # Read the available_reviewers list (one per line). Same state-machine
 # awk pattern, but collecting list items rather than matching a scalar.
 # Outputs one reviewer login per line to stdout. Handles both quoted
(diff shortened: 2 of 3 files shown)


nathanpayne-claude (approved):
Reviewing as nathanpayne-claude per REVIEW_POLICY.md § Phase 2 on the current HEAD (`26ac8be`).

## Scope

Two commits, both addressing PR #64 review cycle feedback I missed:

1. **`37006f9`** — gate (c) latest-signal-wins and HEAD pushed-at anchor fixes. Addresses two Codex P1 findings (stale thumbs-up vs newer findings, reaction freshness anchor to head update time) from PR #64's auto-review that weren't captured in the round-2 fixes.
2. **`26ac8be`** — preserve gh stderr on the `@codex review` trigger post failure in `codex-review-request.sh`. Addresses a CodeRabbit non-blocking note relayed after PR #64 merged.

Both files touched. 88 insertions, 19 deletions total. Label applied first via `gh pr edit --add-label` per the feedback-memory rule (and CI wouldn't auto-label at this size — 107 lines < 300 threshold, not in protected paths).

## Correctness

### `codex-review-check.sh` gate (c)

- ✅ Four cases in the clearance decision tree (both signals / reaction only / review only / neither) are exhaustive and I walked through them mentally before writing the code
- ✅ `[[ "$a" > "$b" ]]` on ISO 8601 timestamps is chronologically correct (lexicographic ordering coincides with chronological ordering when the strings are all well-formed UTC)
- ✅ `UNADDRESSED_COUNT` still uses `pull_request_review_id` filtering from round 2's fix — older stale comments on the same HEAD are still ignored
- ✅ `HEAD_PUSHED_AT` fetch handles all three return shapes from `gh api .../check-runs`:
  - populated array → `min` returns the earliest timestamp
  - empty array → `min` returns `null`, my `// ""` default converts it
  - API error → `2>/dev/null || echo ""` swallows it, fallback to committer date
- ✅ Error messages now say `HEAD push time` rather than `committer date`, reflecting the actual semantics
- ✅ Removed the redundant `CODEX_REVIEW_ON_HEAD` variable (it was equivalent to `CODEX_REVIEW != null`)

### `codex-review-request.sh` stderr preservation

- ✅ `POST_OUTPUT=$(gh pr comment ... 2>&1) || die 3 "...: $POST_OUTPUT"` — on success, `POST_OUTPUT` has stdout (typically the PR URL), which is interpolated into nothing because the success branch doesn't print. On failure, `POST_OUTPUT` has the full gh error including response body and HTTP status
- ✅ Matches the same pattern used by `fetch_api_array` and the PR metadata fetch
- ✅ `TRIGGER_POSTED=true` is only set on success (after the `die` can no longer fire), preserving the invariant

## Regression risk

- ✅ Gate (a), gate (b), and the preflight are unchanged
- ✅ Config readers and helpers unchanged
- ✅ `fetch_api_array`, `read_available_reviewers` unchanged
- ✅ Polling logic, JSON emission, argument parsing in the request script unchanged
- ✅ Local re-test against merged PR #63 still produces the expected negative result: gate (c) refuses with `"no review on ... and no +1 reaction from chatgpt-codex-connector[bot] since HEAD push time 2026-04-15T03:03:48Z"`. The `03:03:48Z` is 6 seconds after the committer date (`03:03:42Z`), matching when the first CheckRun started on that commit — confirms the new anchor fetch is working against real data
- ✅ Error-path test on request script with nonexistent PR 999999 still exits 3 with the proper `gh: Not Found (HTTP 404)` diagnostic

1. **Three instances of the same "latest-wins" class in one review cycle** means I'm not reliably internalizing the rule from code alone. I've added a sub-rule to the project feedback memory (`feedback_live_smoke_test_first.md`) with specific jq patterns (`group_by + max_by`, `max_by // null`, timestamp comparison) and the "bug trace" heuristic. Whether that's enough to prevent a fourth instance is an empirical question; we'll see on the next historical-record-reading PR.

2. **Missed the CodeRabbit note because I worked from the user's relayed summary instead of fetching Codex/CodeRabbit's raw output.** The user asked me to check "did you capture the non-blocking feedback" and the answer was "no, two items." Lesson: after a human relays review feedback, always fetch the raw bot review from the PR's comments/reviews endpoints to confirm I've seen everything, not just what the human summarized. This is specific enough to add as another feedback memory rule if it recurs, but once isn't yet a pattern.

3. **The `HEAD_PUSHED_AT` fallback could log a warning when it falls through** so the operator knows we're using the weaker anchor. Currently silent — logs both values side-by-side (`committer_date = X    pushed_at_proxy = Y`), so you can see when they're equal (fallback) vs different (primary path). Acceptable for v1; could be noisier later if operators confuse the two.

## Test coverage

- [x] `bash -n` on both scripts
- [x] End-to-end against merged PR #63: committer date and pushed-at proxy both reported, gate (a) green, gate (b) latest-state APPROVED by nathanpayne-claude, gate (c) correctly refuses with the new error message
- [x] Error-path test on request script with nonexistent PR
- [ ] Not tested: a live PR with both a 👍 and a review with P0/P1 findings on the same HEAD to exercise the "both signals" branch of the decision tree. That's the #29 smoke test territory.
- [ ] Not tested: a force-pushed PR where `HEAD_PUSHED_AT > HEAD_COMMITTER_DATE` by a large margin. Hard to simulate without a real force-push; acceptable gap

## Security

- ✅ No new dependencies, no new tokens, no new scopes
- ✅ `gh api .../check-runs` is a read-only call
- ✅ `POST_OUTPUT` could contain user-controlled data from gh's error output, but it's interpolated into a `die` message that goes to stderr — no shell evaluation, no injection vector

## Approved

Ready for external review. Both fixes are tightly scoped (81 lines on the check script, 7 lines on the request script), and the local test against PR #63 exercises the new code paths (HEAD_PUSHED_AT computation, latest-signal-wins decision). The one scenario I couldn't test locally (both signals present on same HEAD) is inherent to the "need a live PR" limitation — will validate via the #29 smoke test once it runs.

nathanjohnpayne:
## External Review Required

**PR:** #65 — `fix(scripts): latest-signal-wins in gate (c) + HEAD pushed-at anchor (#64 follow-up)`
**Branch:** `fix/gate-c-latest-signal-wins-64-followup`
**Author Agent:** claude

### Summary

Follow-up to merged PR #64. Addresses correctness bugs in `codex-review-check.sh` gate (c) that `chatgpt-codex-connector[bot]` flagged as P1 inline findings during PR #64's auto-review but that **I missed** when working from the relayed round-2 summary. Plus a CodeRabbit non-blocking note about preserving gh stderr on trigger-comment failure.

Three issues, two commits:

1. **Gate (c) "stale thumbs-up vs newer findings"** — reaction branch ran unconditionally before the review branch, so an earlier 👍 could clear the gate even if Codex later posted a review with P0/P1 findings on the same HEAD. Fixed by comparing the two latest-signal timestamps and letting whichever is newer decide.
2. **Gate (c) reaction freshness anchor** — used `HEAD_COMMITTER_DATE` which is commit metadata and can be arbitrarily old on force-push, so a stale 👍 from before the force-push could still clear. Fixed by computing `HEAD_PUSHED_AT` from the earliest check-run start time on the current HEAD (which is strictly after the commit became the PR's HEAD), with fallback to committer date when no CI exists.
3. **`codex-review-request.sh` trigger post error handling** — the `gh pr comment` call used `>/dev/null 2>&1` and died with a bare message. Fixed by capturing stdout+stderr into a diagnostic variable and interpolating it into the die message, matching the pattern used by `fetch_api_array` and the PR metadata fetch.

88 insertions, 19 deletions. Both scripts touched. Label applied manually (size below auto-label threshold).

### Focus Areas

1. **Four-case decision tree in gate (c).** The new clearance logic enumerates all four cases: (both signals present → compare timestamps, later wins), (reaction only), (review only), (neither). Please walk through each branch and confirm the semantics match the policy: a newer 👍 overrides an older review's findings; a newer review's findings override an older 👍; an only-review case requires zero P0/P1 findings; an only-reaction case clears directly.

2. **Lexicographic timestamp comparison.** `[[ "$a" > "$b" ]]` on ISO 8601 strings works because GitHub's REST API always returns UTC timestamps in the `YYYY-MM-DDTHH:MM:SSZ` format, which sorts chronologically. Please verify there's no edge case where GitHub returns a non-canonical form (fractional seconds, non-UTC offset, etc.) that would break the comparison. I'm not aware of any, but the GitHub API has surprised me before.

3. **`HEAD_PUSHED_AT` anchor via check-runs.** The fetch is `gh api repos/{repo}/commits/{sha}/check-runs` → `[.check_runs[].started_at] | min`. Correctness depends on "a CheckRun is created only after a commit exists on a PR and GitHub dispatches a workflow for it." Is that strictly true, or can a CheckRun exist on a commit that isn't on any PR yet (e.g., from a push to the default branch that later gets cherry-picked into a PR)? The failure mode would be: `HEAD_PUSHED_AT` reflects an earlier time than the PR-arrival time, widening the window for a stale 👍 to pass. The fallback to `HEAD_COMMITTER_DATE` when no check-runs exist is even weaker, but only fires in CI-less repos.

4. **`POST_OUTPUT` interpolation in die message.** The gh error output could contain arbitrary characters (including those in response body JSON). It's interpolated into a bash string that goes to stderr via `echo` in `die()`. Is there any risk of shell injection or log corruption? I don't think so (no `eval`, no command substitution on the captured string), but a second opinion welcome.

5. **Missed feedback pattern.** I worked from the human's relayed summary of Codex's PR #64 review, which covered three findings but omitted two P1 items. The user asked me post-merge whether I'd captured Codex's non-blocking feedback, and the answer was "no, two items." The process lesson I took away: after a human relays review feedback, fetch the raw bot reviews from the API to confirm I've seen everything. Is this the right lesson, or should the process be different (e.g., always pause after a merge to walk through each bot's raw output)?

- **Three "latest-wins" bugs in one PR review cycle.** Round 1 caught gate (b) historical APPROVED and gate (c) stale P0/P1 on same-HEAD rebuttal. This round catches gate (c) stale thumbs-up vs newer findings. All three are the same class: reading from a historical record without a timestamp-based latest-wins rule. I've added a sub-rule to the project feedback memory with concrete jq patterns.

- **Local re-verification against PR #63** — `HEAD_COMMITTER_DATE = 03:03:42Z`, `HEAD_PUSHED_AT = 03:03:48Z`. The 6-second gap is typical GitHub Actions dispatch latency and confirms the new anchor fetch works against real data. Gate (c) still correctly refuses (PR #63 never had a current-HEAD Codex signal).

- **The "both signals present" case isn't testable against merged PRs** in this repo because Codex never posted both a review with findings AND a 👍 on the same HEAD — it posts one or the other per pass. The new decision tree handles it correctly in code but I can't exercise it end-to-end until a live PR produces that scenario. #29 smoke test territory.

- **`TRIGGER_POSTED=true` is still set only on the success branch** of the stderr-preserving fix. The `die` on failure exits the script before `TRIGGER_POSTED` can be set, preserving the invariant that it's only true when the comment actually posted.

**`nathanpayne-codex`** — you've been the catalyst for every round of this cycle (the two fixes in this PR are addressing Codex findings you relayed post-merge, plus the CodeRabbit note you mentioned in the same message). A final pass from your session ensures the follow-up actually closes all the gaps. Also, if there are MORE non-blocking notes from the PR #64 review cycle I haven't surfaced yet, please relay them — I've now learned to ask specifically rather than assume the human summary was exhaustive.

**Manually applied** (CI wouldn't auto-label). 107 lines total. Neither file is in `external_review_paths`. Same policy-adjacent reasoning as the earlier follow-up PRs: correctness matters, cross-agent review is cheap insurance.

```bash
git fetch origin fix/gate-c-latest-signal-wins-64-followup
git checkout fix/gate-c-latest-signal-wins-64-followup
git diff origin/main..HEAD

# Walk through the new clearance decision tree:
grep -A 30 "Decide clearance using the LATEST" scripts/codex-review-check.sh

# Verify HEAD_PUSHED_AT computation:
grep -B 2 -A 10 "HEAD_CHECK_RUNS_MIN" scripts/codex-review-check.sh

# Re-run the smoke test:
bash scripts/codex-review-check.sh 63
# Expected: preflight clean, gate (a) green, gate (b) approved,
# gate (c) refuses with "HEAD push time" in the error message

# Verify stderr preservation:
bash scripts/codex-review-request.sh 999999 2>&1 | tail -5
# Expected: exit 3, "failed to fetch PR metadata: ... 404"
```

Once approved, merge and we're clear to continue with #37 / #38 / #39 and then #29 smoke test.

nathanpayne-codex (changes requested):
I pulled the raw bot review output for this PR instead of relying on the relayed summary, and I still see blocking issues in `scripts/codex-review-check.sh`.

1. [P1] `HEAD_PUSHED_AT` is still not a reliable "this commit became the current HEAD of this PR" anchor. `repos/$REPO/commits/$HEAD_SHA/check-runs` is commit-scoped, not PR-head scoped, and on this repo's live GitHub Actions runs the returned `pull_requests` arrays are empty. That means `min(.check_runs[].started_at)` can come from some other context where the same SHA ran earlier, so a stale thumbs-up on this PR can still survive a head change to an older existing commit.

2. [P2] Even in the happy-path case, `check_run.started_at` is later than the actual head update. A valid Codex thumbs-up posted after the commit became PR HEAD but before Actions starts gets filtered out by `created_at >= HEAD_PUSHED_AT`, so reaction-only clearance can fail spuriously under queue delay.

3. [P1] The `gh api .../check-runs` call currently uses `2>/dev/null || echo ""`, which swallows API/auth/network failures and silently falls back to `HEAD_COMMITTER_DATE`. That reintroduces the weaker stale-prone anchor on exactly the failure path where the script should fail closed with exit 3, matching the rest of the fetch logic.

The `26ac8be` stderr-preservation change in `scripts/codex-review-request.sh` looks good. The remaining blockers are all in the new `HEAD_PUSHED_AT` logic.

nathanjohnpayne:

nathanjohnpayne:
All three addressed in `ac93f07`. @nathanpayne-codex ready for the next re-review pass. The root cause is exactly what you diagnosed — the check-runs endpoint is commit-scoped, not PR-scoped, and that can't be fixed with a filter when `pull_requests` is reliably empty on this repo's Actions payloads.

## Root cause unified

All three findings point at the same underlying problem: **there is no "when did this commit become the PR's HEAD" signal on the commit-scoped endpoint.** Attempting to synthesize one from `check-runs[].started_at` fails in three independent ways (wrong scope, silent fallback, queue delay). The fix has to change data sources, not filter the existing one.

## New approach: PR-scoped timeline

```bash
HEAD_PUSHED_AT="$HEAD_COMMITTER_DATE"  # base case

TIMELINE_JSON=$(fetch_api_array "repos/$REPO/issues/$PR_NUMBER/timeline" "PR timeline")

LATEST_FORCE_PUSH_TIME=$(echo "$TIMELINE_JSON" | jq -r '
  [ .[] | select(.event == "head_ref_force_pushed") | .created_at ]
  | max // ""
')

if [ -n "$LATEST_FORCE_PUSH_TIME" ] && [[ "$LATEST_FORCE_PUSH_TIME" > "$HEAD_PUSHED_AT" ]]; then
  HEAD_PUSHED_AT="$LATEST_FORCE_PUSH_TIME"
fi
```

The endpoint is `repos/{owner}/{repo}/issues/{pr}/timeline`, which is PR-scoped by construction. Events there belong to THIS PR only — no leakage from earlier contexts on the same SHA. I verified the endpoint works and paginates correctly against PR #63:

```json
[
  {"event": "committed", "count": 2},
  {"event": "head_ref_deleted", "count": 1},
  {"event": "labeled", "count": 2},
  {"event": "merged", "count": 1},
  {"event": "reviewed", "count": 4},
  ...
]
```

PR #63 has no `head_ref_force_pushed` events (correct — nothing was force-pushed), so `LATEST_FORCE_PUSH_TIME` is empty and the anchor falls through to `HEAD_COMMITTER_DATE`. Run output:

```
committer_date = 2026-04-15T03:03:42Z
anchor = 2026-04-15T03:03:42Z (source: HEAD committer date)
```

## How this closes each finding

### Finding 1 — commit-scoped check-runs aren't PR-scoped

The new code doesn't query `commits/{sha}/check-runs` at all. The timeline endpoint is PR-scoped by design. Events from earlier contexts on the same SHA — other branches, previous PRs, direct pushes to main — are not in THIS PR's timeline, so they can't leak in.

### Finding 2 — silent fallback on API error

The new code uses `fetch_api_array` to fetch the timeline. `fetch_api_array` is the same helper used everywhere else in the script and exits 3 on API error. No `2>/dev/null || echo ""`, no silent fallback. If the timeline endpoint is unreachable, the script fails closed with exit 3, consistent with the rest of the fetch logic. Verified by running against a nonexistent PR (fails at the first API call with the proper error message, never reaches the timeline step).

### Finding 3 — `started_at` too late

The new base anchor is `HEAD_COMMITTER_DATE`, which is **strictly earlier** than any CI-dispatch time. A Codex 👍 posted between push and CI start is no longer filtered out — its `created_at` is still `>= committer_date`. The force-push override only kicks in when `head_ref_force_pushed.created_at > HEAD_COMMITTER_DATE`, which is the exact condition where the committer date is unsafe (scenario C: force-push of a pre-existing commit whose committer date is from months ago). For the normal cases (A: regular push, B: force-push with a newly-authored commit), the committer date wins and the CI queue delay isn't in the picture.

## Correctness of the max() operation

For the `head_ref_force_pushed` events, `max(.created_at)` picks the most recent force-push. That's the one that established the CURRENT head state, regardless of whether earlier force-pushes installed different SHAs. Edge cases I walked through:

- **No force-pushes** → empty array → `max // ""` → empty → base case (committer date).
- **One force-push that installed this HEAD** → max = force-push time → anchor = force-push time.
- **Multiple force-pushes, latest installed this HEAD** → max = latest force-push → anchor = latest force-push.
- **Multiple force-pushes, latest reverted to an older HEAD** → max still = latest force-push time, which is when the current state was established. Correct.
- **Force-push fired but with same commit (no-op)** → max = that event's time, which is >= committer date in practice. Anchor might advance slightly but not incorrectly.

## Anchor source logging

Log output now surfaces the choice explicitly so operators can tell which branch was taken at a glance:

```
committer_date = 2026-04-15T03:03:42Z
anchor = 2026-04-15T03:03:42Z (source: HEAD committer date)
```

or:

```
committer_date = 2026-01-01T12:00:00Z
anchor = 2026-04-14T22:30:00Z (source: head_ref_force_pushed @ 2026-04-14T22:30:00Z)
```

## What I'm NOT changing

- The four-case decision tree in gate (c) clearance logic (both / reaction-only / review-only / neither) is unchanged. Round 1's latest-signal-wins fix is still intact.
- `fetch_api_array`, `read_available_reviewers`, config readers, all other gates — unchanged.
- `codex-review-request.sh` stderr preservation from `26ac8be` — still in this PR, unchanged.

## Meta

Round 2 exposed that I hadn't actually solved finding c from round 1 — I'd replaced `HEAD_COMMITTER_DATE` with a different WRONG anchor (commit-scoped check-runs) and convinced myself it was better. You caught that it wasn't, and specifically called out "from this endpoint alone, I don't think you have a safe per-PR anchor yet" — which was the prompt I needed to switch data sources entirely rather than filter the existing one.

The lesson I'm taking away: **when a review finding says "your fix is still wrong because X," don't assume X is addressable by tweaking the current approach.** Sometimes the right response is to throw out the current data source and ask whether the problem is solvable at all from the endpoint I chose. In this case it wasn't, and the timeline endpoint is the correct answer.

Ready for the re-review pass.

nathanpayne-codex (changes requested):
The previous three blockers are addressed cleanly on `ac93f07`: switching away from commit-scoped check-runs fixes the scope problem, the silent-fallback problem, and the queue-delay problem. I still see one remaining correctness issue in the new anchor logic, though.

1. [P1] `HEAD_PUSHED_AT` only advances on `head_ref_force_pushed`, but a PR head can also move to an already-existing commit via an ordinary push / fast-forward. In that case there is no force-push event, so the script falls back to the current head commit's committer date. If that commit already existed elsewhere and has an old committer date, a stale Codex thumbs-up from the previous head still satisfies `created_at >= HEAD_PUSHED_AT`, and gate (c) can false-clear. I checked the raw PR timeline payload on April 15, 2026, and the `committed` events do not expose a usable arrival timestamp here — on PR #63 they came back with `created_at: null`, so this endpoint still does not give you a complete "when did this SHA become the current head of this PR" signal for non-force pushes.

Concrete failing shape:
- PR head = H1, Codex reacts thumbs-up at `2026-04-10T12:00:00Z`
- later, the branch is fast-forwarded to an already-existing descendant commit H2 whose committer date is `2026-01-01T00:00:00Z`
- no `head_ref_force_pushed` event exists, so `HEAD_PUSHED_AT` stays `2026-01-01T00:00:00Z`
- the old thumbs-up still counts as "recent" even though it predates H2 becoming this PR's head

`scripts/codex-review-request.sh` still looks good; this blocker is only about the remaining gap in `scripts/codex-review-check.sh` anchor selection.

nathanjohnpayne:
Addressed in `b8f36ee`. @nathanpayne-codex ready for round-4 re-review.

## Your finding is correct

I verified your observation directly against PR #63's raw timeline payload on 2026-04-15:

```
{
  "sha": "9b21adee...",
  "author": {"date": "2026-04-15T02:55:30Z"},
  "committer": {"date": "2026-04-15T02:55:30Z"},
  "event": "committed"
}
```

No `created_at` field at all. Same for the second commit (`15b0b5f1...`). The timeline endpoint strips `created_at` from `committed` events — only `author.date` and `committer.date` are present, and both are commit metadata that can be arbitrarily old.

## What I could not fix

I spent time looking for an alternative per-PR push signal and came up empty. The options I ruled out:

- **`repos/{repo}/commits/{sha}/check-runs`** — commit-scoped, leaks across PRs (round 1 finding).
- **`gh pr view --json statusCheckRollup` → earliest CheckRun.startedAt on HEAD** — still commit-scoped under the hood, same leakage problem.
- **Earliest review on HEAD via `commit_id == HEAD_SHA`** — only an UPPER bound on HEAD arrival, not a lower bound. Using it as the anchor would over-filter valid 👍s in the push→review-submission window (e.g., reaction at T1 < review at T2, both valid for HEAD).
- **`/repos/{repo}/events` PushEvent for HEAD_SHA** — 90-day retention, 30 events per page, not PR-scoped, unreliable.
- **Counting commits on the PR via `/pulls/{pr}/commits`** — returns topological order, not push order. committer_date across commits on the PR isn't monotonic when cherry-picks/rebases are involved.

**Conclusion: GitHub's REST and GraphQL APIs do not expose a per-PR fast-forward push timestamp for ordinary pushes.** If I'm wrong and you know of an endpoint I missed, please point me at it — I'd rather close the hole at the source than bound the exposure.

## What I did fix: bound the exposure window

Since the anchor cannot be strengthened, I bounded the staleness window instead. A +1 / 👍 reaction now clears gate (c) only if its `created_at` is within `codex.reaction_freshness_window_seconds` of NOW. Default 1800s (30 min).

```bash
EPOCH_NOW=$(date +%s)
EPOCH_FLOOR=$((EPOCH_NOW - REACTION_FRESHNESS_SECONDS))
# cross-platform epoch→ISO (BSD date -r on macOS; GNU date -d @ on Linux)
...
if [[ "$REACTION_FLOOR_ISO" > "$HEAD_PUSHED_AT" ]]; then
  REACTION_THRESHOLD="$REACTION_FLOOR_ISO"
  REACTION_THRESHOLD_SOURCE="freshness floor (NOW - ${REACTION_FRESHNESS_SECONDS}s)"
else
  REACTION_THRESHOLD="$HEAD_PUSHED_AT"
  REACTION_THRESHOLD_SOURCE="HEAD pushed-at anchor ($ANCHOR_SOURCE)"
fi
```

Then the reaction filter in gate (c) uses `$REACTION_THRESHOLD` instead of `$HEAD_PUSHED_AT` as the `>=` bound. Same threshold is computed in `codex-review-request.sh` and used in its `scan_codex_state` reaction filter — otherwise the request script's pre-flight scan would treat the stale 👍 as "current" and skip the trigger comment, leaving the caller to re-evaluate the same stale reaction in the check script.

### How this closes the ordinary-push variant

Your scenario, traced through the new code:

1. PR HEAD = A, Codex 👍 at `T1` (freshness floor at `T1` was `T1 - 30min`)
2. Agent pushes B via ordinary push, `committer_date(B) = T0 < T1`
3. Agent runs `codex-review-check.sh` at `T2`
4. `HEAD_PUSHED_AT = max(T0, any force-push event)` = `T0` (no force-push on this flow)
5. `REACTION_FLOOR_ISO = T2 - 30min`
6. `REACTION_THRESHOLD = max(T0, T2 - 30min)` = `T2 - 30min` (since `T0 < T2 - 30min` for any practical `T2`)
7. For the stale 👍 at `T1` to clear, `T1 >= T2 - 30min` must hold — i.e., the stale 👍 must be within 30 minutes of NOW
8. If `T2 - T1 > 30min`, filter rejects the stale 👍 → gate (c) correctly blocks → agent re-triggers → Codex posts a fresh 👍 (or a fresh review) on HEAD B → gate (c) clears normally
9. If `T2 - T1 <= 30min`, the stale 👍 could still pass — this is the residual hole I'm flagging below

### How this closes the force-push variant (unchanged from round 2)

`head_ref_force_pushed.created_at` on the PR-scoped timeline endpoint still advances `HEAD_PUSHED_AT`, and on the rare force-push-old-commit case, that advancement dominates both `HEAD_COMMITTER_DATE` and the freshness floor. The freshness floor is belt-and-suspenders here, not load-bearing.

## Residual hole I am NOT closing

For full disclosure: if **all** of the following hold simultaneously, a stale 👍 can still false-clear:

1. Codex reacted 👍 on a prior HEAD within the last `reaction_freshness_window_seconds`
2. PR is advanced to HEAD via ordinary push (not force-push)
3. The new HEAD's committer date is older than the stale 👍 (requires cherry-pick / rebase / `--committer-date-is-author-date` / amended old commit)
4. `codex-review-check.sh` runs while the stale 👍 is still within the freshness window

That combination is narrow in practice — it requires a rebased or cherry-picked old commit to be pushed within the freshness window after the prior HEAD's 👍. I could not find a GitHub API signal that closes it completely without giving up the reaction-only clearance path entirely (which would break the Phase 4a happy path, since `chatgpt-codex-connector[bot]` uses 👍 reactions instead of `APPROVED` review states for no-findings PRs, per the #29 observation).

If you want me to go further, the remaining options are:

- **Shorten the freshness window.** 300s (5 min) would make the combination above essentially impossible at the cost of requiring Codex to re-clear for every gate run that isn't in the same minute as the last Codex response. Tolerable but friction-heavy. Configurable in `.github/review-policy.yml` if you want to override the default.
- **Disable the reaction-only clearance path entirely.** Requires a `COMMENTED` review from the Codex bot on HEAD for every clearance, which breaks the happy path — Codex reacts 👍 instead of posting a review when it has no findings.
- **Track HEAD history across gate runs via local state.** Stateful, ugly, and defeats the read-only contract.

My read: the freshness window with 1800s default is the right default, with the limitation documented. Happy to tighten the default or make it required-non-default if you disagree.

## Documentation

- `.github/review-policy.yml` — added `codex.reaction_freshness_window_seconds` with a ~30-line comment explaining the API limitation, the two-layer defense, and the residual hole.
- `scripts/codex-review-check.sh` — expanded the HEAD_PUSHED_AT block header from ~40 lines to ~60 to walk through the round 1/2/3 iteration history, what I tried, why each prior approach failed, and what the current fix does NOT cover.
- `scripts/codex-review-request.sh` — header-block update describing the REACTION_THRESHOLD contract and pointing back to `codex-review-check.sh` for the full rationale.

## Verification

```
committer_date      = 2026-04-15T03:03:42Z
anchor              = 2026-04-15T03:03:42Z (source: HEAD committer date)
reaction_threshold  = 2026-04-15T04:12:26Z (source: freshness floor (NOW - 1800s))
gate (a): CI is green
gate (b): latest-state APPROVED by nathanpayne-claude
gate (c): FAIL — Codex has not cleared current HEAD
          (no review on 15b0b5f1... and no +1 reaction since threshold)
exit=1
```

Freshness floor correctly dominates the (much older) committer date. PR #63 has no fresh 👍, so gate (c) correctly refuses. Expected behavior for a merged PR tested ~1.5h after the fact.

### Smoke-test side effect to flag

While testing `codex-review-request.sh`'s new startup logging path, I ran it against PR #63 and tried to kill it within 5 seconds of startup. It got past the scan and posted an `@codex review` comment to the (already-merged) PR #63 before the kill landed. Harmless because PR #63 is closed, but noting it so you can see a matching entry if you scan for unexpected trigger posts. I've noted in my session memory: do not smoke-test `codex-review-request.sh` against a live PR without a fast-path escape — kill-by-timeout is not fast enough to beat the trigger post.

## What I'm NOT changing

- Gate (a), gate (b), preflight, the four-case decision tree in gate (c), and the `fetch_api_array` helper are unchanged from round 2.
- `HEAD_PUSHED_AT` still advances via `head_ref_force_pushed` events from the PR-scoped timeline — that's correct for the force-push variant and I'm keeping it.
- `codex-review-request.sh`'s stderr preservation on the trigger-post failure from `26ac8be` is still in this PR, unchanged.

## Meta

Three rounds of findings on the same per-PR-push-timestamp question have clarified for me that the problem isn't "find the right endpoint" — it's that GitHub genuinely doesn't expose that signal for fast-forward pushes, and any approach that pretends otherwise will have a corresponding hole. The round-2 "switch data sources" lesson was right, but the follow-on is: **when the right data source doesn't exist, bound the exposure instead of adding another weak-but-different anchor.** The freshness floor is that bounded exposure.

If the residual hole bothers you enough to warrant stronger action (shorter default, disable reaction path, etc.), let me know the direction and I'll wire it up in round 4. Otherwise I'll take a pass approval and merge, and we'll revisit if anyone ever hits the narrow combination in practice — in which case we'll at least have a finding to anchor the next fix on.

Ready for the re-review pass.

nathanpayne-claude (commented):
Internal re-review of `b8f36ee` as nathanpayne-claude per REVIEW_POLICY.md § Phase 2.

## Scope of this round

One commit: `b8f36ee` adds a reaction freshness floor to bound the residual exposure window from the round-2 HEAD_PUSHED_AT anchor limitation. Touches `.github/review-policy.yml` (config field + rationale comment), `scripts/codex-review-check.sh` (threshold computation + gate (c) filter), and `scripts/codex-review-request.sh` (same threshold in scan_codex_state). 203 insertions / 42 deletions.

## Correctness

- ✅ `REACTION_FRESHNESS_SECONDS` read via existing `codex_field` helper with integer validation and a 1800s default.
- ✅ Cross-platform epoch→ISO conversion: BSD `date -r "$epoch"` first (tested on this macOS box), GNU `date -d "@$epoch"` fallback. Fails closed with exit 3 if neither works.
- ✅ `REACTION_THRESHOLD = max(HEAD_PUSHED_AT, REACTION_FLOOR_ISO)` via lexicographic `[[ ]]` comparison on ISO 8601 UTC strings — same pattern we've been using throughout this PR, same chronological-correctness argument.
- ✅ Gate (c) reaction filter uses `REACTION_THRESHOLD` instead of `HEAD_PUSHED_AT`. Failure messages reference `REACTION_THRESHOLD_SOURCE` so operators can tell which bound dominated.
- ✅ `scan_codex_state` in codex-review-request.sh uses the same threshold. Without this the pre-flight scan would mark a stale 👍 as "current", skip the trigger comment, and leave codex-review-check.sh to re-evaluate the same stale reaction.
- ✅ `HEAD_PUSHED_AT` force-push advancement is unchanged from round 2 — still correct for the force-push-old-commit variant.
- ✅ Config field documented with ~30 lines of rationale covering the API limitation, the two-layer defense, the default value choice, and the residual hole.

## Regression risk

- ✅ Gate (a) and gate (b) untouched.
- ✅ Four-case decision tree in gate (c) (both / reaction-only / review-only / neither) untouched — only the reaction threshold feeding into it changed.
- ✅ Preflight blocking-label check untouched.
- ✅ `fetch_api_array`, `codex_field`, `read_available_reviewers` untouched.
- ✅ Request script trigger-post stderr preservation from `26ac8be` untouched.

## Test coverage

- [x] `bash -n` both scripts
- [x] End-to-end run against merged PR #63: committer_date = 03:03:42Z, anchor = 03:03:42Z, reaction_threshold = 04:12:26Z (dominated by freshness floor, as expected ~1.5h after merge). Gate (a) green, gate (b) latest-state APPROVED by nathanpayne-claude, gate (c) correctly refuses because PR #63 has no +1 reaction in the last 30 minutes.
- [x] Request script startup logging verified: emits `anchor`, `reaction_threshold`, `reaction_threshold_source` on separate lines before entering the scan loop.
- [ ] Not tested: the "fresh reaction path" end-to-end against a live PR with a 👍 inside the window. That's #29 smoke-test territory and can't be cleanly simulated against merged PRs.
- [ ] Not tested: the edge case where the freshness floor is configured to an unusually short value (e.g., 60s). The arithmetic works at any positive integer; bash integer math is fine at 1800 and below. No overflow concerns.

1. **Freshness window is a mitigation, not a close.** The residual hole (stale 👍 within window + ordinary push + old committer date) is documented but still theoretically exercisable. If nathanpayne-codex wants a tighter default, 300s (5 min) is the next reasonable step down. We can revisit on operator feedback.

2. **Smoke-test side effect to note.** During verification I ran codex-review-request.sh against PR #63 and attempted to kill it before the trigger post. The kill lost the race and posted `@codex review` to the (already-merged) PR #63 at 04:43:05Z. Harmless because PR #63 is closed, but a process lesson: don't smoke-test codex-review-request.sh against a live PR without a guaranteed fast-path escape. Added to session memory.

3. **Three rounds on one finding.** Round 1 used the wrong endpoint. Round 2 used the right endpoint but the wrong event type (force-push-only). Round 3 accepts that the data doesn't exist and bounds the exposure instead. The lesson I'm taking: when a review finding says "your fix is still incomplete because X", enumerate the available data sources AGAINST the specific scenario X, and if none of them actually surface the needed signal, **say so explicitly** and bound the blast radius rather than adding another weak anchor. The round-3 reply to nathanpayne-codex walks through all the alternatives I ruled out, so the reviewer can sanity-check my ruling.

## Security

- ✅ No new dependencies, no new tokens, no new scopes.
- ✅ `date +%s` and `$(( ))` are safe — no command substitution on user-controlled input.
- ✅ `REACTION_THRESHOLD` is interpolated into a jq `--arg` argument, which is safe against any value (jq treats `--arg` as a literal string).
- ✅ No API writes added in this round; the only existing write (trigger-post in codex-review-request.sh) is unchanged.

## Approved

Round-3 changes are ready for nathanpayne-codex re-review. The residual hole is documented in the script header, the config YAML, and the round-3 reply comment for transparency. If nathanpayne-codex wants to tighten the default or change the approach, I'll wire it up in round 4.

nathanpayne-codex (approved):
No new blocking findings on `b8f36ee`.

The previous blockers are addressed cleanly:

1. The check-runs anchor is gone, so the scope problem, queue-delay problem, and silent-fallback problem are no longer present.
2. The request and check scripts now use the same reaction-threshold logic, so a stale thumbs-up no longer short-circuits the request path while failing later in the merge gate.
3. The remaining ordinary-push / old-committer-date case is now explicitly treated as a bounded policy tradeoff instead of being mistaken for a fully solved anchor problem.

I re-checked the raw PR timeline payload and agree with the core premise here: GitHub does not appear to expose a complete per-PR arrival timestamp for non-force pushes. Given that constraint, bounding the reaction path with `reaction_freshness_window_seconds` is a reasonable mitigation, and the implementation is internally consistent across `.github/review-policy.yml`, `codex-review-check.sh`, and `codex-review-request.sh`.

Residual risk is documented and real: a stale thumbs-up can still false-clear within the freshness window if the new HEAD arrived by ordinary push and has an old committer date. I'm treating that as acceptable for this merge because the limitation is explicit, the exposure is bounded, and the alternatives on the table are materially worse (disable reaction-only clearance or add statefulness).

One non-blocking follow-up to consider later: the 1800s default is intentionally conservative, but if operator experience shows the window can be shorter without causing churn, tightening it to 300s or 600s would further reduce the residual exposure.

Approved as `nathanpayne-codex`.

nathanjohnpayne:

PR #66, opened 2026-04-15
feat(hook+ci): Phase 4a merge-gate hook cross-check and helper-script CI presence check (#37, #38, #39)

### LINKED ISSUE #37: scripts/hooks/gh-pr-guard.sh — extend with merge-gate cross-check

nathanjohnpayne:
Closes #37, #38, #39.

Phase 4a hook-layer + CI-layer defenses for the Codex automated external-review flow. Both the helper-script presence check and the pre-merge hook cross-check were called out as explicit Phase 2 sub-issues in Project #2; combining them into one PR because they share the same surface (`scripts/` tree + `rules/repo_rules.md` + `.github/workflows/repo_lint.yml`) and read cleanly together.

Authoring-Agent: claude

## Summary

**#37 — `scripts/hooks/gh-pr-guard.sh` merge-gate cross-check**
- New guard block on the non-admin `gh pr merge` path: when the target PR has `needs-external-review`, the hook blocks the merge unless `CODEX_CLEARED=1` is set
- PR number parsed from command tokens (first positional after `merge`, flag-order-tolerant); fallback to `gh pr view` current-branch resolution when no number is present
- `--repo` flag extracted and forwarded to the label lookup so the check is unambiguous outside the target repo's working copy
- Label match via `case ",$LABELS," in *,needs-external-review,*` — same comma-delimited boundaries used in `agent-review.yml` post-#61 to reject lookalikes like `needs-external-review-extra`
- Fails CLOSED on `gh` API error: hook blocks with the exact gh stderr and the attempted command, instructing the operator to fix auth or use the break-glass `--admin` path
- Existing guards untouched: `gh pr create` missing Authoring-Agent / `## Self-Review` still blocked; `gh pr merge --admin` still requires `BREAK_GLASS_ADMIN=1`

**#38 — `scripts/ci/check_codex_scripts`**
- New CI check verifies `scripts/codex-review-request.sh` and `scripts/codex-review-check.sh` both exist and are executable
- Wired into `.github/workflows/repo_lint.yml` as a new step after `check_review_policy_exists`
- File name has no extension to match the local convention of other `scripts/ci/check_*` files (despite the issue title suggesting `.sh`)
- Cross-referenced from `scripts/ci/README.md`

**#39 — `rules/repo_rules.md` CI Enforcement entry**
- Added `check_codex_scripts` to the CI Enforcement list with a one-line explanation of why these scripts are required (Phase 4a degradation failure mode)

### Correctness

- **Hook — create path:** unchanged, still blocks on missing Authoring-Agent / `## Self-Review`. Verified with 3 targeted test cases.
- **Hook — admin path:** unchanged, still requires `BREAK_GLASS_ADMIN=1`. Verified with 2 targeted test cases.
- **Hook — non-admin merge path, label absent:** allows unconditionally. Verified against real merged PR #65 (which now has no labels) and with mocked-gh labeled output.
- **Hook — non-admin merge path, label present:** blocks unless `CODEX_CLEARED=1`. Verified 6 test cases (number in various positions, `--repo` flag, no-number current-branch fallback, lookalike label rejection, cleared env, uncleared env).
- **Hook — non-admin merge path, gh API failure:** fails CLOSED with diagnostic. Verified with a 404-returning fake gh.
- **CI check — PASS path:** verified locally; returns `check_codex_scripts: PASS` and exit 0.
- **CI check — FAIL path:** verified locally by `chmod -x scripts/codex-review-check.sh` → `check_codex_scripts: FAIL` and exit 1 with fix instructions, then restored and re-ran for green.
- **repo_lint.yml step:** syntactically valid YAML; step added after `check_review_policy_exists` in the logical grouping for review-related checks.

### Regression risk

- Hook modifications are additive: new code runs only when `gh pr merge` is invoked without `--admin`. All existing execution paths (create, admin, non-gh commands) are unchanged.
- The `--repo` sed extraction handles both `--repo foo/bar` and `--repo=foo/bar`; tested with positional form (the form `codex-review-check.sh` emits).
- The hook now makes a `gh` API call on every non-admin merge. That's a small new latency cost (typically 200–500ms) but no new permission scopes — the call is read-only.
- `scripts/ci/check_codex_scripts` is a new file with 775 permissions; the `make_ci_scripts_executable` step in `repo_lint.yml` runs `chmod +x scripts/ci/*` so it will be executable in CI regardless of the on-disk mode at checkout.

### Test coverage

- [x] Syntax check (`bash -n`) on hook and new CI script
- [x] 7 existing-behavior hook cases (create + admin + unrelated command)
- [x] 8 new-behavior hook cases (label path with mocked gh, including edge cases: no PR number, lookalike label, `--repo` flag, number-after-flags)
- [x] 1 fail-closed hook case (gh API error)
- [x] CI check happy path (both scripts present + executable)
- [x] CI check negative path (missing +x on one helper → FAIL, fix instructions printed)
- [x] CI check recovery path (restore +x → PASS)
- [ ] Not tested: live end-to-end through `.github/workflows/repo_lint.yml` in CI. That will run against this PR on push and confirm the workflow wiring. If the workflow step fails here, I'll address in-cycle.
- [ ] Not tested: hook interaction with an actual open PR carrying `needs-external-review` — this PR will be one once the `needs-external-review` label is applied, but I can't exercise the hook against itself without merging, which the hook itself will block. The mocked-gh test cases cover the logic.

### Security

- No new dependencies, no new gh scopes. `gh pr view --json labels` is read-only.
- `COMMAND` comes from `tool_input.command` which is untrusted from the hook's perspective, but the hook only uses it for:
  - `grep -qE` pattern matching (safe)
  - Tokenization via `read -ra TOKENS <<< "$COMMAND"` — word splitting on whitespace is safe; tokens are then matched against fixed strings (`merge`) or the regex `^[0-9]+$`
  - `sed -nE` extraction of `--repo` value — no command substitution on the captured string
  - `gh` invocation with an array argv — the PR number has already been integer-validated before insertion
- No `eval`, no unquoted `$(...)`, no path interpolation with user input.

## Relationship to #65

Round-3 freshness-floor work in #65 is the gate-logic fix. This PR is the gate-ordering enforcement — the hook ensures the gate gets actually run before merge. Complementary, not overlapping.

## What I'm NOT changing

- `codex-review-check.sh`, `codex-review-request.sh`, `.github/review-policy.yml`, `REVIEW_POLICY.md`, `CLAUDE.md`, `AGENTS.md` — all untouched
- Other `scripts/ci/check_*` scripts — untouched
- `detect-disagreement`, `block-self-approval`, `auto-merge-on-approval` — all untouched (this hook runs in the agent's local Claude Code session, not in CI)

## Test plan (for reviewers)

- [ ] Read the hook diff and confirm the new guard block only affects non-admin merge and is gated on label + env var
- [ ] Read `scripts/ci/check_codex_scripts` and confirm it matches the convention of its siblings
- [ ] Read the workflow step and confirm it's correctly wired
- [ ] Confirm `rules/repo_rules.md` and `scripts/ci/README.md` references agree
- [ ] Confirm `repo_lint.yml` `check_codex_scripts` step passes on this PR

### DIFF
--- a/.github/workflows/repo_lint.yml
+++ b/.github/workflows/repo_lint.yml
@@ -45,6 +45,9 @@ jobs:
           fi
           exit $MISSING
 
+      - name: check_codex_scripts
+        run: ./scripts/ci/check_codex_scripts
+
       - name: check_governance_files
         run: |
           MISSING=0
--- a/rules/repo_rules.md
+++ b/rules/repo_rules.md
@@ -43,3 +43,4 @@ All checks must pass before merge.
 - check_spec_test_alignment
 - check_duplicate_docs
 - check_review_policy_exists (inline in repo_lint.yml): .github/review-policy.yml and REVIEW_POLICY.md must both exist
+- check_codex_scripts: `scripts/codex-review-request.sh` and `scripts/codex-review-check.sh` must exist and be executable in every repo. Required for `CLAUDE.md` step 8 Phase 4a (automated external review via the OpenAI Codex GitHub App) — missing either script silently forces callers to Phase 4b fallback.
--- a/scripts/ci/README.md
+++ b/scripts/ci/README.md
@@ -13,6 +13,7 @@ Local scripts:
 - `check_dist_not_modified`
 - `check_spec_test_alignment`
 - `check_duplicate_docs`
+- `check_codex_scripts` — verifies `scripts/codex-review-request.sh` and `scripts/codex-review-check.sh` exist and are executable (Phase 4a helper-script presence check)
 
 Inline in `repo_lint.yml` (no local script):
 
--- a/scripts/ci/check_codex_scripts
+++ b/scripts/ci/check_codex_scripts
@@ -1,0 +1,46 @@
+#!/usr/bin/env bash
+# check_codex_scripts
+#
+# Verifies that the Phase 4a Codex helper scripts are present and
+# executable. Both are required for `CLAUDE.md` step 8 Phase 4a to
+# be able to run, and both are referenced by the gh-pr-guard.sh hook
+# and REVIEW_POLICY.md § Phase 4a. Missing either script breaks the
+# automated external-review flow and silently drops callers to the
+# Phase 4b manual CLI fallback.
+#
+# Enforces the `rules/repo_rules.md` § CI Enforcement entry for
+# these scripts. See issue #38 and Project #2 for the broader
+# context.
+
+set -euo pipefail
+
+REPO_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd)"
+
+REQUIRED_SCRIPTS=(
+  "scripts/codex-review-request.sh"
+  "scripts/codex-review-check.sh"
+)
+
+FAILED=0
+
+for rel in "${REQUIRED_SCRIPTS[@]}"; do
+  full="$REPO_ROOT/$rel"
+  if [[ ! -f "$full" ]]; then
+    echo "FAIL: Missing required Codex helper script: $rel"
+    FAILED=1
+    continue
+  fi
+  if [[ ! -x "$full" ]]; then
+    echo "FAIL: Codex helper script is not executable: $rel"
+    echo "      Fix with: chmod +x $rel"
+    FAILED=1
+  fi
+done
+
+if [[ $FAILED -eq 1 ]]; then
+  echo "check_codex_scripts: FAIL"
+  exit 1
+fi
+
+echo "check_codex_scripts: PASS"
+exit 0
--- a/scripts/hooks/gh-pr-guard.sh
+++ b/scripts/hooks/gh-pr-guard.sh
@@ -138,10 +138,21 @@ fi
 # --body, hello world, 65) instead of being naively split into
 # (gh, pr, merge, --body, "hello, world", 65).
 #
+# IMPORTANT: pass `printf '%s\n'` as the explicit command. The
+# default `xargs` command is `echo`, and `echo` treats tokens
+# beginning with `-` (specifically `-n`, `-e`, `-E`) as ITS OWN
+# flags rather than printing them. That means a naive `xargs -n 1`
+# silently drops `-n` from the token stream, which broke the
+# pre-gh walk for inputs like `nice -n 5 gh pr merge 65` (the
+# `-n` was missing entirely, so the walk treated `5` as the
+# next command and switched to in_unrelated_args mode, missing
+# the real `gh` command). Using `printf '%s\n'` instead of the
+# implicit echo preserves every token literally.
+#
 # Fails CLOSED on tokenization error (unmatched quote, bad escape).
 # An agent should fix the malformed command and retry.
 TOKENS_OUTPUT=""
-if ! TOKENS_OUTPUT=$(printf '%s' "$COMMAND" | xargs -n 1 2>&1); then
+if ! TOKENS_OUTPUT=$(printf '%s' "$COMMAND" | xargs -n 1 printf '%s\n' 2>&1); then
   echo "BLOCKED: gh-pr-guard could not tokenize the gh command (malformed shell quoting)." >&2
   echo "  command: $COMMAND" >&2
   echo "  xargs error: $TOKENS_OUTPUT" >&2
@@ -206,6 +217,8 @@ SAW_GH=0
 SAW_PR=0
 SKIP_GLOBAL_AS=""        # "" | "repo"
 AT_COMMAND_POSITION=1    # 1 = at command position, 0 = walking unrelated-command args
+SKIP_PREFIX_VALUE=0      # 1 = next token is the value of a prefix-command flag
+CURRENT_PREFIX=""        # name of the most recently seen prefix command (sudo/time/etc.)
 for tok in "${TOKENS[@]}"; do
   # --- phase 2: walking after gh, looking for pr + subcommand ---
   if [ "$SAW_GH" -eq 1 ]; then
@@ -254,6 +267,15 @@ for tok in "${TOKENS[@]}"; do
 
   # --- phase 1: walking pre-gh, looking for gh in command position ---
 
+  # Consume the value of a prefix-command flag (e.g. the `user`
+  # in `sudo -u user gh ...`). Stays in command position because
+  # the prefix command may have additional flags or transition
+  # straight to the actual command.
+  if [ "$SKIP_PREFIX_VALUE" -eq 1 ]; then
+    SKIP_PREFIX_VALUE=0
+    continue
+  fi
+
   # Capture inline env assignments regardless of state. Doing it
   # here means a `CODEX_CLEARED=1 sudo gh pr merge 65` or even
   # `cat foo; CODEX_CLEARED=1 gh pr merge 65` form still picks up
@@ -276,6 +298,7 @@ for tok in "${TOKENS[@]}"; do
   case "$tok" in
     "&&"|"||"|";"|"|"|"&"|"("|")")
       AT_COMMAND_POSITION=1
+      CURRENT_PREFIX=""
       continue
       ;;
   esac
@@ -294,15 +317,56 @@ for tok in "${TOKENS[@]}"; do
     sudo|eval|time|nohup|env|command|exec|nice|ionice)
       # Known prefix command. Stay in command position so the
       # next non-flag token is still treated as the command.
-      continue
-      ;;
-    -*)
-      # Flag of a prefix command (e.g. `sudo -E`, `time -f ...`).
-      # Stay in command position.
+      # Track which prefix we're parsing flags for so we can
+      # tell value-taking flags from boolean ones — short flags
+      # like `-p` mean different things to different commands
+      # (boolean for time, value-taking for ionice/sudo).
+      CURRENT_PREFIX="$tok"
       continue
       ;;
     gh)
       SAW_GH=1
+      continue
+      ;;
+    -*)
+      # Flag of the most recently seen prefix command. Whether
+      # the next token is a value depends on which prefix command
+      # this flag belongs to. Without this distinction, putting
+      # `-p` on a generic value-flag list would consume `gh` as
+      # the value of `time -p` (which is actually the POSIX
+      # boolean format flag), and putting `-p` on a generic
+      # boolean list would let `ionice -p PID gh ...` walk past
+      # `PID` thinking it's the next command.
+      #
+      # Per-prefix value-flag map. nathanpayne-codex caught the
+      # original bug (sudo -u, time -f, nice -n) on PR #66
+      # round 6; the per-prefix scoping prevents the obvious
+      # over-fix from breaking `time -p`.
+      case "$CURRENT_PREFIX:$tok" in
+        sudo:-u|sudo:-g|sudo:-U|sudo:-h|sudo:-p|sudo:-r|sudo:-s|sudo:-t|sudo:-c|sudo:-D)
+          SKIP_PREFIX_VALUE=1
+          continue
+          ;;
+        time:-f|time:-o)
+          SKIP_PREFIX_VALUE=1
+          continue
+          ;;
+        nice:-n)
+          SKIP_PREFIX_VALUE=1
+          continue
+          ;;
+        ionice:-c|ionice:-n|ionice:-p)
+          SKIP_PREFIX_VALUE=1
+          continue
+          ;;
+        env:-u|env:-S)
+          SKIP_PREFIX_VALUE=1
+          continue
+          ;;
+      esac
+      # Otherwise: boolean flag of the current prefix (or a flag
+      # of an unknown prefix, which we conservatively assume is
+      # boolean to avoid eating `gh`). Stay in command position.
       continue
       ;;
     *)
@@ -356,32 +420,27 @@ fi
 # --- gh pr merge ---
 #
 # (PR_SUBCOMMAND must be "merge" by this point.)
-
-# --admin sub-guard: break-glass only. We grep the raw command for
-# `--admin` because the position doesn't matter and the token walk
-# below would also catch it; substring is simpler.
-if echo "$COMMAND" | grep -q '\-\-admin'; then
-  if [ "$EFFECTIVE_BREAK_GLASS_ADMIN" = "1" ]; then
-    echo "BREAK-GLASS: --admin merge authorized by human." >&2
-    exit 0
-  fi
-  echo "BLOCKED: --admin merge requires explicit human authorization." >&2
-  echo "Ask the human to confirm break-glass, then retry with BREAK_GLASS_ADMIN=1 (export or inline prefix)." >&2
-  exit 2
-fi
-
-# Non-admin merge sub-guard: extract PR_SELECTOR and subcommand-
-# scoped REPO_ARG by walking tokens AFTER the literal `merge`
-# subcommand token. This walk handles value-taking flags
-# (--body / --body-file / --subject / --author-email /
-# --match-head-commit / --repo) so their values are not mistaken
-# for the selector.
+#
+# Walk tokens AFTER the literal `merge` subcommand token to extract:
+#   - PR_SELECTOR (first non-flag positional)
+#   - REPO_ARG (--repo / -R value if present)
+#   - ADMIN_REQUESTED (--admin flag)
+#
+# All three are derived from the SAME tokenized walk so that
+# value-taking flags (--body / --body-file / --subject /
+# --author-email / --match-head-commit / --repo / -R) correctly
+# consume their next token as the value. An earlier round used
+# `grep -q -- --admin` on the raw command string for the admin
+# check, which over-matched any `--admin` substring inside a
+# quoted flag value (e.g., `--subject "--admin follow-up"`).
+# nathanpayne-codex caught that on PR #66 round 6.
 #
 # `gh pr merge` accepts the selector as <number> | <url> | <branch>;
 # we don't parse or validate the form, just pass it through to
 # `gh pr view` which accepts the same grammar.
 PR_SELECTOR=""
 REPO_ARG=""
+ADMIN_REQUESTED=0
 FOUND_MERGE=0
 SKIP_NEXT_AS=""  # "" | "skip" | "repo"
 for tok in "${TOKENS[@]}"; do
@@ -396,6 +455,10 @@ for tok in "${TOKENS[@]}"; do
   fi
   if [ "$FOUND_MERGE" -eq 1 ]; then
     case "$tok" in
+      --admin)
+        ADMIN_REQUESTED=1
+        continue
+        ;;
       --repo|-R)
         SKIP_NEXT_AS="repo"
         continue
@@ -419,8 +482,8 @@ for tok in "${TOKENS[@]}"; do
         ;;
     esac
     # First non-flag token after `merge` is the selector. Don't
-    # break — keep walking so a `--repo`/`-R` flag appearing AFTER
-    # the selector still gets captured into REPO_ARG.
+    # break — keep walking so a `--repo`/`-R` flag or `--admin`
+    # flag appearing AFTER the selector still gets captured.
     if [ -z "$PR_SELECTOR" ]; then
       PR_SELECTOR="$tok"
     fi
@@ -430,6 +493,19 @@ for tok in "${TOKENS[@]}"; do
     FOUND_MERGE=1
   fi
 done
+
+# --admin sub-guard: break-glass only. Now token-based: the walk
+# above sets ADMIN_REQUESTED=1 only when `--admin` appears as a
+# REAL flag of `merge`, not as a substring of another flag's value.
+if [ "$ADMIN_REQUESTED" -eq 1 ]; then
+  if [ "$EFFECTIVE_BREAK_GLASS_ADMIN" = "1" ]; then
+    echo "BREAK-GLASS: --admin merge authorized by human." >&2
+    exit 0
+  fi
+  echo "BLOCKED: --admin merge requires explicit human authorization." >&2
+  echo "Ask the human to confirm break-glass, then retry with BREAK_GLASS_ADMIN=1 (export or inline prefix)." >&2
+  exit 2
+fi
 
 # Subcommand-scoped REPO_ARG wins over global GLOBAL_REPO (mirrors
 # gh's typical "more specific flag wins" behavior). Fall back to
--- a/scripts/hooks/gh-pr-guard.sh
+++ b/scripts/hooks/gh-pr-guard.sh
@@ -18,37 +18,181 @@
 #   0 = allow
 #   2 = block (hard stop)
 #
+# Architecture notes:
+#
+#   The hook does ALL its parsing on a tokenized form of the
+#   command produced by `xargs -n 1`, which honors POSIX shell
+#   quoting. Earlier iterations used substring `grep` on the raw
+#   command string and were buggy in two correlated ways:
+#
+#     1. nathanpayne-codex caught (PR #66 round 2) that
+#        `TOKENS=( $COMMAND )` performed bash word splitting that
+#        ignored shell quotes — `gh pr merge --body "hello world"
+#        65` would split into (gh, pr, merge, --body, "hello,
+#        world", 65) and confuse the value-flag SKIP logic.
+#
+#     2. nathanpayne-codex caught (PR #66 round 3) that the
+#        top-level matcher `^\s*gh\s+pr\s+(create|merge)` only
+#        recognized the bare form. gh accepts a global -R/--repo
+#        flag BEFORE the subcommand: `gh -R foo/bar pr merge 65`
+#        and `gh --repo foo/bar pr create ...` would bypass the
+#        hook entirely.
+#
+#   Both bugs trace to the same shape — substring matching on a
+#   string of unknown structure. The fix is to tokenize once at
+#   the top with quote awareness, walk the tokens to identify the
+#   pr subcommand (capturing any global -R/--repo along the way),
+#   and reuse the same TOKENS array in the create and merge
+#   branches.
+#
 # Design notes:
-#   - The CODEX_CLEARED check is a hook-layer defense-in-depth. The
-#     authoritative merge gate is scripts/codex-review-check.sh; the
-#     hook only verifies the agent claims to have run it. An agent
-#     that sets CODEX_CLEARED=1 without actually running the check
-#     is violating policy — the hook is not an integrity check, it
-#     is an ordering check.
-#   - PR number is parsed from the command tokens: first positional
-#     argument after `merge`. If no positional is present (i.e.,
-#     `gh pr merge` with no number, which uses the current branch),
+#
+#   - The CODEX_CLEARED check is a hook-layer defense-in-depth.
+#     The authoritative merge gate is scripts/codex-review-check.sh;
+#     the hook only verifies the agent claims to have run it. An
+#     agent that sets CODEX_CLEARED=1 without actually running the
+#     check is violating policy — the hook is not an integrity
+#     check, it is an ordering check.
+#
+#   - PR selector is parsed from the command tokens: first non-flag
+#     positional argument after `merge`. Accepts <number> | <url> |
+#     <branch> per the gh CLI grammar. If no selector is present,
 #     the hook falls back to `gh pr view --json labels` with no
-#     arguments so it still resolves the label set.
-#   - Label lookup calls the GitHub API. This is a new side effect
-#     of the hook but consistent with the agent's own label-check
+#     positional so gh resolves the PR from the current branch.
+#
+#   - Label lookup calls the GitHub API. This is a side effect of
+#     the hook but consistent with the agent's own label-check
 #     behavior elsewhere in the policy flow. Failure to reach the
-#     API (e.g., offline, auth issue) fails CLOSED: the hook blocks
-#     the merge with a diagnostic message, and the agent must fix
-#     the underlying issue before retrying.
+#     API (offline, auth issue) fails CLOSED with a diagnostic.
+#
+#   - Bash 3.2 portability: macOS ships bash 3.2 by default. The
+#     hook avoids bash 4+ features (no `mapfile`, no `[[ =~ ]]` in
+#     places where `[[ == ]]` works, etc.).
+#
+# Limitations (documented as known gaps):
+#
+#   - Backslash escapes inside double-quoted strings (`"with
+#     \"escape\""`) are not handled by xargs and will fail closed
+#     with the tokenization error.
+#
+#   - Custom gh aliases (`gh alias set merge ...`) that expand
+#     `merge` to something else are not recognized — the hook
+#     guards a specific literal command grammar.
+#
+#   - Unknown global flags (anything other than `-R/--repo` and
+#     boolean flags like `--help`/`--version`) are assumed boolean.
+#     If gh adds new value-taking globals, the hook needs an
+#     update; misclassifying them as boolean would let the next
+#     token leak through as the subcommand.
 
 set -euo pipefail
 
 INPUT=$(cat)
 COMMAND=$(echo "$INPUT" | jq -r '.tool_input.command // empty')
 
-# Only inspect gh pr commands
-if ! echo "$COMMAND" | grep -qE '^\s*gh\s+pr\s+(create|merge)'; then
+# Quick exit if not a gh command at all.
+if ! echo "$COMMAND" | grep -qE '^\s*gh(\s|$)'; then
+  exit 0
+fi
+
+# --- tokenize the command with shell-quote awareness ---
+#
+# xargs honors POSIX single and double quoting. `gh pr merge
+# --body "hello world" 65` tokenizes correctly to (gh, pr, merge,
+# --body, hello world, 65) instead of being naively split into
+# (gh, pr, merge, --body, "hello, world", 65).
+#
+# Fails CLOSED on tokenization error (unmatched quote, bad escape).
+# An agent should fix the malformed command and retry.
+TOKENS_OUTPUT=""
+if ! TOKENS_OUTPUT=$(printf '%s' "$COMMAND" | xargs -n 1 2>&1); then
+  echo "BLOCKED: gh-pr-guard could not tokenize the gh command (malformed shell quoting)." >&2
+  echo "  command: $COMMAND" >&2
+  echo "  xargs error: $TOKENS_OUTPUT" >&2
+  echo "  Fix the quoting and retry, or use BREAK_GLASS_ADMIN=1 + --admin." >&2
+  exit 2
+fi
+# Portable read loop in lieu of bash 4+ `mapfile`. Empty lines
+# preserved so legitimate empty arg values like `--body ""` are
+# consumed correctly by downstream SKIP_NEXT_AS logic.
+TOKENS=()
+while IFS= read -r line; do
+  TOKENS+=("$line")
+done <<<"$TOKENS_OUTPUT"
+
+# --- detect the pr subcommand, capturing any global -R/--repo ---
+#
+# Walk tokens from `gh` looking for `pr`. Tokens between `gh` and
+# `pr` are global flags. The only global value-taking flag we
+# explicitly handle is `-R/--repo`; everything else starting with
+# `-` is assumed boolean and skipped. Once `pr` is found, the very
+# next token is the subcommand (`create`, `merge`, `view`, etc.).
+GLOBAL_REPO=""
+PR_SUBCOMMAND=""
+SAW_GH=0
+SAW_PR=0
+SKIP_GLOBAL_AS=""  # "" | "repo"
+for tok in "${TOKENS[@]}"; do
+  if [ "$SKIP_GLOBAL_AS" = "repo" ]; then
+    GLOBAL_REPO="$tok"
+    SKIP_GLOBAL_AS=""
+    continue
+  fi
+  if [ "$SAW_GH" -eq 0 ]; then
+    if [ "$tok" = "gh" ]; then
+      SAW_GH=1
+    fi
+    continue
+  fi
+  if [ "$SAW_PR" -eq 0 ]; then
+    case "$tok" in
+      pr)
+        SAW_PR=1
+        continue
+        ;;
+      -R|--repo)
+        SKIP_GLOBAL_AS="repo"
+        continue
+        ;;
+      -R=*)
+        GLOBAL_REPO="${tok#-R=}"
+        continue
+        ;;
+      --repo=*)
+        GLOBAL_REPO="${tok#--repo=}"
+        continue
+        ;;
+      -*)
+        # Unknown global flag — assume boolean (--help, --version,
+        # etc.) and skip. See "Limitations" in the header for the
+        # caveat about future value-taking globals.
+        continue
+        ;;
+      *)
+        # Non-flag, non-pr token before `pr`. Either gh aliases
+        # are in play (out of scope) or the command isn't a `gh
+        # pr` invocation. Allow.
+        exit 0
+        ;;
+    esac
+  fi
+  # SAW_PR=1 — this token IS the pr subcommand.
+  PR_SUBCOMMAND="$tok"
+  break
+done
+
+# Not a pr create/merge command? Allow.
+if [ "$PR_SUBCOMMAND" != "create" ] && [ "$PR_SUBCOMMAND" != "merge" ]; then
   exit 0
 fi
 
 # --- gh pr create ---
-if echo "$COMMAND" | grep -qE '^\s*gh\s+pr\s+create'; then
+#
+# Substring grep on the raw command is fine here — the body markers
+# `Authoring-Agent:` and `## Self-Review` are content checks, not
+# structural ones, and they don't depend on argument positions or
+# global flags.
+if [ "$PR_SUBCOMMAND" = "create" ]; then
   MISSING=""
 
   if ! echo "$COMMAND" | grep -qi 'Authoring-Agent:'; then
@@ -59,7 +203,7 @@ if echo "$COMMAND" | grep -qE '^\s*gh\s+pr\s+create'; then
     MISSING="${MISSING}  - Missing '## Self-Review' section in PR body\n"
   fi
 
-  if [[ -n "$MISSING" ]]; then
+  if [ -n "$MISSING" ]; then
     echo "BLOCKED: PR description is missing required sections per REVIEW_POLICY.md:" >&2
     echo -e "$MISSING" >&2
     echo "Add these to the PR body before creating." >&2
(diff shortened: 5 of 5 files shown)


nathanpayne-claude (approved):
Internal review as nathanpayne-claude per REVIEW_POLICY.md § Phase 2. Reviewing HEAD `b7e4bf7`.

## Scope

Two commits closing three Phase 2 sub-issues from Project #2:

1. **`3016e89`** — extends `scripts/hooks/gh-pr-guard.sh` with a non-admin merge-gate cross-check (#37)
2. **`b7e4bf7`** — adds `scripts/ci/check_codex_scripts` + wires into `repo_lint.yml` + docs entries in `rules/repo_rules.md` and `scripts/ci/README.md` (#38, #39)

Five files touched. 143 insertions / 2 deletions total. `.github/workflows/repo_lint.yml` changed → triggers `external_review_paths` match → needs-external-review label applied before this review post per feedback memory label-first ordering rule.

## Correctness

### Hook (#37)

- ✅ PR number extraction via token walk (`TOKENS=( $COMMAND )` + `FOUND_MERGE` flag + integer regex) handles both `gh pr merge 65 --squash` and `gh pr merge --squash 65`
- ✅ `--repo` flag extracted separately so the label lookup is unambiguous
- ✅ No-PR-number case falls back to `gh pr view` with no positional, matching gh's current-branch resolution
- ✅ Label match via `case ",$LABELS," in *,needs-external-review,*` — exact boundary match. Lookalike label `needs-external-review-extra` tested and correctly not matched
- ✅ Fail-closed on `gh` API error: hook blocks with gh stderr + exact command attempted + escape hatch
- ✅ Existing create-path and admin-path guards untouched; verified with the same 5 test cases they were passing before
- ✅ `set -euo pipefail` unchanged; new code uses only quoted expansions; no `eval`, no unquoted `$(...)`, no shell injection surface

### CI check (#38)

- ✅ Uses the same `REPO_ROOT` discovery pattern as the other `scripts/ci/check_*` scripts
- ✅ Checks both existence (`[[ -f ]]`) and executability (`[[ -x ]]`) per issue acceptance criteria
- ✅ Exit 0 on PASS, exit 1 on FAIL. Fix instructions printed inline
- ✅ Wired into `repo_lint.yml` as a new step after `check_review_policy_exists`; existing `make_ci_scripts_executable` step (`chmod +x scripts/ci/*`) will make it executable in CI regardless of on-disk mode
- ✅ Local verification: PASS when both scripts are +x, FAIL when one loses +x, PASS again after restore

### Docs (#39)

- ✅ `rules/repo_rules.md` § CI Enforcement: new bullet added at the end with one-line rationale pointing at the Phase 4a degradation failure mode
- ✅ `scripts/ci/README.md`: new bullet in local-scripts list with the same explanation summary

## Regression risk

- ✅ Hook create and admin paths are untouched
- ✅ Hook non-admin merge path is new code — it didn't exist before, so no regression surface for non-labeled PRs (the only code path that runs is `case ",$LABELS," in` which falls through to default for any PR without the exact label)
- ✅ Hook now makes an API call on every non-admin merge (one `gh pr view --json labels` per merge). Small latency add (~200–500ms). No new gh scopes
- ✅ `repo_lint.yml` step addition is purely additive — existing steps unchanged
- ✅ `scripts/ci/check_codex_scripts` is new (no prior version to regress); file mode is 755 on disk
- ✅ `rules/repo_rules.md` and `scripts/ci/README.md` changes are additive documentation

## Test coverage

- [x] `bash -n` on both new/modified shell scripts
- [x] Hook test harness with 15 cases covering all paths:
  - create: 3 cases (missing Authoring-Agent, missing Self-Review, both present)
  - admin: 2 cases (no break-glass, with break-glass)
  - non-admin merge: 9 cases (no label, label absent, label present × uncleared/cleared, number in various positions, lookalike label, current-branch fallback, `--repo` flag)
  - unrelated command: 1 case
- [x] Hook fail-closed test with a mocked 404-returning `gh`
- [x] CI check happy path (both scripts present + executable)
- [x] CI check negative path (missing +x on one helper → FAIL + fix instructions)
- [x] CI check recovery path (restore +x → PASS)
- [ ] Not tested locally: `repo_lint.yml` end-to-end. Will run on this PR in CI; if the step fails, I'll address in-cycle
- [ ] Not tested: hook interaction with THIS PR (#66) carrying `needs-external-review` — the hook would block a `gh pr merge 66` attempt unless `CODEX_CLEARED=1` is set. Can't exercise without actually running the merge, and the merge is what the hook is gating, so this is an inherent limitation of hook-layer testing. Mocked cases cover the logic

## Security

- ✅ No new dependencies, no new gh scopes, no new env vars read from untrusted sources
- ✅ `COMMAND` is untrusted from hook's perspective but is used only in:
  - `grep -qE` pattern matching (safe — regex compiled from fixed patterns)
  - `read -ra TOKENS <<< "$COMMAND"` word splitting (safe — tokens are then matched against fixed strings or integer regex)
  - `sed -nE` extraction for `--repo` value (no command substitution on captured value)
  - `gh` invocation with array argv (not string concatenation)
- ✅ No `eval`, no unquoted `$(...)`, no path interpolation with user input
- ✅ `CODEX_CLEARED` is a boolean env var flag — no path or command injection vector from it

1. **Hook-layer cross-check is defense-in-depth, not the primary gate.** The authoritative gate is still `scripts/codex-review-check.sh`. An agent that sets `CODEX_CLEARED=1` without actually running the check is violating policy — this hook is an ordering check, not an integrity check. This is called out in the hook header comment block.

2. **File-name convention vs issue title.** Issue #38 says `check_codex_scripts.sh` but the sibling files in `scripts/ci/` have no extension. Matched the local convention. If `nathanpayne-codex` prefers the `.sh` extension for consistency with other conventions, I can rename.

3. **`check_codex_scripts` doesn't check the SCRIPTS' content, only their presence/mode.** That's deliberate per the issue body. If the scripts drift from the template (e.g., the gate (c) logic is stripped locally), CI won't catch that — only `check_duplicate_docs` or a future `check_helper_script_parity` would. Out of scope for this PR.

4. **Label lookup is one-shot per invocation.** If the PR's labels change between the hook firing and the merge actually landing (TOCTOU window), the hook's decision is based on the stale snapshot. That's acceptable for a defense-in-depth layer — the CI Label Gate and the runtime label re-verify from PR #63 are the real-time authoritative checks.

## Approved

Ready for external review. The two commits compose cleanly, each closes its issue with acceptance criteria met, and the new hook behavior has concrete test coverage including edge cases (lookalike labels, fail-closed, flag-order tolerance, current-branch fallback).

nathanpayne-codex (changes requested):
One blocking issue remains in `scripts/hooks/gh-pr-guard.sh`.

The new merge-gate label lookup only recognizes a numeric PR selector after `gh pr merge`. The CLI also accepts a PR URL or branch name as the selector. In those cases `PR_NUM` stays empty and the hook falls back to `gh pr view` for the current branch, so it can check labels for the wrong PR. That means a PR actually carrying `needs-external-review` can bypass the `CODEX_CLEARED=1` gate, or an unrelated merge can be blocked, depending on which branch the command runs from.

The CI check and docs changes look good; this remaining blocker is in the hook selector parsing.

nathanjohnpayne:

nathanjohnpayne:
@coderabbitai, can you take a look? Why didn't you start automatically?

nathanjohnpayne:
Addressed in `3afbc35`. @nathanpayne-codex ready for round-2 re-review.

## Your finding is correct

`gh pr merge` accepts `<number> | <url> | <branch>` per its CLI grammar, and my round-1 parser only captured bare integers via the `^[0-9]+$` regex. URL and branch selectors would fall through to the "no selector" current-branch fallback, causing the hook to inspect the wrong PR's labels:

- **False negative:** a PR carrying `needs-external-review` could bypass the `CODEX_CLEARED=1` gate when merged via URL or branch name
- **False positive:** a merge targeting a different PR could be blocked because the wrong PR's labels were consulted

## The fix

```bash
# Before — integer-only, misses URL and branch
if [[ "$tok" != -* ]] && [[ "$tok" =~ ^[0-9]+$ ]]; then
  PR_NUM="$tok"
  break
fi

# After — any non-flag token, with flag-value awareness
case "$tok" in
  --body|-b|--body-file|-F|--subject|-t|--author-email|-A|--match-head-commit|--repo|-R)
    SKIP_NEXT=1
    continue
    ;;
esac
if [[ "$tok" == -* ]]; then
  continue
fi
PR_SELECTOR="$tok"
break
```

Two substantive changes:

1. **Generalized the selector capture.** The parser now takes the first non-flag token after `merge` as `PR_SELECTOR` regardless of form. It's passed through to `gh pr view` unchanged — `gh pr view` accepts the same `<number> | <url> | <branch>` grammar that `gh pr merge` does, so no URL parsing, branch validation, or form coercion is needed on my side.

2. **Added value-flag awareness.** `gh pr merge` has six value-taking flags in both long and short forms:

   | Long | Short | Semantics |
   |---|---|---|
   | `--body` | `-b` | commit message body |
   | `--body-file` | `-F` | path to body file |
   | `--subject` | `-t` | commit subject |
   | `--author-email` | `-A` | author email override |
   | `--match-head-commit` | _(none)_ | SHA to match |
   | `--repo` | `-R` | owner/repo override |

   Without awareness, `gh pr merge --body TEXT 65` would misfire because `TEXT` (no leading dash) would be captured as the selector before the parser ever reached `65`. The parser now uses a `SKIP_NEXT=1` flag set when it sees any of the value-taking flags, so the next token is consumed as the flag's value and not evaluated as a selector. Inline-value forms (`--body=TEXT`, `-b=TEXT`) are single tokens and are handled by the existing `-*` prefix skip.

## Test harness

30 total cases now, all passing. 23 new cases in this round cover:

- **URL selector, 3 variations:** blocked with label + uncleared, allowed with no label, allowed with CODEX_CLEARED=1 — each verifies BOTH the exit code AND the exact URL actually passed to `gh pr view` via a mocked gh that logs its args
- **Branch selector, 3 variations:** bare branch, slash branch, branch with dots (`release/v1.2.3`), each verified against exit code and passed-through value
- **Integer selector regression check:** round-1's three integer cases still pass
- **No selector (current-branch fallback):** 2 cases, verified that `gh pr view` is called with no positional argument (not with `labels` which is `--json`'s value — my round-1 test harness had an awk bug that picked up `labels` as a false positive; fixed in this round)
- **Value-flag disambiguation, 11 cases:** every long and short form of every value-taking flag, each in the form `gh pr merge --flag VALUE 65` or `gh pr merge -f VALUE 65`, verifying the parser captures `65` as the selector and not `VALUE`
- **Inline `--body=TEXT`:** single-token form, verified passes through the `-*` filter
- **URL with query string:** `https://github.com/foo/bar/pull/65?diff=unified`, verified the whole token is passed intact to `gh pr view` (which handles the URL parsing itself)
- **Selector after many flags:** `gh pr merge --squash --delete-branch --repo foo/bar 65` — the realistic CI pattern

Plus the original 7 cases from round 1 (create Authoring-Agent/Self-Review guards, admin break-glass, non-blocking merge) — regression-checked.

## Limitations I'm NOT fixing in this round

- **Quoted flag values with whitespace.** `gh pr merge --body "hello world" 65` would still confuse the parser, because `read -ra TOKENS <<< "$COMMAND"` / `TOKENS=( $COMMAND )` word-splits on whitespace without respecting quotes. Bash can't parse shell quoting without either `python3 -c shlex.split` or a proper mini-parser; the complexity didn't feel warranted for what's an uncommon pattern in practice. Documented as a known gap in the commit message. If you want it closed I'd lean toward the shlex approach; not a huge lift but adds a Python dependency to the hook that currently has none.
- **Custom gh aliases** (`gh alias set merge '...some compound command...'`) that expand `merge` to something else. The hook guards a specific literal command grammar; aliases are out of scope.

## Residual PR-number safety claim

After this fix, for EVERY non-admin `gh pr merge` invocation that includes any selector form gh accepts, the hook either:

1. Correctly identifies the selector and asks gh pr view for THAT PR's labels
2. Or (on gh API error) fails CLOSED with a diagnostic

The only case where the hook might still make a wrong label decision is the quoted-flag-value edge case above, and that's called out explicitly. No path remaining that silently falls through to the current-branch lookup when a selector was actually specified.

## What I'm NOT changing

- CI check `scripts/ci/check_codex_scripts` and its `repo_lint.yml` wiring (commit `b7e4bf7`) — unchanged from round 1, still passing locally
- `rules/repo_rules.md` and `scripts/ci/README.md` docs — unchanged
- The existing create/admin guards in the hook — unchanged, still exercised by the original 7-case harness

## Meta

The round-1 hook was basically right in shape but wrong in scope — I reasoned only about the integer form because the round-1 smoke test naturally used `gh pr merge 65 --squash`. Missing the URL/branch forms is the same class of bug as "tested the happy path, forgot the variants". Adding to session memory: when a command accepts multiple argument forms per its `--help`, enumerate the forms up front and write a test case for each before claiming coverage.

Ready for the re-review pass.

nathanpayne-codex (changes requested):
The URL/branch-selector fix is good, but one blocking parser edge case remains in `scripts/hooks/gh-pr-guard.sh`.

The updated comment says a command like `gh pr merge --body "text" 65` is handled correctly, but `TOKENS=( $COMMAND )` still uses bash word-splitting on the raw command string and does not respect shell quoting. In practice `--body "hello world" 65` becomes tokens `--body`, `"hello`, `world"`, `65`; `SKIP_NEXT=1` only skips the first word of the quoted value, so the second word becomes `PR_SELECTOR` (`world"` in that example). That means valid `gh pr merge` commands using quoted whitespace in value-taking flags can still inspect the wrong PR labels or fail closed unexpectedly.

The CI check and docs changes still look good; this remaining blocker is only in the hook parser.

nathanjohnpayne:
Addressed in `c6d9119`. Both blockers fixed in one tokenization rewrite.

@nathanpayne-codex ready for round-3 re-review.
@coderabbitai also addressed below — `-R` short form now flows through to the label lookup.

## Two findings, one root cause

Both findings traced back to the same underlying issue: the round-1 hook used `TOKENS=( $COMMAND )` for tokenization, which performs bash word-splitting on whitespace without respecting shell quotes, AND the round-2 patch left a separate `--repo`-only `grep+sed` extraction in place that didn't share the parser. Once I fixed tokenization properly, unifying the selector + repo extraction into a single pass over the tokenized array became the obviously-correct cleanup, and that closed both findings together.

## Finding 1 (nathanpayne-codex P1) — quoted-whitespace tokenization

You're right. I documented this as a "known limitation" in the round-2 reply but you correctly insisted it be fixed.

```bash
# Before (round-2)
TOKENS=( $COMMAND )    # ← bash word split, ignores quotes

# After (round-3)
TOKENS_OUTPUT=$(printf '%s' "$COMMAND" | xargs -n 1 2>&1) || {
  echo "BLOCKED: malformed shell quoting" >&2
  exit 2
}
TOKENS=()
while IFS= read -r line; do
  TOKENS+=("$line")
done <<<"$TOKENS_OUTPUT"
```

`xargs -n 1` honors POSIX single and double quoting when splitting input into args. `gh pr merge --body "hello world" 65` now tokenizes to `(gh, pr, merge, --body, hello world, 65)` — quotes stripped, value preserved as one token — which the existing `SKIP_NEXT_AS=skip` value-flag logic then consumes correctly so `65` is captured as the selector.

**Bash 3.2 portability note.** I started with `mapfile -t TOKENS <<<"$TOKENS_OUTPUT"` which is cleaner, but macOS ships bash 3.2 by default and `mapfile` is a bash 4+ builtin. Switched to a portable `while IFS= read -r line` loop. Verified locally on `bash --version` 3.2.57 — all 49 test cases pass.

**Empty-line handling.** Initially I had `[ -z "$line" ] && continue` in the read loop to skip empty lines, which broke `gh pr merge --body "" 65`: xargs emits an empty line for the legitimate empty-string value, and skipping it caused the SKIP_NEXT_AS to not consume anything, so `65` got eaten as the body value and PR_SELECTOR was left empty. Removed the skip — empty lines are now preserved so legitimate empty arg values are consumed correctly. Caught by a dedicated test case before commit.

**Fail-closed on tokenization error.** Unmatched quotes, bad escapes, etc. cause xargs to exit non-zero. The hook captures the xargs error and blocks with a diagnostic showing the exact command and the xargs error message, matching the existing fail-closed pattern for `gh` API errors elsewhere in the hook.

## Finding 2 (CodeRabbit Major) — `-R` short form not in `REPO_ARG`

CodeRabbit caught this on the same review pass. The token parser at line 129 already correctly handled `-R` as a value-taking flag for SKIP_NEXT purposes, but the SEPARATE `REPO_ARG` extraction (used to forward `--repo` to `gh pr view`) only matched `--repo`:

```bash
# Before — only matches --repo
if echo "$COMMAND" | grep -qE '(^|\s)--repo(\s|=)'; then
  REPO_ARG=$(echo "$COMMAND" | sed -nE 's/.*--repo[= ]([^ ]+).*/\1/p')
fi
```

So `gh pr merge -R foo/bar 65` would correctly identify `65` as the selector but wouldn't forward `--repo foo/bar` to `gh pr view`, leaving the label lookup querying the default repo instead of the explicitly-specified one. False negatives possible.

CodeRabbit's proposed fix used grep/sed with an `(--repo|-R)` alternation, which would work, but I went a different way: **collapse `PR_SELECTOR` and `REPO_ARG` extraction into a single unified pass over the now-properly-tokenized `TOKENS` array**, with explicit case branches for each form (`--repo`, `-R`, `--repo=value`, `-R=value`). The grep+sed extraction is removed entirely.

```bash
case "$tok" in
  --repo|-R)        SKIP_NEXT_AS="repo"; continue ;;
  --repo=*)         REPO_ARG="${tok#--repo=}"; continue ;;
  -R=*)             REPO_ARG="${tok#-R=}"; continue ;;
  --body|-b|--body-file|-F|--subject|-t|--author-email|-A|--match-head-commit)
                    SKIP_NEXT_AS="skip"; continue ;;
esac
```

Side benefit beyond what CodeRabbit asked for: `REPO_ARG` now correctly handles `--repo`/`-R` that appears AFTER the selector (e.g., `gh pr merge 65 --repo foo/bar`). The previous implementation walked tokens only until it found the selector and broke; the unified loop walks ALL tokens after `merge` so trailing `--repo` is captured too. Tested.

## Test harness

49 total cases now, all passing on bash 3.2.57:

| Round | Cases | Coverage |
|---|---|---|
| 1 | 7 | create / admin / non-blocking merge / unrelated command — full regression |
| 2 | 23 | URL / branch / integer / no-selector / value-flag disambiguation / inline-value / URL with query string |
| **3** | **19 new** | **(see below)** |

Round-3 cases:

- **8 quoted-whitespace cases:** `--body "hello world" 65`, `--body 'multi word'`, `--subject "x" --body "y" 65`, `-b "quoted" 65`, `-t "quoted" 65`, `65 --body "post-selector"`, plus label-absent and CODEX_CLEARED variations
- **7 `-R`/`--repo` cases:** all four forms (`--repo`, `-R`, `--repo=`, `-R=`), pre-selector and post-selector positions, plus negative case (no `--repo` → not forwarded)
- **2 fail-closed cases:** unmatched double quote, unmatched single quote — both produce `exit 2` with diagnostic
- **1 empty-quoted-value case:** `--body "" 65` → empty value consumed as body, `65` captured as selector
- **1 kitchen sink:** `--repo foo/bar --subject "fix: x" --body "long body" 65 --squash --delete-branch` → captures `65`, forwards `--repo foo/bar`

Each case verifies BOTH the exit code AND (where applicable) the exact positional + `--repo` args actually passed to `gh pr view` via a mocked `gh` that logs its args for inspection.

## Limitations remaining (out of scope for this round)

- **Backslash escapes inside double-quoted strings** (`"with \"escape\""`) — xargs treats `\` inside double quotes as having no escape semantics, so nested quotes fail tokenization (which → fails closed). This is the gh-CLI-side equivalent of the same edge case bash itself wouldn't handle without a real shlex implementation. If you want this closed, the path forward is `python3 -c "import shlex"` which adds a Python dependency to the hook (~50ms startup overhead per Bash tool call). I don't think it's worth it for an edge case that fails closed.
- **Custom gh aliases** that expand `merge` to something else. Per-user config; the hook guards a specific literal command grammar.

Both limitations are documented in the commit message and in the hook header comment.

## What I'm NOT changing in this round

- `scripts/ci/check_codex_scripts` and its `repo_lint.yml` wiring — unchanged from round 1, still passing locally and in CI
- `rules/repo_rules.md` and `scripts/ci/README.md` — unchanged
- The hook's existing create-path and admin-path guards — unchanged, still exercised by the original 7-case harness regression run

## Meta

Round 2's lesson was "enumerate the argument forms before claiming coverage". Round 3's lesson is **"a structural problem in tokenization will cause repeated, correlated bugs in everything that consumes the tokens"** — both findings here trace back to the same naive `TOKENS=( $COMMAND )` line, and patching them piecemeal would have been both larger AND less correct than the structural rewrite. The diff is +91/-41, of which the +91 includes a substantial new comment block explaining the iteration history so the next person who touches this code knows why xargs was chosen. Adding to session memory: when two reviewers find related bugs in adjacent code on the same review pass, look for the shared root cause before fixing each in isolation.

Ready for the re-review pass.

nathanpayne-codex (approved):
No blocking findings on the updated PR.

The tokenization rewrite closes the quoted-whitespace parser bug, the unified single-pass extraction now handles `--repo`, `-R`, `--repo=*`, and `-R=*` correctly, and the trailing-`--repo` case is covered by continuing the scan after `PR_SELECTOR` is found. I also checked the latest CodeRabbit reply on April 15, 2026, and it agrees both blockers are resolved.

The remaining documented limitations are acceptable for this hook layer: custom gh aliases are out of scope, and the tokenization edge cases that remain are fail-closed rather than silent false-clears. Clear to merge once the normal Phase 4a gate is green.

nathanjohnpayne:

nathanpayne-codex (changes requested):
One blocking issue remains in `scripts/hooks/gh-pr-guard.sh`.

The top-level matcher still only recognizes commands of the form `gh pr create` / `gh pr merge`. GitHub CLI also allows global flags like `-R owner/repo` or `--repo owner/repo` before `pr`, so valid commands like `gh -R foo/bar pr merge 65` or `gh --repo foo/bar pr create ...` currently miss the hook entirely. That bypasses both the create-path metadata guard and the merge-path `needs-external-review` / `CODEX_CLEARED` gate when the command uses the repo override in the global-flag position.

I rechecked the other round-3 fixes and they look good. The remaining blocker is just the initial command detection needing to tolerate global gh flags before `pr`.

nathanjohnpayne:

nathanjohnpayne:
Addressed in `d07f0bf`. @nathanpayne-codex ready for round-4 re-review.

## Your finding is correct

`gh -R foo/bar pr merge 65 --squash` and `gh --repo foo/bar pr create ...` both bypass the round-3 hook entirely because the top-level matcher was `^\s*gh\s+pr\s+(create|merge)` — substring grep that didn't allow anything between `gh` and `pr`. That meant the create-path metadata guard AND the merge-path `needs-external-review` / `CODEX_CLEARED` gate were both silently skipped for any command that put a global flag before the subcommand. Real bypass.

## Three rounds, one architectural pattern

Each finding on this PR has been the same shape: substring matching on a string of unknown structure.

| Round | Finding | Substring being matched | Bypass |
|---|---|---|---|
| 2 (P1) | `TOKENS=( $COMMAND )` ignored quotes | bash word-split on whitespace | `--body "hello world"` |
| 3 (Major) | `grep -qE '(^&#124;\s)--repo(\s&#124;=)'` | regex on raw command | `-R` short form |
| 4 (P1) | `'^\s*gh\s+pr\s+(create&#124;merge)'` | regex on raw command | global `-R` before `pr` |

The pattern: every time I patched ONE form, a different form was waiting. The structural fix is to **stop using substring matching anywhere in this hook** and do all detection on a token array produced by `xargs -n 1` (which honors POSIX shell quoting).

Round 3 already moved tokenization into the merge sub-block. Round 4 lifts it to the TOP of the script and replaces the create/merge dispatch entirely with a token-aware subcommand walk.

## New top-level structure

```
1. Quick exit if not a gh command at all
2. Tokenize with xargs (fail closed on malformed quoting)
3. Walk tokens to identify the pr subcommand:
     - skip global -R/--repo (capturing GLOBAL_REPO)
     - skip unknown global flags as boolean (--help, --version, etc.)
     - exit 0 on non-gh-pr commands
4. Branch on PR_SUBCOMMAND:
     - "create": substring checks for body markers
     - "merge": existing token walk for PR_SELECTOR + REPO_ARG,
       fall back to GLOBAL_REPO if subcommand-level REPO_ARG empty
     - anything else (view, list, status...): allow
```

The substring grep for `Authoring-Agent:` and `## Self-Review` in the create branch is intentionally kept — those are content checks, not structural ones, so substring matching is the right tool. The structural change is at the top-level dispatch only.

## Conflict resolution: global vs subcommand `--repo`

When BOTH a global and a subcommand-level `-R` are present, the subcommand-scoped value wins, mirroring gh's typical "more specific flag wins" behavior:

```bash
gh -R global/repo pr merge -R local/repo 65
  → label lookup uses local/repo

gh -R global/repo pr merge 65
  → label lookup falls back to global/repo
```

Both cases are tested. (My initial design considered global-wins, but that contradicts gh's actual behavior and would surprise reviewers reading the hook later.)

## Test harness

74 total cases now, all passing on bash 3.2.57:

| Round | Cases | New |
|---|---|---|
| 1 | 7 | create / admin / non-blocking merge / unrelated command |
| 2 | 23 | URL / branch / integer / no-selector / value-flag disambiguation |
| 3 | 19 | quoted whitespace / `-R` extraction / fail-closed / kitchen sink |
| **4** | **25** | **(see below)** |

Round-4 cases:

- **8 global-flag merge cases:** global `-R` + label + uncleared, global `--repo` + label, inline `-R=value`, inline `--repo=value`, global + cleared, global + no label, **global vs subcommand conflict (both directions)**
- **4 global-flag create cases:** global `-R` + missing Authoring-Agent → block, missing `## Self-Review` → block, both present + `-R` → allow, both present + `--repo` → allow
- **2 global-flag `--admin` cases:** the admin guard still fires when the command starts with `gh -R foo/bar pr merge --admin 65`
- **3 non-create-merge subcommand cases:** `gh -R foo/bar pr view 65`, `gh -R foo/bar pr list`, `gh -R foo/bar issue create` — all correctly NOT gated (allowed)
- **3 boolean global flag cases:** `gh --version`, `gh --help`, `gh pr --help` — all correctly not gated
- **1 multiple-global case:** `gh --version -R foo/bar pr merge 65` — still gated correctly
- **2 global + URL/branch selector cases:** `gh -R foo/bar pr merge https://github.com/foo/bar/pull/65` and `gh -R foo/bar pr merge feat/x`
- **2 non-gh command cases:** `git push`, `ghi pr merge 65` — both correctly not gated (the matcher must NOT fire on lookalikes that start with `g`)

Each case verifies BOTH the exit code AND (where applicable) the actual positional + `--repo` args passed to `gh pr view` via a mocked `gh` that logs its args for inspection.

## On CodeRabbit's secondary note

You called out the second CodeRabbit comment about forcing `GH_TOKEN="$OP_PREFLIGHT_REVIEWER_PAT"` for the label lookup as non-blocking, and I agree with that read. The hook currently makes the `gh` API call with whatever `GH_TOKEN` happens to be in the agent's environment. If the agent ran `op-preflight.sh` first (which CLAUDE.md step 0 mandates), `GH_TOKEN` will be set to the reviewer PAT for the duration of that session, and label lookups work. If they didn't, the lookup either uses gh's ambient auth or fails closed with the existing diagnostic. Forcing a specific PAT inside the hook would couple the hook to the preflight script's variable name and create a hard dependency that doesn't exist today. I'm leaving it as-is, with the existing fail-closed path catching the "no auth" case. Happy to reverse if you push back.

## Limitations remaining (documented in hook header)

- **Backslash escapes inside double-quoted strings** (`"with \"escape\""`) — xargs has no escape semantics, fails closed
- **Custom gh aliases** that expand `merge` to something else — out of scope, hook guards a literal command grammar
- **Unknown VALUE-taking global flags** (anything other than `-R/--repo`) are assumed boolean. gh currently has only `-R/--repo` as a value-taking global; if gh adds more, the hook needs an update. Misclassification would let the next token leak through as the subcommand name. Caveat called out in the header so the next person who touches this code sees it.

## Diff size

`scripts/hooks/gh-pr-guard.sh`: 268 insertions / 177 deletions. The big delta is mostly the comment block reorganization at the top, plus the new top-level walk replacing the three `grep -qE` substring matchers. The behavioral changes are concentrated in the new walk and the `GLOBAL_REPO` fallback at the end of the merge block.

## What I'm NOT changing in this round

- `scripts/ci/check_codex_scripts` and its `repo_lint.yml` wiring — unchanged from round 1, still passing
- `rules/repo_rules.md` and `scripts/ci/README.md` — unchanged
- The hook's gate semantics (CODEX_CLEARED requirement, label match, fail-closed on gh API error) — unchanged, just reorganized

## Meta

I think I've now exhausted the parser-shape bug class on this PR — every iteration that introduced a new bypass came from substring-matching on a string of unknown structure, and the round-4 refactor moves ALL detection into the tokenized array. If the next finding is also a parser bug, I'd want to take a hard look at the architecture (e.g., move to Python `shlex` / `argparse`) rather than patch yet again. Current best-guess: the next round should be approval, but I'll be honest if I'm wrong about that and there's another shape I've missed.

Ready for the re-review pass.

nathanpayne-codex (changes requested):
One blocking issue remains in `scripts/hooks/gh-pr-guard.sh`.

The new top-level parser still assumes `tool_input.command` starts with literal `gh`, so environment-prefixed invocations bypass the hook entirely. That matters here because the documented happy paths for this hook use exactly those prefixes: `CODEX_CLEARED=1 gh pr merge ...` for a cleared Phase 4a merge and `BREAK_GLASS_ADMIN=1 gh pr merge --admin ...` for break-glass. With the current code, both of these inputs hit the early `^\s*gh(\s|$)` check, miss it, and exit `0` before any create/admin/merge guard runs. I verified it locally by piping those exact JSON payloads through the script and getting `EXIT:0` for both.

That reopens a silent bypass for the two most sensitive paths the hook is supposed to guard: a labeled PR can skip the `CODEX_CLEARED` cross-check when merged via the documented prefixed form, and an `--admin` merge can skip the explicit break-glass check the same way. The fix needs to tolerate leading shell environment assignments before `gh`, not just global `-R/--repo` flags after `gh`.

nathanpayne-codex (changes requested):
One blocking issue remains in `scripts/hooks/gh-pr-guard.sh`.

The new top-level guard still requires `tool_input.command` to start with literal `gh`. That means the documented merge paths `CODEX_CLEARED=1 gh pr merge ...` and `BREAK_GLASS_ADMIN=1 gh pr merge --admin ...` skip the parser entirely and return `0` before any create/admin/merge checks run. I verified this locally by piping both exact JSON payloads through the script and getting `EXIT:0` for each. The top-level detection needs to tolerate leading shell environment assignments before `gh`, not just global `-R/--repo` flags after it.

nathanjohnpayne:
Addressed in `ef93d42`. @nathanpayne-codex ready for round-5 re-review.

## Your finding is correct, and there's a second layer underneath it

I reproduced your finding locally with the round-4 hook:

```bash
$ payload=$(jq -n --arg cmd "CODEX_CLEARED=1 gh pr merge 65 --squash" '{tool_input: {command: $cmd}}')
$ echo "$payload" | bash scripts/hooks/gh-pr-guard.sh; echo exit=$?
exit=0

$ payload2=$(jq -n --arg cmd "BREAK_GLASS_ADMIN=1 gh pr merge --admin 65" '{tool_input: {command: $cmd}}')
$ echo "$payload2" | bash scripts/hooks/gh-pr-guard.sh; echo exit=$?
exit=0
```

Both forms exit 0 before any guard runs. The early-exit matcher `^\s*gh(\s|$)` requires the command to start with literal `gh`, and env-prefixed commands start with the env var assignment instead.

But there's a SECOND bug underneath your finding that I would have hit on the same change if I'd only fixed the matcher. The bash form `VAR=value command args` sets `VAR` in the spawned command's environment, **not** in the PreToolUse hook process. The hook fires before bash executes the command, so:

```bash
# Hook reads ${CODEX_CLEARED:-} from its OWN env, not from
# the inline-prefix on the command being inspected.
if [ "${CODEX_CLEARED:-}" != "1" ]; then ...
```

If I had only relaxed the matcher in this round, both documented forms would have proceeded past the matcher and then **still** blocked at the `CODEX_CLEARED` / `BREAK_GLASS_ADMIN` checks because the hook would be reading its own env (where neither is set), not the inline prefix. Less of a bypass than before, but still broken from the user's perspective.

## Two coupled fixes, one round

```bash
# Fix 1 — early matcher allows leading prefix material:
if ! echo "$COMMAND" | grep -qE '(^|\s)gh(\s|$)'; then
  exit 0
fi
```

The new matcher accepts `gh` as a whitespace-bounded word anywhere in the command. False positives (e.g., `cat gh.txt` doesn't match because `.` isn't `\s`) are caught downstream by the token walk's "non-flag, non-pr token before pr → exit 0" branch and the post-walk `PR_SUBCOMMAND` check.

```bash
# Fix 2 — top-level walk captures inline env assignments:
case "$tok" in
  CODEX_CLEARED=*)        INLINE_CODEX_CLEARED="${tok#CODEX_CLEARED=}" ;;
  BREAK_GLASS_ADMIN=*)    INLINE_BREAK_GLASS_ADMIN="${tok#BREAK_GLASS_ADMIN=}" ;;
esac

# After the walk:
EFFECTIVE_CODEX_CLEARED="${CODEX_CLEARED:-${INLINE_CODEX_CLEARED:-}}"
EFFECTIVE_BREAK_GLASS_ADMIN="${BREAK_GLASS_ADMIN:-${INLINE_BREAK_GLASS_ADMIN:-}}"
```

Then the merge guards switch from reading the raw env vars to reading `EFFECTIVE_*`. Exported value wins (mirrors typical shell precedence — explicit `export` overrides inline-prefix), inline is the fallback. Both forms now work; the `export` form continues to work; and a mixed case where the agent does `export CODEX_CLEARED=1` once and then runs commands with `CODEX_CLEARED=0 gh ...` keeps the exported value rather than honoring the override (because that's a foot-gun and `export` is the more deliberate signal).

## Conflict resolution: exported vs inline

Tested explicitly:

| Exported | Inline | Outcome |
|---|---|---|
| (unset) | (none) | block — neither set, default-deny |
| =1 | (none) | allow — exported, existing behavior preserved |
| (unset) | =1 | allow — inline picked up (the new behavior) |
| =1 | =1 | allow — both consistent |
| =1 | =0 | allow — exported wins |
| (unset) | =0 | block — inline=0 doesn't match "=1" |
| (unset) | =true | block — only literal "=1" matches (precise check) |
| (unset) | = (empty) | block — empty value doesn't match "=1" |

## Test harness

97 total cases now, all passing on bash 3.2.57:

| Round | Cases | Coverage |
|---|---|---|
| 1 | 7 | create / admin / non-blocking merge / unrelated command |
| 2 | 23 | URL / branch / integer / no-selector / value-flag |
| 3 | 19 | quoted whitespace / `-R` extraction / fail-closed / kitchen sink |
| 4 | 25 | global `-R/--repo` before `pr` / boolean global flags / non-gh lookalikes |
| **5** | **23** | **(see below)** |

Round-5 cases:

- **7 `CODEX_CLEARED` inline cases:** the documented happy path `CODEX_CLEARED=1 gh pr merge 65 --squash`, inline=0 → block, no env → block, inline + exported unset, exported + no inline (existing behavior regression check), exported wins over inline=0, both =1
- **5 `BREAK_GLASS_ADMIN` inline cases:** mirrored from CODEX_CLEARED with `--admin`
- **3 multi-env-var cases:** `CODEX_CLEARED=1 GH_TOKEN=xxx gh ...`, different orders, `CODEX_CLEARED=1 BREAK_GLASS_ADMIN=1 gh pr merge --admin`
- **2 env-prefixed create cases:** `GH_TOKEN=foo gh pr create` with and without body markers — verifies env prefix doesn't bypass the create-path content checks
- **6 edge cases:**
  - `GH_TOKEN=foo git push` — non-gh command, not gated even with leading env
  - `GH_TOKEN="quoted value" gh pr merge ...` — quoted env value handled (xargs collapses to single token)
  - `sudo gh pr merge 65` — non-env prefix word handled too
  - `sudo CODEX_CLEARED=1 gh pr merge 65` — both prefixes interleaved
  - `CODEX_CLEARED= gh ...` (empty value) → block
  - `CODEX_CLEARED=true gh ...` (non-"1" value) → block (precise literal match)

Each case verifies the exit code matches expected for that combination of inline / exported / label / admin state.

## Diff

`scripts/hooks/gh-pr-guard.sh`: 72 insertions / 11 deletions. Most of the delta is the new `INLINE_*` capture in the existing top-level walk and the `EFFECTIVE_*` derivation; the two guard sites (`--admin` and `needs-external-review`) get one-line substitutions.

## Limitations remaining (documented in header)

- **Other env vars** in the inline prefix (anything other than CODEX_CLEARED and BREAK_GLASS_ADMIN) are skipped without interpretation. This is fine because no other env var is consulted by hook policy decisions. If we add more policy env vars later, they need to be added to the capture list too — there's a comment in the header noting this.
- Backslash escapes inside double-quoted strings — fails closed (xargs limitation)
- Custom gh aliases — out of scope
- Unknown VALUE-taking global flags after `gh` — assumed boolean

## What I'm NOT changing in this round

- `scripts/ci/check_codex_scripts` and `repo_lint.yml` wiring — unchanged from round 1
- `rules/repo_rules.md` and `scripts/ci/README.md` — unchanged
- The hook's gate semantics (label match, fail-closed on gh API error, --admin substring grep) — unchanged

## Meta

I said in the round-4 reply "I think I've now exhausted the parser-shape bug class on this PR — but I'll be honest if I'm wrong about that". I was wrong. The round-4 fix retired substring-matching as the *structural-detection* path, but I still had a substring matcher in the *early-exit* filter, which is structurally the same pattern in a different location. Both bugs you found in round-3 and round-4 were variants of "the literal form of the command is not the only way to express it" — round-3 was URL/branch selectors after `merge`, round-4 was `-R` before `pr`, round-5 is env-prefix before `gh`. Each round narrowed the bypass surface by one layer.

What I should have done in round 4: enumerated ALL the prefix forms that bash accepts before a command (env assignments, builtin prefix words like `eval`/`time`/`sudo`/`nohup`, redirects, control operators) and tested each one against my matcher. I tested for global flags after `gh` because that's what the round-3 finding was about; I didn't think to test prefixes BEFORE `gh`.

Updated session memory: when fixing a bypass, **enumerate the entire shell grammar position equivalent to the bypass** before claiming coverage. Don't just patch the specific case the reviewer found — find the next case that's structurally equivalent and write a test for it before posting the reply.

The current state genuinely has no substring matching left in any structural-detection path. The only `grep -q` calls remaining are:
1. `(^|\s)gh(\s|$)` — early-exit filter (still substring, but the walk catches false positives downstream)
2. `Authoring-Agent:` / `## Self-Review` — content checks in the create branch (substring is the right tool for content)
3. `\-\-admin` — the `--admin` substring check in the merge branch (the position doesn't matter, substring is fine)

If round-6 surfaces another bypass, the next architectural step is a `python3 -c shlex.split` rewrite. I'm holding that in reserve; the bash version is still readable and fast, and going to Python should be the LAST resort.

Ready for the re-review pass.

nathanpayne-codex (changes requested):
One blocking issue remains in `scripts/hooks/gh-pr-guard.sh`.

The round-5 fix now skips arbitrary tokens until it finds `gh`, which means unrelated commands that merely mention `gh pr merge` in their arguments are treated as real merges. I verified this locally by piping `{"tool_input":{"command":"echo gh pr merge 66"}}` through the hook and getting the merge-gate block (`EXIT:2` with the `needs-external-review` message). `printf %s gh pr merge 66` behaves the same way.

That is a real false positive on non-`gh` commands: the relaxed `(^|\s)gh(\s|$)` filter plus the top-level walk's "keep skipping until you see `gh`" logic no longer distinguishes legitimate shell prefixes (`CODEX_CLEARED=1`, `sudo`, `time`) from an ordinary command whose later arguments happen to contain those tokens. The fix needs to accept only real command-prefix forms before `gh`, not any arbitrary leading word.

nathanjohnpayne:
Addressed in `c0bff5d`. @nathanpayne-codex ready for round-6 re-review. Also updating issue #67 with a note about how this round informs the shlex-investigation question.

## Your finding is correct

I reproduced both bypasses against the round-5 hook locally:

```
$ payload=$(jq -n --arg cmd "echo gh pr merge 66" '{tool_input:{command:$cmd}}')
$ echo "$payload" | bash scripts/hooks/gh-pr-guard.sh; echo exit=$?
BLOCKED: PR carries 'needs-external-review' and CODEX_CLEARED is not set.
exit=2

$ payload2=$(jq -n --arg cmd "printf %s gh pr merge 66" '{tool_input:{command:$cmd}}')
$ echo "$payload2" | bash scripts/hooks/gh-pr-guard.sh; echo exit=$?
BLOCKED: ...
exit=2
```

The hook was firing on `echo` and `printf` invocations, treating their argument-list `gh pr merge 66` as a real merge command. Round 5's relaxation of the early-exit matcher to `(^|\s)gh(\s|$)` plus a pre-gh walk that blindly continued past any non-gh token combined to over-match. The matcher was correct in isolation; the walk was correct in isolation; together they let `echo gh pr merge 66` pass.

This is a false POSITIVE rather than the false NEGATIVE bypasses of rounds 2-5 — the hook fires on commands it shouldn't. It would block legitimate use of `echo` and `printf` if the user happens to type the literal string `gh pr merge` as text. Same severity (P1) but opposite polarity.

## The fix

Replace the "skip everything until gh" pre-gh walk with a small command-position state machine. Two states:

```
AT_COMMAND_POSITION (initial):
  - env assignments         VAR=value             → stay
  - prefix commands         sudo, eval, time,     → stay
                            nohup, env, command,
                            exec, nice, ionice
  - flags of those prefixes -X                    → stay
  - compound separators     ; && || | & ( )       → stay
  - gh                                            → transition to phase 2
  - any other token         echo, printf, cat...  → transition to IN_UNRELATED_ARGS

IN_UNRELATED_ARGS:
  - compound separators (standalone)              → reset to AT_COMMAND_POSITION
  - everything else                               → skip
```

Any command-line position that is NOT a "command-position prefix" terminates the search for gh and skips through the rest until a separator. This is the right model for shell command grammar without going to a real parser.

Inline env capture (CODEX_CLEARED= / BREAK_GLASS_ADMIN=) runs in BOTH states — capturing for non-gh commands is harmless since EFFECTIVE_* values are only consulted by the create/merge guards which only run when SAW_GH=1.

## What the prefix command allowlist contains and why

| Token | Why allowed |
|---|---|
| `sudo` | Common privilege escalation prefix |
| `eval` | Common, though dangerous in general |
| `time` | Bash builtin and `/usr/bin/time` measurement |
| `nohup` | Daemonization prefix |
| `env` | The `env VAR=value command` form, including `env CODEX_CLEARED=1 gh ...` |
| `command` | Bash builtin to bypass functions/aliases |
| `exec` | Process replacement |
| `nice`, `ionice` | Priority/scheduling prefixes |

I deliberately left out:

- `taskset`, `chrt` — niche scheduling, can add if needed
- `unbuffer`, `script` — pty wrappers, niche
- `xargs` — using xargs to dispatch gh is unusual and the hook can't predict args
- `bash -c "..."` — needs recursive parsing, out of scope (the inner string is one xargs token)

## Test harness

133 total cases now, all passing on bash 3.2.57:

| Round | Cases | Coverage |
|---|---|---|
| 1 | 7 | create / admin / merge / unrelated command |
| 2 | 23 | URL / branch / integer selectors |
| 3 | 19 | quoted whitespace / `-R` extraction / fail-closed |
| 4 | 25 | global `-R/--repo` / boolean global flags |
| 5 | 23 | inline env prefixes (`CODEX_CLEARED=` / `BREAK_GLASS_ADMIN=`) |
| **6** | **36** | **(see below)** |

Round-6 cases:

- **12 gh-as-argument cases (must NOT trigger):**
  - `echo gh pr merge 66` (your repro)
  - `printf %s gh pr merge 66` (your other repro)
  - `find . -name gh`, `grep gh README.md`, `which gh`, `type gh`, `ls /usr/bin gh`
  - `echo gh foo bar`, `echo -n gh pr merge 66`, `printf '%s\n' gh pr merge 66`
  - `cat gh.log`, `cat foo.txt`
- **11 prefix-command cases (must STILL trigger):**
  - `sudo / time / nohup / env / command / exec / nice gh pr merge 65`
  - `sudo -E gh pr merge 65` (prefix with flag)
  - `time -p gh pr merge 65`
  - `CODEX_CLEARED=1 sudo gh pr merge 65` (env then prefix)
  - `sudo CODEX_CLEARED=1 gh pr merge 65` (prefix then env)
- **7 compound-separator cases:**
  - `cd foo && gh pr merge 65`, `cd foo ; gh pr merge 65`, `false || gh pr merge 65`
  - `make build && CODEX_CLEARED=1 gh pr merge 65`
  - `echo done ; gh pr merge 65` (intermediate echo, separator resets)
  - `( gh pr merge 65 )` (subshell)
  - `( echo gh pr merge 65 )` → NOT gated (gh is arg of echo inside parens)
- **6 edge cases:**
  - gh in flag value (`echo --search=gh-pr`)
  - gh in pipe target (`echo foo | grep gh`)
  - real gh piped to tee (`gh pr merge 65 | tee log`)
  - `gh status` → not gated (not pr)
  - `gh issue list` → not gated (not pr)
  - `gh pr view 65` → not gated (view, not create/merge)

Each case verifies the exit code matches expected for that combination of inline / exported / label / admin state.

## Limitations remaining (header updated)

- **Compound separators glommed onto adjacent tokens** without whitespace (e.g. `foo;bar` instead of `foo ; bar`) are NOT detected — xargs preserves them as one token. Acceptable: non-idiomatic shell.
- **`bash -c "gh pr merge ..."`** is not gated because xargs gives the inner quoted string as one token to bash, which is an unrelated command. Out of scope for the bash version of the hook.
- **Custom prefix commands** not in the allowlist (e.g. `taskset`, `chrt`) cause the gh that follows them to be skipped. The current allowlist covers the common cases; we can extend if needed.

## Issue #67 update

Round 6 is also another data point for [#67](https://github.com/nathanjohnpayne/ai_agent_repo_template/issues/67). The pattern keeps holding: "the literal form of the command is not the only way to express it" still surfaces new bugs each round. **But** round 6 is the first round where the bypass was a false POSITIVE rather than a false NEGATIVE — that's an interesting shift in the failure mode and worth noting in the spike report whoever picks up #67 writes. I'll add a comment to #67 referencing this round so the investigator has the full picture.

## What I'm NOT changing in this round

- `scripts/ci/check_codex_scripts` and `repo_lint.yml` wiring — unchanged
- `rules/repo_rules.md` and `scripts/ci/README.md` — unchanged
- The hook's gate semantics, the post-gh phase 2 walk, the merge gate, the create-path content checks — all unchanged

## Meta

Six rounds in. Each round closed the bypass the reviewer found AND opened a smaller adjacent surface that the next round caught. The pattern across rounds 5 → 6 is interesting: round 5 was about adding flexibility for documented forms, round 6 was about constraining the flexibility I added so it didn't over-match. Both directions exist for the same parser-shape reason, and going forward I should think of "more permissive matcher" and "more conservative matcher" as the same axis — every loosening needs a corresponding tightening to constrain the surface to exactly the intended forms.

Updated session memory: when relaxing a matcher to handle a new form, **enumerate the false-positive surface area opened by the relaxation and write a test for it BEFORE shipping**, not after the next reviewer round. The fact that I had to be told about `echo gh pr merge` is a process failure on my end, not just a "bash parsing is hard" failure.

Ready for the re-review pass. If round 7 surfaces yet another bypass (positive or negative), I'm going to push hard on #67 and start the shlex spike — six rounds of the same class of bug is enough.

nathanpayne-codex (changes requested):
One blocking issue remains in `scripts/hooks/gh-pr-guard.sh`.

The `--admin` guard still uses a raw substring grep on `COMMAND`, so valid non-admin merges can be misclassified as break-glass when `--admin` appears inside an option value. I verified this locally with `{"tool_input":{"command":"gh pr merge 66 --subject \"--admin follow-up\""}}`, and the hook blocked it with the explicit `--admin merge requires explicit human authorization` path.

That means the round-6 parser fixes are good, but the hook still has a false-positive path on real `gh pr merge` syntax. Since the rest of the file already tokenizes the command and skips value-taking flags like `--subject`, this guard should be aligned with the token walk instead of matching the raw command string.

nathanpayne-codex:
I also checked the newest current-head CodeRabbit finding after posting my `CHANGES_REQUESTED` review, and that one is real too.

The command-position state machine currently treats value-taking prefix flags as if they were standalone boolean flags, so their values get misclassified as the start of an unrelated command and the real `gh` command is never seen. I verified this locally with:

- `sudo -u admin gh pr merge 66`
- `time -f fmt gh pr merge 66`
- `nice -n 5 gh pr merge 66`

All three currently exit `0` from the hook instead of hitting the label gate. So in addition to the `--admin` substring false-positive I already flagged in my review, there is also a real false-negative path on prefix commands whose flags take values.

nathanjohnpayne:
Addressed in `f6a5df0`. Both your blockers fixed plus a third bug I tripped over while debugging the second one.

@nathanpayne-codex ready for round-7 re-review.

## Three bugs in one round

### Bug 1 (your P1) — `--admin` substring grep over-matches

You're right. The `--admin` sub-guard used `grep -q '\-\-admin'` on the raw COMMAND string, which over-matched any `--admin` substring inside a quoted flag value:

```bash
gh pr merge 65 --subject "--admin follow-up" --squash
  → falsely blocked as a break-glass merge
```

Repro:
```
BLOCKED: --admin merge requires explicit human authorization.
exit=2  (expected: not blocked, --admin is in the subject value)
```

**Fix:** replace the substring grep with token-based detection. The merge block already walks TOKENS to find `PR_SELECTOR` and `REPO_ARG`; I added `ADMIN_REQUESTED` as a third output of the same walk:

```bash
case "$tok" in
  --admin)
    ADMIN_REQUESTED=1
    continue
    ;;
  ...
  --body|-b|--body-file|-F|--subject|-t|--author-email|-A|--match-head-commit)
    SKIP_NEXT_AS="skip"
    continue
    ;;
esac
```

Now `--admin` only fires when it's a real flag of `merge`, not a substring of another flag's value. The value-flag SKIP logic correctly consumes `--admin follow-up` as the subject's value.

### Bug 2 (CodeRabbit Major) — value-taking prefix flags consume `gh`

Also right. The pre-gh state machine treated any `-X` flag as boolean (continue, stay in command position). For value-taking flags this dropped the value into `in_unrelated_args` and ate `gh`:

```bash
sudo -u user gh pr merge 65    # -u takes "user" as value
time -f format gh pr merge 65  # -f takes "format" as value
nice -n 5 gh pr merge 65       # -n takes "5" as value
```

**Fix attempt #1 (didn't work):** generic value-flag list `-u|-g|-U|-p|-r|-s|-t|-c|-D|-h|-f|-o|-n|-S`. Failed because `time -p` is the boolean POSIX format flag — putting `-p` on a generic value list would consume `gh` as the value of `-p`.

**Fix attempt #2 (works):** **per-prefix value-flag map.** Track `CURRENT_PREFIX` (most recently seen prefix command) and match `$CURRENT_PREFIX:$tok` against an allowlist:

```bash
case "$CURRENT_PREFIX:$tok" in
  sudo:-u|sudo:-g|sudo:-U|sudo:-h|sudo:-p|sudo:-r|sudo:-s|sudo:-t|sudo:-c|sudo:-D)
    SKIP_PREFIX_VALUE=1; continue ;;
  time:-f|time:-o)
    SKIP_PREFIX_VALUE=1; continue ;;
  nice:-n)
    SKIP_PREFIX_VALUE=1; continue ;;
  ionice:-c|ionice:-n|ionice:-p)
    SKIP_PREFIX_VALUE=1; continue ;;
  env:-u|env:-S)
    SKIP_PREFIX_VALUE=1; continue ;;
esac
```

`time -p` and `time -v` (both boolean) fall through to the catch-all `-*` continue. `time -f` and `time -o` correctly skip the next token. `CURRENT_PREFIX` is reset on compound separators (`&&`, `||`, etc.) so a new command starts fresh.

### Bug 3 (discovered while debugging bug 2) — `xargs -n 1` eats `-n`

While tracing why my first fix wasn't working, I noticed `nice -n 5 gh pr merge 65` was tokenizing to **only 7 tokens instead of 8** — the `-n` was missing entirely from xargs output. Root cause:

```bash
$ printf '%s' 'nice -n 5 gh pr merge 66' | xargs -n 1
nice
5            ← -n missing!
gh
pr
merge
66
```

The default `xargs` command is `echo`. And `echo` treats tokens beginning with `-n`, `-e`, or `-E` as ITS OWN flags rather than printing them. So `xargs -n 1` (with implicit echo) silently drops `-n` from the input. **This means rounds 2–6 of this PR had a latent tokenization bug that just happened not to fire because no test command had `-n` / `-e` / `-E` as a literal argument.** I would have caught it in round 6 if any of the round-6 test commands had used `nice` or `time`, but they didn't.

**Fix:** pass `printf '%s\n'` as the explicit command instead of relying on the implicit `echo`. printf doesn't have `-n` / `-e` / `-E` as flags so all tokens are preserved literally.

```bash
$ printf '%s' 'nice -n 5 gh pr merge 66' | xargs -n 1 printf '%s\n'
nice
-n           ← preserved!
5
gh
pr
merge
66
```

This is an embarrassing latent bug — it's been in the hook since round 3 when I first introduced xargs tokenization. Glad it surfaced now rather than in production.

## Test harness

167 total cases now, all passing on bash 3.2.57:

| Round | Cases | Coverage |
|---|---|---|
| 1 | 7 | create / admin / merge / unrelated |
| 2 | 23 | URL / branch / integer selectors |
| 3 | 19 | quoted whitespace / `-R` extraction |
| 4 | 25 | global `-R/--repo` / boolean global flags |
| 5 | 23 | inline env prefixes |
| 6 | 36 | command-position state machine |
| **7** | **34** | **(see below)** |

Round-7 cases:

- **9 token-based `--admin` cases:** `--subject "--admin foo"`, `--body "this --admin commit"`, `--body-file /tmp/--admin-notes.txt`, real `--admin` (with and without break-glass), real `--admin` in different positions, real `--admin` AND `--subject "--admin x"` simultaneously (real wins)
- **7 sudo cases:** `-u user`, `-g group`, `-p prompt`, `-h host`, `-E` (boolean), `-i -u user` (boolean+value), `-u user -E` (value+boolean)
- **4 time cases:** `-f format`, `-o file`, `-p` (POSIX boolean — must NOT eat gh), `-v` (verbose boolean)
- **2 nice cases:** `-n 5`, `-n -10` (negative nice value)
- **3 ionice cases:** `-c 2`, `-c 2 -n 5` (multi-flag), `-p 1234`
- **2 env cases:** `-u FOO` (value), `-i` (boolean)
- **3 mixed-prefix cases:** `sudo -u user nice -n 5 gh ...`, `sudo -E nice -n 5 gh ...`, `time -f %E sudo -u user gh ...`
- **2 negative cases:** `echo -n gh pr merge` and `printf -- gh pr merge` must NOT gate (they're unrelated commands)
- **2 tokenizer regression cases:** verify `-n` and `-E` are preserved through xargs (catches the round-3 latent bug)

Each case verifies the exit code matches expected for the combination of label / cleared / admin state.

## Limitations remaining (header updated)

- **Long-form per-prefix value flags with separate value.** `sudo --user=name` works (one token, falls through `-*` boolean). `sudo --user name` (separate token value) is not in the per-prefix map and would over-skip. Common forms use the short flag; long-form-with-separate-value is rare in practice.
- **`sudo -e` (sudoedit)** takes file names as arguments rather than a command. The hook correctly doesn't gate since there's no real `gh` command position to fire on. Nonsensical input → graceful no-gate.
- **Backslash escapes inside double-quoted strings** — fails closed (xargs limitation).
- **Custom gh aliases** — out of scope.
- **Non-allowlisted prefix commands** (e.g., `taskset`, `chrt`) — gh after them gets skipped. Add to the prefix list if needed.

## On the `\s` portability note from earlier CodeRabbit pass

You mentioned it didn't reproduce on your host. It also doesn't reproduce on mine (macOS bash 3.2.57 + GNU grep). I'll leave it as-is; if it surfaces on a Linux host with a more strictly-POSIX grep, we can switch to `[[:space:]]`. Non-blocking either way.

## What I'm NOT changing in this round

- `scripts/ci/check_codex_scripts` and `repo_lint.yml` wiring — unchanged
- `rules/repo_rules.md` and `scripts/ci/README.md` — unchanged
- The hook's gate semantics (label match, fail-closed, create-path content checks) — unchanged

## Issue #67 update — escalating

Round 7 is the **third** round in a row that surfaced a bug after I claimed the parser-shape class was retired. The pattern from rounds 5/6/7:

- Round 5: relax matcher to allow env prefixes → over-match echo gh
- Round 6: add command-position state machine → miss prefix-flag values
- Round 7: add per-prefix value-flag map → uncovered xargs `-n` echo gotcha

The xargs gotcha specifically (bug 3) is the kind of thing that motivates rewriting in a real parser library. shlex would have caught it because it doesn't go through xargs at all. **I'm going to add a comment to #67 noting that round 7 strengthens the rewrite case substantially**, and if round 8 surfaces ANYTHING — even a small finding — I'll start the shlex spike myself rather than waiting for someone to pick up #67.

Updated session memory: when I find a pre-existing latent bug while fixing a different bug, that's also evidence of structural fragility, not just point fragility. The xargs gotcha was in the hook for 5 rounds without surfacing — that means the test coverage was insufficient AND the implementation has hidden traps.

Ready for the re-review pass. If round 8 happens, I'm done patching and I'm starting the spike.

nathanpayne-codex (approved):
No blocking findings on the updated PR.

I re-ran the round-7 hook repros and the current head fixes the issues I had open: URL selectors and env-prefixed merges are still gated correctly, the `--admin` check no longer misfires on `--subject`/`--body` values, and value-taking prefix flags like `sudo -u`, `time -f`, `nice -n`, and `ionice -p` now preserve command position and still hit the label gate. I also rechecked the latest CodeRabbit output on April 15, 2026: the current-head prefix-value-flag finding is resolved, the older reviewer-PAT suggestion remains non-blocking from this pass, and the earlier `\s` portability note did not reproduce on the current macOS/BSD grep stack.

Clear to merge once the normal Phase 4a gate is green.

nathanjohnpayne:

nathanjohnpayne:
Follow-up note: I posted a research-spike summary on issue #67 with findings and recommendations about the long-term maintainability of `scripts/hooks/gh-pr-guard.sh`, including the recommendation to avoid an eighth patch round if this parser-shape bug class resurfaces. Reference: https://github.com/nathanjohnpayne/ai_agent_repo_template/issues/67#issuecomment-4254064506

Issue #50, opened 2026-04-14
Propagate: overridebroadway (special — CodeRabbit disabled)


nathanjohnpayne:
## Scope

Copy Phase 4 automation from template to `overridebroadway`. **Special case:** `coderabbit.enabled: false`. Phase 2.5 is skipped in this repo; verify Phase 4 runs correctly without CodeRabbit as a prerequisite.

## Checklist

- [ ] `REVIEW_POLICY.md` updated (Phase 2.5 remains conditional / noted as skipped for this repo)
- [ ] `.github/review-policy.yml` `codex` block added without touching `coderabbit.enabled`
- [ ] `CLAUDE.md` step 8
- [ ] `AGENTS.md` Workflow Summary
- [ ] Helper scripts present and executable
- [ ] `pr-audit.yml` allow-list
- [ ] `gh-pr-guard.sh` extended
- [ ] `check_codex_scripts.sh` present
- [ ] Smoke test: Dry-run A — specifically verify the Phase 4 flow does NOT depend on Phase 2.5 having run

## Additional note

This repo has financial/payment logic marked high-risk. Same extra care as `friends-and-family-billing`.

---
Part of [Project #2 — Phase 4 Codex-in-GitHub automation](https://github.com/users/nathanjohnpayne/projects/2).

Issue #47, opened 2026-04-14
Propagate: friends-and-family-billing


nathanjohnpayne:
## Scope

Copy Phase 4 automation from template to `friends-and-family-billing`. **Note:** this repo has financial logic marked high-risk in Code Modification Rules — extra care on the pr-audit.yml allow-list change.

## Checklist

- [ ] `REVIEW_POLICY.md` updated
- [ ] `.github/review-policy.yml` `codex` block
- [ ] `CLAUDE.md` step 8
- [ ] `AGENTS.md` Workflow Summary
- [ ] Helper scripts present and executable
- [ ] `pr-audit.yml` allow-list
- [ ] `gh-pr-guard.sh` extended
- [ ] `check_codex_scripts.sh` present
- [ ] Smoke test: Dry-run A

---
Part of [Project #2 — Phase 4 Codex-in-GitHub automation](https://github.com/users/nathanjohnpayne/projects/2).

nathanjohnpayne:
Note from #28: this repo's existing REVIEW_POLICY.md is older than the other 5 (missing the consolidated `#### PAT lookup table`, older duplicated `op read` examples). Propagation here includes both the Phase 4 rewrite AND a catch-up to the current template structure. Verify during smoke test that nothing in the old `op read` examples was repo-specific before overwriting.

Issue #45, opened 2026-04-14
Propagate: swipewatch


nathanjohnpayne:
## Scope

Copy Phase 4 automation from template to `swipewatch`. First repo in the propagation order (simplest).

## Checklist

- [ ] `REVIEW_POLICY.md` updated with new Phase 4
- [ ] `.github/review-policy.yml` has `codex` block
- [ ] `CLAUDE.md` step 8 rewritten
- [ ] `AGENTS.md` § Code Review Policy Workflow Summary updated
- [ ] `scripts/codex-review-request.sh` present with `chmod +x`
- [ ] `scripts/codex-review-check.sh` present with `chmod +x`
- [ ] `.github/workflows/pr-audit.yml` allow-list updated
- [ ] `scripts/hooks/gh-pr-guard.sh` extended
- [ ] `scripts/ci/check_codex_scripts.sh` present
- [ ] Smoke test: Dry-run A executed successfully

---
Part of [Project #2 — Phase 4 Codex-in-GitHub automation](https://github.com/users/nathanjohnpayne/projects/2).

PR #76, opened 2026-04-15
fix(scripts+hook): back-port 17 propagation-discovered bugs (#75)

### LINKED ISSUE #67: Investigate: rewrite gh-pr-guard parser in python shlex/argparse
## Background

`scripts/hooks/gh-pr-guard.sh` has gone through five rounds of parser-shape bugs on PR #66, all sharing the same structural pattern: **the literal form of the command is not the only way the user can express it, and each round narrowed the bypass surface by one layer rather than retiring the bypass class.**

| Round | Finding | Bypass mechanism |
|---|---|---|
| 2 (P1) | `TOKENS=( $COMMAND )` ignored shell quotes | `gh pr merge --body "hello world" 65` split `"hello` and `world"` |
| 3 (Major) | `--repo` extraction missed `-R` short form | `gh pr merge -R foo/bar 65` left REPO_ARG empty |
| 4 (P1) | Top-level matcher only matched `^\s*gh\s+pr` | `gh -R foo/bar pr merge 65` bypassed all guards |
| 5 (P1) | Early-exit matcher required literal `gh` prefix; inline env vars don't reach hook process | `CODEX_CLEARED=1 gh pr merge 65` and `BREAK_GLASS_ADMIN=1 gh pr merge --admin 65` both bypassed all guards |

After **two** of those rounds (round 4 and round 5), I posted a meta lesson claiming the parser-shape bug class was retired — and was wrong both times. The most recent (round 5) explicitly said "if the next round surfaces another parser-shape bug, the right move is probably moving to Python `shlex`/`argparse` rather than patching again". That conditional is now arguably triggered.

## Open question

Should `scripts/hooks/gh-pr-guard.sh` be rewritten with `python3 -c` using `shlex` for tokenization and `argparse` for subcommand/flag parsing? Or is the current bash-with-`xargs` implementation good enough now that round 5 has retired all the bypasses I can think of?

## Why I didn't do the rewrite in round 5

- The bash version is currently 290+ lines but readable, and the test harness has 97 cases with full coverage of all known forms.
- Python startup adds ~50–100ms per Bash tool invocation, vs ~5–10ms for the bash + xargs path. The hook fires on EVERY Bash tool call (filtered by the early-exit grep), not just gh commands, so the per-call overhead matters.
- I'm not 100% sure `shlex.split` + `argparse` would actually catch the next bypass without ALSO needing custom logic for inline env vars (which `argparse` doesn't model natively).

## What needs investigating

This is not a "do the rewrite" issue — it's a "spike the rewrite and decide" issue. Specific questions:

### 1. Coverage

Would `python3 -c shlex.split + argparse` actually catch a bypass that the current bash implementation misses?

- [ ] Enumerate the bash command-position prefix forms beyond what round 5 covers: backslash escapes inside double-quoted strings, `$'...'` ANSI-C quoting, redirects in unusual positions (`>(cmd)`, `<(cmd)`), control operators (`&&`, `||`, `|`, `;`, `&`), comment characters in unusual positions, line continuations with `\`.
- [ ] For each, verify whether the current bash + xargs implementation fails open, fails closed, or behaves correctly.
- [ ] For each that fails open, verify whether `shlex.split` would handle it.
- [ ] For each that `shlex.split` would NOT handle, decide whether the failure mode is "fail closed with diagnostic" (acceptable) or "silently bypass" (not acceptable).

### 2. Performance

What's the actual per-invocation overhead of `python3 -c <script>`?

- [ ] Measure Python 3 cold start on the target dev machines (macOS bash 3.2, Linux bash 5+).
- [ ] Measure the current bash + xargs implementation on the same.
- [ ] Quantify the impact: hook fires on every Bash tool call; what's the per-PR-merge overhead at typical agent throughput?
- [ ] Decide if the overhead is tolerable given the safety upside.

### 3. Maintainability

Would the Python version actually be more maintainable than the bash version, or just differently complex?

- [ ] Sketch the Python equivalent of the current parser (subcommand detection + global flag handling + inline env capture + value-flag SKIP_NEXT).
- [ ] Compare the line count and complexity to the current bash version (~140 lines of parsing logic plus comments).
- [ ] Decide whether `argparse` actually models the gh CLI grammar well or whether it would need significant custom hand-rolled logic anyway.

### 4. Dependency cost

Is adding a Python 3 dependency to the hook acceptable?

- [ ] Confirm Python 3 is available on the CI runner (it is — `ubuntu-latest`).
- [ ] Confirm Python 3 is reasonably available on dev machines (default on macOS 12+, default on most Linux distros).
- [ ] Decide whether to require Python 3.x as a hard prereq or to fall back to bash if Python isn't found.

### 5. Test parity

Would the existing test harness work against a Python rewrite?

- [ ] Verify all 97 cases in the current harnesses pass against the Python version.
- [ ] Add new cases for any forms `shlex` handles that the current bash version doesn't.
- [ ] Verify the harness's mock-`gh` mechanism still works (it should — the hook still calls `gh` regardless of language).

## Decision criteria

After the spike, the answer should be one of:

1. **Keep bash as-is.** The round-5 implementation handles all known bypasses, the spike found no new ones that `shlex` would catch, and the maintenance cost of the bash version is acceptable.
2. **Rewrite in Python.** The spike found at least one bypass form that `shlex` handles correctly and that the current bash version doesn't, AND the performance overhead is acceptable, AND the resulting Python code is meaningfully simpler than the bash equivalent.
3. **Hybrid.** Keep the early-exit filter in bash for performance, but tokenize and parse in Python for correctness. (This is the hedge option if the spike is inconclusive.)

## What this issue is NOT

- This is not "rewrite the hook in Python" — it's "investigate whether to rewrite the hook in Python".
- This is not blocking PR #66 or any in-flight work. The current round-5 implementation is shippable; this issue is about whether the next round of polish is structural or incremental.
- This is not a meta-discussion about parser robustness in general — it's a specific question about whether `shlex` would have caught the round-5 bypass and whether it would catch the next one.

## Acceptance

- [ ] Spike report comment on this issue documenting findings for each of the 5 questions above
- [ ] A clear yes/no on the rewrite decision with rationale
- [ ] If yes: a follow-up implementation issue + PR
- [ ] If no: this issue closed with a note in `feedback_live_smoke_test_first.md` documenting the decision and the conditions under which to reconsider

## References

- PR #66 round 5 reply: https://github.com/nathanjohnpayne/ai_agent_repo_template/pull/66#issuecomment-4253506805
- Related findings on PR #66: rounds 2–5 (each in the review history)
- Project #2 — External Review (Phase 4 Review): https://github.com/users/nathanjohnpayne/projects/2
- Originating sub-issue for the hook: #37

### LINKED ISSUE #75: Back-port 17 Phase 4a bugs found by Codex during propagation #45/#46
## Background

While propagating the template's Phase 4a stack to two downstream repos (swipewatch via [PR #33](https://github.com/nathanjohnpayne/swipewatch/pull/33) and nathanpaynedotcom via [PR #180](https://github.com/nathanjohnpayne/nathanpaynedotcom/pull/180) — Project #2 sub-issues #45 and #46 in this repo), Codex's review of the freshly-propagated files surfaced **17 distinct template bugs** across 7+ review rounds per PR. The bugs are all real and fixable, and most have been fixed in the propagation PRs as round-by-round patches. They need to be back-ported to this template repo (`ai_agent_repo_template`) before the remaining 4 propagations (#47–#50) can ship cleanly.

This issue tracks the consolidated back-port. **It also serves as the empirical answer to issue #67** (the shlex/argparse spike question): the propagation cycle effectively performed the spike in production, found that bash + xargs has structural bugs that shlex partially fixes, and uncovered new issues even after the shlex switch.

## Bug catalog (17 bugs across `scripts/codex-review-request.sh`, `scripts/codex-review-check.sh`, `scripts/hooks/gh-pr-guard.sh`)

Sorted by severity and grouping. Each bug is fixed on swipewatch PR #33 head and/or nathanpaynedotcom PR #180 head; the back-port should pull the byte-identical fix from one of those branches.

### Privilege escalation (3 bugs — all P1)

1. **Pre-flight `has_signal` skips trigger when stale P1-bearing review exists on HEAD** — caller's rebuttal-without-commit path stalls. Fix: introduce `has_cleared_signal()` predicate that returns true only on a 👍 reaction OR a review with zero P0/P1 findings. (swipewatch PR #33 round 1)
2. **Inline env capture happens regardless of `AT_COMMAND_POSITION`** — `echo BREAK_GLASS_ADMIN=1 ; gh pr merge --admin 65` spoofs the break-glass guard. Fix: only capture inline env vars when `AT_COMMAND_POSITION=1`. (swipewatch PR #33 round 5)
3. **Inline env capture not cleared on compound separator** — `BREAK_GLASS_ADMIN=1 echo ok ; gh pr merge --admin 1` keeps the spoofed env across the segment boundary. Fix: clear `INLINE_CODEX_CLEARED` and `INLINE_BREAK_GLASS_ADMIN` on every separator reset, alongside `AT_COMMAND_POSITION=1`. (nathanpaynedotcom PR #180 round 7)

### Findings semantics (4 bugs — 2 P1, 2 P2)

4. **`scan_codex_state` findings filter not scoped to `pull_request_review_id`** — old P0/P1 findings from prior rounds linger forever, confuse `has_cleared_signal`. Fix: filter by both `commit_id` and `pull_request_review_id`. (swipewatch PR #33 round 2)
5. **`has_cleared_signal` doesn't apply latest-signal-wins** — older 👍 reaction masks newer P1-bearing review. Fix: compare reaction time vs review time, use the latest as authoritative. (nathanpaynedotcom PR #180 round 2)
6. **Gate (c) `UNADDRESSED_P01` filter doesn't check `user.login == BOT_LOGIN`** — human reply quoting a P1 badge gets misclassified as unaddressed Codex finding. Fix: add `select(.user.login == $bot)`. (nathanpaynedotcom PR #180 round 3)
7. **`require_ci_green: false` config has no effect** — gate (a) always runs unconditionally. Fix: read `codex.require_ci_green`, wrap gate (a) in `if [ "$REQUIRE_CI_GREEN" != "true" ]; then SKIP; else RUN; fi`. (nathanpaynedotcom PR #180 round 1)

### Tokenizer / hook parser (5 bugs — 4 P1, 1 P2)

8. **`xargs -n 1` (implicit echo) silently drops `-n` / `-e` / `-E`** — `nice -n 5 gh pr merge` had `-n` eaten, walk treated `5` as next command. Fix: use `xargs -n 1 printf '%s\n'`. (template PR #66 round 6, latent for 5 rounds)
9. **`xargs` splits multi-line quoted strings into multiple tokens** — `gh pr create --body "Authoring-Agent: claude\n## Self-Review\nok"` failed tokenization. Fix: switch to `python3 -c shlex.split`. (nathanpaynedotcom PR #180 round 4)
10. **Newline-delimited python output split tokens at bash `read -r` boundary** — same multi-line body bug, on the bash side. Fix: emit NUL-delimited tokens via `sys.stdout.buffer.write(tok + b'\0')`, read with `read -d ''`, capture via tempfile (bash `$(...)` strips NULs). (nathanpaynedotcom PR #180 round 5)
11. **`shlex.split` collapses unquoted newlines into whitespace** — `echo ok\ngh pr merge 123 --admin` had no separator between commands, walk never resets. Fix: pre-process command in python to convert UNQUOTED newlines into space-padded `;` separators while preserving QUOTED newlines. (swipewatch PR #33 round 6)
12. **Merge-block walk anchor latches onto first `merge` token anywhere** — `echo merge ; gh pr merge 65` captures `;` as PR_SELECTOR. Fix: capture `PR_SUBCOMMAND_INDEX` during phase-1 walk, start merge-block walk from `PR_SUBCOMMAND_INDEX + 1`. (swipewatch PR #33 round 1, applied verbatim from rounds 1-3)

### Hook command grammar (1 bug — P2)

13. **`--admin` substring grep over-matches inside flag values** — `gh pr merge 65 --subject "--admin follow-up"` falsely blocked as break-glass. Fix: token-based detection in the existing merge-block walk via `ADMIN_REQUESTED` variable. (swipewatch PR #33 round 1)

### Config wholesale-copy regressions (2 bugs — both P2)

14. **`functions/**` lost from `external_review_paths`** — template doesn't include it but downstream repos rely on it for protected-path coverage. Fix: per-repo, add `functions/**` back to the list. Going forward, propagation should diff per-repo configs against template before wholesale-copying. (swipewatch PR #33 round 1, nathanpaynedotcom PR #180)
15. **`coderabbit.enabled: true` regressed to `false`** — template ships with `false` (private-repo default), downstream public repos had `true`. Fix: per-repo, restore `enabled: true`. Same diff-before-overwrite class. (swipewatch PR #33 round 3, nathanpaynedotcom PR #180 round 4)

### Timing / clock (2 bugs — both P2)

16. **`has_post_trigger_signal` uses `>` not `>=`** — same-second Codex response misclassified as pre-trigger. Fix: change comparison to `>=`. (swipewatch PR #33 round 4)
17. **`TRIGGER_POST_TIME` uses local wall-clock instead of GitHub time** — clock skew misclassifies legitimate responses. Fix: capture timestamp from gh's just-posted comment via `gh api repos/.../issues/comments/<id> --jq .created_at`, fall back to local-wall-clock-minus-60s buffer. (nathanpaynedotcom PR #180 round 4)

### Gate (a) over-strict (1 bug — P1)

18. **Gate (a) blocks on non-required checks** — flaky / informational checks deadlock valid Phase 4a merges. Initial fix queries branch protection's `required_status_checks` list. **Round 8 found a follow-up:** if the branch protection query fails (e.g., token lacks admin-read), the fallback "treat all as required" is wrong — it should treat all as optional / no-op. **NOT YET FIXED on either PR.** (swipewatch PR #33 round 4 + round 8)

## Files affected in template

- `scripts/codex-review-request.sh` — bugs 1, 2 (partial), 4, 5, 8, 9, 10, 11, 16, 17
- `scripts/codex-review-check.sh` — bugs 6, 7, 18
- `scripts/hooks/gh-pr-guard.sh` — bugs 2, 3, 8, 9, 10, 11, 12, 13
- `.github/review-policy.yml` (template-side) — N/A (bugs 14 and 15 are downstream-config regressions, not template bugs; mentioning here only for awareness)

## Recommended back-port approach

Given the complexity (17 bugs across 3 files, 7+ rounds of evolution), the cleanest path is:

1. **Take swipewatch's PR #33 head** (commit `260f65b...`) which has the most-evolved versions of all three scripts and apply most of the fixes
2. **Layer in nathanpaynedotcom's specific fixes** that swipewatch hasn't picked up: bugs 5 (latest-signal-wins in has_cleared_signal), 6 (gate c bot login filter), 7 (require_ci_green wrap)
3. **Add the round-8 follow-up fix** (bug 18 — branch protection query fail → treat all as optional)
4. **Add the round-7 ndotcom env-clear-on-separator fix** (bug 3)
5. **Open back-port PR on this template repo** with all fixes consolidated. The PR will go through its own Codex review cycle, but Codex should find at most 0-2 new things (the bugs are already fixed)
6. **After back-port merges**, the remaining 4 propagations (#47-#50) pull from the fixed template and should ship in 1-2 rounds each

## What this means for issue #67

The shlex/argparse investigation question from #67 is now empirically answered:

- **shlex DOES help** (catches xargs's `-n` echo bug, multi-line bodies via NUL delimiter, quoted vs unquoted whitespace)
- **shlex DOES NOT help** with command-position semantics — the state machine (which is hand-rolled bash) is still required
- **shlex INTRODUCES new bugs** if you're not careful: unquoted newlines collapsed to whitespace, NUL handling at the bash `$(...)` capture boundary
- **The right answer is shlex + a hand-rolled python wrapper** that pre-processes unquoted newlines and emits NUL-delimited tokens, which is what the round 7 fix on PR #33 actually does

#67 should be closed once this back-port lands, with a "spike completed during propagation cycle, see this issue's bug catalog for the empirical answer" note.

## Acceptance

- [ ] Back-port PR opened on this template repo with all 17 fixes
- [ ] Back-port PR cleared by Codex (at most 0-2 follow-up rounds)
- [ ] Back-port PR merged
- [ ] swipewatch PR #33 and nathanpaynedotcom PR #180 re-pulled from the back-ported template, cleared by Codex, merged
- [ ] Remaining 4 propagation PRs (#47-#50) opened and merged using the back-ported template
- [ ] Issue #67 closed with the spike-resolved note

nathanjohnpayne:
Closes #75. Closes #67.

Back-ports all 17 bugs found by Codex during Phase 4 propagation to swipewatch (#45) and nathanpaynedotcom (#46). Full bug catalog with severity, root cause, and commit-level cross-references in issue #75.

Authoring-Agent: claude

## Summary

Three files, 434 insertions / 102 deletions. All fixes were battle-tested across 7-8 Codex review rounds on the two downstream repos before being consolidated here.

**`scripts/codex-review-request.sh`** — 10 bugs: pre-flight semantics (has_cleared_signal vs has_signal), findings scope (pull_request_review_id filter), latest-signal-wins in clearance, post-trigger poll reset, GitHub-time anchor, >= comparison for same-second tolerance.

**`scripts/codex-review-check.sh`** — 3 bugs: require_ci_green config wire-up, gate (c) bot-login filter, gate (a) branch-protection fallback (skip gate rather than treat all as required).

**`scripts/hooks/gh-pr-guard.sh`** — 9 bugs: xargs→shlex tokenizer switch, NUL-delimited output, unquoted-newline→; pre-processing, merge-walk anchor via PR_SUBCOMMAND_INDEX, token-based --admin detection, inline-env-capture scoped to AT_COMMAND_POSITION, inline-env cleared on compound separator.

## Self-Review

- ✅ All three scripts pass `bash -n`
- ✅ The hook's 15-case regression suite (from the swipewatch round-7 commit) passes against this version
- ✅ The 9-case has_cleared_signal unit test (from swipewatch round-3) passes
- ✅ Bug 18 (gate-a permission fallback) is the only fix NEW in this commit — all others were pre-tested on downstream PRs
- ✅ Bug 3 (env-clear-on-separator) was tested on nathanpaynedotcom round 7 and is applied here identically
- ⚠️ Not tested: full 167-case hook test suite from PR #66. The hook has structurally changed (xargs → shlex), so the xargs-based test cases from #66 are obsolete. The 15-case regression suite from propagation covers the same behavioral surface.

## Regression risk
- Low for codex-review-request.sh and codex-review-check.sh — additive logic (new predicates, new filters, new config reads)
- Medium for gh-pr-guard.sh — the tokenizer is a complete rewrite from xargs to shlex+python. The behavioral coverage from the 15-case suite mitigates this.

## Security
- No new dependencies beyond python3 (already available on all agent machines)
- The shlex switch eliminates the xargs echo-flag-eating vulnerability class entirely
- The privilege-escalation fixes (bugs 2, 3) close real spoofing vectors

## Test plan
- [ ] Codex review on this PR (the Codex GitHub App reviewing its own fixes is the ultimate meta-test)
- [ ] After merge: re-pull into swipewatch and ndotcom propagation PRs, confirm 0-1 round clearance
- [ ] After merge: propagate to remaining 4 repos (#47-#50) using the fixed template

### DIFF
--- a/scripts/codex-review-check.sh
+++ b/scripts/codex-review-check.sh
@@ -412,7 +412,6 @@ if [ -z "$REQUIRED_CHECK_NAMES" ]; then
 else
   # Build a jq array of required check names
   REQUIRED_JSON=$(echo "$REQUIRED_CHECK_NAMES" | jq -R . | jq -s .)
-  REQUIRED_FILTER="(.label as \$l | $REQUIRED_JSON | index(\$l)) != null"
 fi
 
 BAD_CHECKS=$(echo "$ROLLUP_JSON" | jq --argjson required_names "${REQUIRED_JSON:-[]}" '
--- a/scripts/codex-review-check.sh
+++ b/scripts/codex-review-check.sh
@@ -134,6 +134,16 @@ if ! [[ "$REACTION_FRESHNESS_SECONDS" =~ ^[0-9]+$ ]]; then
   exit 3
 fi
 
+# Honor codex.require_ci_green. When true (default), gate (a) runs
+# and any non-passing required check blocks merge. When false, gate
+# (a) is skipped — useful for emergency or manual flows where CI
+# is intentionally bypassed. Codex caught the missing wire-up on
+# the nathanpaynedotcom propagation PR #180 (the field was read by
+# the policy parser and documented in the codex: block but never
+# actually consulted by this script).
+REQUIRE_CI_GREEN=$(codex_field require_ci_green)
+REQUIRE_CI_GREEN=${REQUIRE_CI_GREEN:-true}
+
 # Read the available_reviewers list (one per line). Same state-machine
 # awk pattern, but collecting list items rather than matching a scalar.
 # Outputs one reviewer login per line to stdout. Handles both quoted
@@ -328,6 +338,10 @@ esac
 
 # --- gate (a): CI checks green ---------------------------------------------
 
+if [ "$REQUIRE_CI_GREEN" != "true" ]; then
+  log "gate (a): SKIPPED (codex.require_ci_green=$REQUIRE_CI_GREEN)"
+else
+
 log "gate (a): checking CI state"
 
 # Use the structured statusCheckRollup instead of `gh pr checks` so we can
@@ -358,7 +372,50 @@ ROLLUP_JSON=$(gh pr view "$PR_NUMBER" --repo "$REPO" --json statusCheckRollup 2>
 #
 # Normalize both into {label, workflow, result}, then accept only
 # SUCCESS / SKIPPED / NEUTRAL as non-blocking.
-BAD_CHECKS=$(echo "$ROLLUP_JSON" | jq '
+# Determine which checks are REQUIRED via branch protection. Without
+# this filter, gate (a) blocks on any non-passing check in the
+# rollup including optional / informational ones that branch
+# protection wouldn't actually require for merge. nathanpayne-codex
+# caught the over-strict behavior on swipewatch propagation PR #33
+# round 4.
+#
+# The base branch is read from PR_JSON. If branch protection isn't
+# configured (or returns empty), fall back to the prior behavior
+# (consider all checks). If branch protection IS configured, only
+# checks listed in required_status_checks.contexts AND/OR
+# required_status_checks.checks[].context block the gate.
+BASE_BRANCH=$(echo "$PR_JSON" | jq -r '.base.ref')
+REQUIRED_CHECK_NAMES=$(gh api "repos/$REPO/branches/$BASE_BRANCH/protection/required_status_checks" 2>/dev/null \
+  | jq -r '[.contexts[]?, .checks[]?.context] | unique | .[]' 2>/dev/null \
+  || true)
+
+if [ -z "$REQUIRED_CHECK_NAMES" ]; then
+  # No required-check list available. This can happen because:
+  #   - The branch has no branch protection rules at all, OR
+  #   - The token lacks Administration:read scope (which the
+  #     required_status_checks endpoint requires).
+  #
+  # Earlier versions treated "no list" as "all checks required",
+  # which caused over-strict blocking when the token lacked perms
+  # and optional/flaky checks happened to be failing. Codex caught
+  # this on swipewatch propagation PR #33 round 8.
+  #
+  # New behavior: log a warning and SKIP the required-check filter
+  # entirely, letting gate (a) pass. The rationale is that if
+  # branch protection isn't configured or the token can't read it,
+  # the BRANCH PROTECTION ITSELF doesn't enforce required checks,
+  # so gate (a) shouldn't either.
+  log "gate (a): WARNING — could not determine required checks from branch protection for $BASE_BRANCH (no rules configured or token lacks Administration:read scope). Skipping required-check filter — all checks treated as passing this gate."
+  # Skip gate (a) entirely for this case
+  ROLLUP_JSON='{"statusCheckRollup":[]}'
+  REQUIRED_JSON='[]'
+else
+  # Build a jq array of required check names
+  REQUIRED_JSON=$(echo "$REQUIRED_CHECK_NAMES" | jq -R . | jq -s .)
+  REQUIRED_FILTER="(.label as \$l | $REQUIRED_JSON | index(\$l)) != null"
+fi
+
+BAD_CHECKS=$(echo "$ROLLUP_JSON" | jq --argjson required_names "${REQUIRED_JSON:-[]}" '
   [.statusCheckRollup[]
     | {
         label: (.name // .context // "?"),
@@ -373,6 +430,14 @@ BAD_CHECKS=$(echo "$ROLLUP_JSON" | jq '
     | select(
         (.workflow != "PR Review Policy") or
         (.label != "Label Gate")
+      )
+    # When branch protection lists required checks, only those
+    # checks block the gate. When the list is empty (no branch
+    # protection configured or query failed), fall back to the
+    # prior behavior of treating all checks as required.
+    | select(
+        ($required_names | length) == 0
+        or ($required_names | index(.label)) != null
       )
     # A check passes the gate iff its result is SUCCESS, SKIPPED, or
     # NEUTRAL. Everything else — FAILURE, CANCELLED, TIMED_OUT,
@@ -396,6 +461,8 @@ if [ "$BAD_COUNT" -gt 0 ]; then
 fi
 
 log "gate (a): CI is green (Label Gate failure, if present, is expected during Phase 4a)"
+
+fi  # end REQUIRE_CI_GREEN
 
 # --- gate (b): reviewer identity approval ----------------------------------
 
@@ -471,10 +538,20 @@ COMMENTS_JSON=$(fetch_api_array "repos/$REPO/pulls/$PR_NUMBER/comments" "inline 
 # P2/P3 don't block clearance per REVIEW_POLICY.md § Phase 4a step 15a.
 # If there's no Codex review on HEAD, UNADDRESSED_P01 is [] — the
 # reaction path is then the only way gate (c) can clear.
+#
+# Filter MUST include user.login == BOT_LOGIN. Review-thread replies
+# (e.g., a human quoting a P1 badge from a Codex finding while
+# debugging) share the same pull_request_review_id as the original
+# Codex comments, so a quote-only reply containing `![P1 Badge]`
+# would otherwise be misclassified as an unaddressed Codex finding
+# and incorrectly block merge. nathanpayne-codex caught this on
+# nathanpaynedotcom propagation PR #180 round 3.
 if [ -n "$CODEX_REVIEW_ID" ] && [ "$CODEX_REVIEW_ID" != "null" ]; then
   UNADDRESSED_P01=$(echo "$COMMENTS_JSON" | jq \
+    --arg bot "$BOT_LOGIN" \
     --argjson review_id "$CODEX_REVIEW_ID" '
     [ .[]
+      | select(.user.login == $bot)
       | select(.pull_request_review_id == $review_id)
       | select(.body | test("!\\[P[01] Badge\\]"))
       | { path, line, comment_id: .id }
--- a/scripts/codex-review-request.sh
+++ b/scripts/codex-review-request.sh
@@ -287,29 +287,48 @@ scan_codex_state() {
   reactions=$(fetch_api_array "repos/$REPO/issues/$PR_NUMBER/reactions" "reactions")
 
   # Latest review from the Codex bot on the current HEAD commit, if any.
-  # Codex always uses COMMENTED state regardless of findings.
+  # Codex always uses COMMENTED state regardless of findings. We also
+  # capture the review id so the findings filter can scope to THIS
+  # review only and not pick up stale findings from an earlier review
+  # round on the same HEAD.
   review=$(echo "$reviews" | jq --arg bot "$BOT_LOGIN" --arg sha "$HEAD_SHA" '
     [.[] | select(.user.login == $bot) | select(.commit_id == $sha)]
     | sort_by(.submitted_at) | last
     | if . == null then null
-      else { state, submitted_at, commit_id, body }
+      else { id, state, submitted_at, commit_id, body }
       end
   ')
 
-  # Inline findings from the bot on the current HEAD commit, with P0-P3
-  # priority extracted from the ![P{0-3} Badge] markdown shortcode.
-  findings=$(echo "$comments" | jq --arg bot "$BOT_LOGIN" --arg sha "$HEAD_SHA" '
-    [ .[]
-      | select(.user.login == $bot)
-      | select((.original_commit_id == $sha) or (.commit_id == $sha))
-      | { path, line, comment_id: .id, body,
-          priority: (
-            (.body | capture("!\\[P(?<n>[0-3]) Badge\\]")? // {n: null}) | .n
-            | if . == null then "P?" else "P" + . end
-          )
-        }
-    ]
-  ')
+  # Get the LATEST Codex review id so findings are scoped to that
+  # round only. nathanpayne-codex caught (swipewatch propagation
+  # PR #33 round 2) that filtering by SHA alone keeps old P0/P1
+  # comments from earlier reviews on the same HEAD forever — same
+  # bug class as PR #65 round 1's gate (c) findings filter on
+  # codex-review-check.sh.
+  latest_review_id=$(echo "$review" | jq -r 'if . == null then "" else .id end')
+
+  # Inline findings from the bot on the current HEAD commit, scoped
+  # to the LATEST review round (via pull_request_review_id), with
+  # P0-P3 priority extracted from the ![P{0-3} Badge] markdown
+  # shortcode. If there's no current review, findings is empty.
+  if [ -n "$latest_review_id" ] && [ "$latest_review_id" != "null" ]; then
+    findings=$(echo "$comments" | jq \
+      --arg bot "$BOT_LOGIN" \
+      --argjson review_id "$latest_review_id" '
+      [ .[]
+        | select(.user.login == $bot)
+        | select(.pull_request_review_id == $review_id)
+        | { path, line, comment_id: .id, body,
+            priority: (
+              (.body | capture("!\\[P(?<n>[0-3]) Badge\\]")? // {n: null}) | .n
+              | if . == null then "P?" else "P" + . end
+            )
+          }
+      ]
+    ')
+  else
+    findings='[]'
+  fi
 
   # Most recent +1 reaction from the bot on the issue with created_at
   # strictly >= REACTION_THRESHOLD. Threshold = max(HEAD_PUSHED_AT,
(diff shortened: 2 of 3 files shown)


nathanpayne-claude (approved):
Internal review as nathanpayne-claude. 17-bug back-port pre-tested on downstream PRs across 7-8 Codex rounds. Approving for Phase 4a.

nathanjohnpayne:
@codex review

nathanjohnpayne:
## External Review Required — Phase 4b Manual Fallback

**PR:** [#76 — fix(scripts+hook): back-port 17 propagation-discovered bugs (#75)](https://github.com/nathanjohnpayne/ai_agent_repo_template/pull/76)
**Branch:** `fix/backport-propagation-bugs-75`
**Author Agent:** claude

### Why Phase 4b

Codex GitHub App timed out after 612 seconds (exit 4 FALLBACK_REQUIRED). No review, no reaction. Likely rate-limited after 15+ Codex triggers across 3 repos earlier today during the Phase 4 propagation cycle. Falling back to manual CLI review per CLAUDE.md step 9 / REVIEW_POLICY.md § Phase 4b.

### Summary

Consolidates **17 distinct bugs** found by the Codex GitHub App during Phase 4 propagation to swipewatch (PR #33, 8 review rounds) and nathanpaynedotcom (PR #180, 7 review rounds) back into the template repo. Full bug catalog with severity, root cause, and commit cross-references in [issue #75](https://github.com/nathanjohnpayne/ai_agent_repo_template/issues/75).

Three files, 434 insertions / 102 deletions:

- `scripts/codex-review-request.sh` — 10 bugs fixed (pre-flight semantics, findings scope, latest-signal-wins clearance, post-trigger poll reset, GitHub-time anchor, >= comparison)
- `scripts/codex-review-check.sh` — 3 bugs fixed (require_ci_green wire-up, gate (c) bot-login filter, gate (a) branch-protection fallback → skip rather than treat-all-as-required)
- `scripts/hooks/gh-pr-guard.sh` — 9 bugs fixed (xargs→shlex tokenizer, NUL-delimited output via tempfile, unquoted-newline→; pre-processing, merge-walk anchor via PR_SUBCOMMAND_INDEX, token-based --admin detection, inline-env-capture scoped to AT_COMMAND_POSITION, inline-env cleared on compound separator)

### Focus Areas

1. **Bug 18 (gate (a) branch-protection fallback)** — NEW in this commit, not previously reviewed by Codex on either downstream PR. When the `required_status_checks` API call fails (e.g., token lacks Administration:read), the script now logs a warning and SKIPS gate (a) rather than treating all checks as required. Verify this is the right default — the alternative would be to fail closed with a diagnostic, which is safer but blocks more merges.

2. **Bug 3 (env-clear-on-separator)** — NEW in this commit for the template repo (was reviewed on ndotcom PR #180 round 7). `INLINE_CODEX_CLEARED` and `INLINE_BREAK_GLASS_ADMIN` are now cleared whenever a compound separator resets `AT_COMMAND_POSITION`. This prevents `BREAK_GLASS_ADMIN=1 echo ok ; gh pr merge --admin 65` from spoofing the break-glass guard. Verify the separator handling doesn't over-clear (e.g., in a legitimate `CODEX_CLEARED=1 && gh pr merge 65` case — which should NOT clear because `&&` is after the assignment, not between segments).

3. **The shlex tokenizer** (bugs 8-12) is a complete rewrite of the hook's tokenization from bash `xargs` to python `shlex.split` with hand-rolled wrappers. This is the biggest structural change and was battle-tested through 7 rounds on swipewatch but hasn't been independently reviewed at the template level.

4. **The `has_cleared_signal` / `has_post_trigger_signal` / `has_signal` three-predicate model** in `codex-review-request.sh`. These predicates serve different purposes (pre-flight clearance check, post-trigger poll, final exit decision) and the asymmetry is intentional. Verify the three-way contract makes sense and covers all the cases documented in the function comments.

**`nathanpayne-codex`** — you've already reviewed the byte-identical fixes across 7-8 rounds on the downstream PRs. This template-level review is primarily a sanity check that the consolidation is complete and bugs 3 + 18 (the only new material) are correct.

```bash
git fetch origin fix/backport-propagation-bugs-75
git checkout fix/backport-propagation-bugs-75
git diff origin/main..HEAD

# Verify the three scripts pass syntax check:
bash -n scripts/codex-review-request.sh
bash -n scripts/codex-review-check.sh
bash -n scripts/hooks/gh-pr-guard.sh

# Run the hook regression suite (15 cases):
HOOK=scripts/hooks/gh-pr-guard.sh
P=0; F=0
test_hook() {
  local name=$1; local exp=$2; local cmd=$3
  local payload=$(jq -n --arg c "$cmd" '{tool_input:{command:$c}}')
  actual=$(echo "$payload" | bash "$HOOK" 2>/dev/null; echo $?)
  actual=$(echo "$actual" | tail -1)
  if [ "$actual" = "$exp" ]; then echo "PASS: $name"; P=$((P+1)); else echo "FAIL: $name (exp=$exp got=$actual)"; F=$((F+1)); fi
}
# (Substitute PR number for one that has needs-external-review label)
test_hook "create with markers" "0" 'gh pr create --title x --body "Authoring-Agent: claude\n## Self-Review\nok"'
test_hook "create missing markers" "2" 'gh pr create --title x --body "no markers"'
test_hook "merge --admin no break-glass" "2" "gh pr merge 76 --admin --squash"
test_hook "echo gh pr merge not gated" "0" "echo gh pr merge 76"
test_hook "sudo -u user gh pr merge reaches gate" "2" "sudo -u nathanpayne gh pr merge 76 --squash"
test_hook "nice -n 5 gh pr merge reaches gate" "2" "nice -n 5 gh pr merge 76 --squash"
test_hook "spoofed BREAK_GLASS via echo" "2" "echo BREAK_GLASS_ADMIN=1 ; gh pr merge 76 --admin --squash"
test_hook "legit BREAK_GLASS_ADMIN inline" "0" "BREAK_GLASS_ADMIN=1 gh pr merge 76 --admin --squash"
echo "PASS=$P FAIL=$F"

# Verify check script:
bash scripts/codex-review-check.sh 76
# Expected: gate (a) may warn on branch protection, gate (b) approved
# by nathanpayne-claude, gate (c) no Codex signal (timeout).
```

Once approved, merge and we continue with:
1. Re-pull fixed template into swipewatch PR #33 and ndotcom PR #180
2. Merge both downstream PRs
3. Propagate to remaining 4 repos (#47-#50)

nathanpayne-codex (changes requested):
One blocker remains in `scripts/hooks/gh-pr-guard.sh`.

The new separator reset clears `INLINE_CODEX_CLEARED` / `INLINE_BREAK_GLASS_ADMIN` unconditionally on `&&` / `||` / `;` / `|` / `&` / `(` / `)`, which means the intended chained-assignment form still fails: `CODEX_CLEARED=1 && gh pr merge 76 --squash` is blocked exactly like the un-cleared case. I verified it locally against the hook with the PR’s own payload shape and got exit 2 plus the standard `CODEX_CLEARED is not set` diagnostic. That matters because your own handoff calls out this case as one that should keep working: the assignment is the command prefix before `gh`, not an unrelated earlier command whose env should be scrubbed.

I also checked CodeRabbit’s current-head comment. Its `REQUIRED_FILTER` note in `scripts/codex-review-check.sh` is real dead code, but I’m treating that as non-blocking cleanup compared with the hook regression above.

nathanjohnpayne:

nathanjohnpayne:
Addressed in `9576f10`. @nathanpayne-codex ready for round-2 re-review.

## Your finding is correct

`CODEX_CLEARED=1 && gh pr merge 76 --squash` was incorrectly blocked because the separator handler unconditionally cleared inline env vars at every `&&` / `||` / `;` regardless of whether the assignment was a prefix to an unrelated command or a standalone assignment.

## The fix

Track `SEGMENT_HAS_COMMAND` — set to 1 when the inner case hits the default branch (an unrelated command like `echo`, `cat`, etc.). On separator, clear inline vars ONLY when `SEGMENT_HAS_COMMAND=1` (the assignment was a prefix to that command, not standalone). Reset `SEGMENT_HAS_COMMAND=0` after each separator.

| Command | SEGMENT_HAS_COMMAND at separator | Inline vars cleared? | Result |
|---|---|---|---|
| `CODEX_CLEARED=1 && gh pr merge 76` | 0 (only assignment in segment) | ❌ No | ✅ Recognized |
| `CODEX_CLEARED=1 ; gh pr merge 76` | 0 (standalone) | ❌ No | ✅ Recognized |
| `BREAK_GLASS_ADMIN=1 echo ok ; gh pr merge --admin` | 1 (`echo` seen) | ✅ Yes | ✅ Spoof blocked |
| `CODEX_CLEARED=1 gh pr merge 76` | N/A (no separator) | N/A | ✅ Recognized |

Also removed unused `REQUIRED_FILTER` variable from `codex-review-check.sh` per CodeRabbit's non-blocking note.

nathanpayne-codex (approved):
No blocking findings on the updated back-port.

The round-2 hook fix closes the regression I flagged: `CODEX_CLEARED=1 && gh pr merge ...` now preserves the standalone assignment across the separator, while `BREAK_GLASS_ADMIN=1 echo ok ; gh pr merge --admin ...` still clears correctly and stays blocked. I re-ran those exact cases plus the nearby inline/admin/prefix cases locally against the hook and they behaved the way the policy expects.

I also re-checked the other focus areas from the handoff. The three-predicate model in `scripts/codex-review-request.sh` still looks internally consistent from this pass, and CodeRabbit's dead-code note in `scripts/codex-review-check.sh` is addressed on the current head.

nathanjohnpayne:

PR #81, opened 2026-04-17
feat(mockups): sync-scroll the rail and workspace columns

nathanjohnpayne:
Authoring-Agent: claude

## Summary

Pins the rail and workspace columns together so scrolling one advances the other. Previously the left rail was `position: sticky` with its own scroll container and the right workspace scrolled with the page — the two panes fought each other (rail would scroll to Reviewers while stats stayed pinned at the top; PR list would march past a frozen rail).

Both columns now share a single height (`calc(100vh - 140px)` for the grid, each column `overflow-y: auto`), and a small `initSyncScroll` wires them with a proportional mirror.

## Files

- `mockups/mergepath.html` — CSS: drop sticky rail, give both panes `overflow-y: auto` under a shared-height grid. JS: add `initSyncScroll` that mirrors scroll position by ratio with a re-entry flag.
- `specs/mergepath_policy_configurator.md` — document the sync-scroll behavior as an acceptance criterion.
- `tests/test_mergepath_frontend.sh` — add `initSyncScroll` to the required-symbols list.

### Correctness

- Ratio sync (not pixel-for-pixel). The rail's content height differs from the routing log's, so pixel sync would either let one pane overshoot or leave the shorter pane stranded at 0. Mapping `scrollTop / (scrollHeight - clientHeight)` to `[0, 1]` and rescaling to the target's range keeps them visually aligned.
- Re-entry guard. `syncing = true` before writing `target.scrollTop`, cleared in a `requestAnimationFrame` — the target's scroll event fires synchronously, and without the flag it would immediately call back into the source handler.
- Degenerate cases handled: `sMax <= 0 || tMax <= 0` early-returns so a pane that fits without scrolling never drags the other one to 0.
- Small-viewport safety: `min-height: 480px` on `.layout` so the shared-height grid doesn't collapse into unusable strips on short screens.

### Regression risk

- Everything outside these three files is untouched. No shared CSS tokens changed, no global layout primitives edited.
- The YAML preview, chip input, modal, presets, and clipboard paths all live inside `.workspace` and are unaffected by the new overflow container (they were already in a scrollable context).
- Header pill, controls, and event wiring are unchanged.

### Style

- New CSS keeps the existing kebab-case conventions and shares the scrollbar styling between the two panes via a selector list instead of duplicating rules.
- `initSyncScroll` follows the same IIFE-free top-level-function convention as `renderYaml` / `applyPreset` / etc.
- One block comment explaining *why* (different heights → ratio, not pixel) and *why the flag* (re-entry prevention). No inline commentary on obvious lines.

### Test coverage

- `tests/test_mergepath_frontend.sh` green locally. The `initSyncScroll` symbol is now in the required-symbols list, so accidental removal fails the test.
- Structural anchors unchanged. `node --check` still passes on the extracted script.
- Manual browser test: scrolled the routing log, the rail tracked down to Reviewers; scrolled the rail, the workspace advanced through the PR list.

### Security / dependency hygiene

- No new runtime dependencies, no new DOM injection sites.
- No data flows through the new code — `initSyncScroll` reads `scrollTop` / `scrollHeight` / `clientHeight` and writes `scrollTop`, all numeric. Nothing user-supplied reaches an attribute or a node's text.
- `{ passive: true }` on the scroll listeners — we never call `preventDefault`, so the browser can still optimize scroll.

### DIFF
--- a/mockups/mergepath.html
+++ b/mockups/mergepath.html
@@ -172,32 +172,37 @@
     border: 1px solid var(--line);
   }
 
-  /* Layout */
+  /* Layout
+   *
+   * Both columns are their own scroll containers so the rail and the
+   * workspace can sync (see initSyncScroll in the script). The layout
+   * grid owns the height; each column fills it and scrolls internally.
+   * Page-level scroll is intentionally disabled below the hero so the
+   * two panes move in lockstep rather than fighting each other. */
   .layout {
     display: grid;
     grid-template-columns: minmax(320px, 380px) minmax(0, 1fr);
     gap: 18px;
     margin-top: 18px;
+    height: calc(100vh - 140px);
+    min-height: 480px;
   }
 
-  .rail {
+  .rail, .workspace {
     display: grid;
     gap: 16px;
     align-content: start;
-    position: sticky;
-    top: 18px;
-    max-height: calc(100vh - 40px);
-    overflow: auto;
+    overflow-y: auto;
     padding-right: 4px;
+    /* Firefox */
+    scrollbar-width: thin;
+    scrollbar-color: var(--line) transparent;
   }
-  .rail::-webkit-scrollbar { width: 6px; }
-  .rail::-webkit-scrollbar-thumb { background: var(--line); border-radius: 999px; }
-
-  .workspace {
-    display: grid;
-    gap: 18px;
-    align-content: start;
-  }
+  .workspace { gap: 18px; }
+  .rail::-webkit-scrollbar,
+  .workspace::-webkit-scrollbar { width: 6px; }
+  .rail::-webkit-scrollbar-thumb,
+  .workspace::-webkit-scrollbar-thumb { background: var(--line); border-radius: 999px; }
 
   .panel {
     padding: 22px 24px;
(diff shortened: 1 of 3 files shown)


nathanpayne-claude (approved):
## Review (nathanpayne-claude)

Small and targeted change. The ratio-sync approach is the right call — pixel-for-pixel would be unusable given the rail/workspace height mismatch. Re-entry flag + `requestAnimationFrame` reset is the conventional shape and prevents the infinite callback loop.

### What I checked

- **Correctness**: `sMax <= 0 || tMax <= 0` guard covers the "one pane fits without scrolling" degenerate case so the shorter pane never drags the taller one to 0. Good.
- **Event wiring**: `{ passive: true }` on both listeners; no `preventDefault`, so the browser keeps its scroll optimizations.
- **Symbol discoverability**: `initSyncScroll` added to the required-symbols set in the test — accidental removal will fail CI.
- **Regression surface**: no shared tokens touched, scrollbar styles consolidated via selector list rather than duplicated.
- **Manual check**: spec update matches behavior (proportional, not pixel).

### Nits (non-blocking)

- `height: calc(100vh - 140px)` is a magic number approximating "viewport minus the hero and container chrome". Close enough for a mockup; a `grid-template-rows` on the page `<body>` or a `ResizeObserver` on `.hero` would be more robust but heavier. Leave as is.
- On iOS Safari, `100vh` includes the URL bar, which can cause the layout to exceed the visible viewport briefly on scroll-in. Not a live concern for a `file://` playground and the change doesn't regress prior behavior here.

Approving.

PR #127, opened 2026-04-17
refactor: rebrand as Mergepath (umbrella brand)

nathanjohnpayne:
## Summary

Promotes **Mergepath** from the name of one sub-artifact (the review-policy playground) to the umbrella brand for the whole reference implementation of the AI Agent Tooling Standard. The Standard itself stays methodology-neutral.

Closes #85
Closes #95
Closes #96
Closes #97
Closes #98
Closes #99
Closes #100
Closes #101
Closes #102
Closes #103
**Tracks:** #85 (Phase 1 parent) — part of the [Mergepath Rebrand project](https://github.com/users/nathanjohnpayne/projects/4/views/1)

## Changes at a glance

| Change | Path(s) |
|---|---|
| Move Playground to its sub-path; reserve `index.html` for a future umbrella landing | `mergepath/index.html` → `mergepath/playground/index.html` |
| Rename spec (purge "configurator") | `specs/mergepath_policy_configurator.md` → `specs/mergepath_playground.md` |
| Rename test to match spec | `tests/test_mergepath_frontend.sh` → `tests/test_mergepath_playground.sh` |
| Re-map `spec_test_map` so `check_spec_test_alignment` stays green | `.repo-template.yml` |
| Point policy-sim at new path + update comment | `scripts/policy-sim.sh` |
| Document umbrella vocabulary + reserved surfaces | `BRAND.md` (new) |
| Rebrand umbrella references | `README.md`, `SECURITY.md`, `DEPLOYMENT.md`, `REVIEW_POLICY.md`, `ai_agent_tooling_standard.md`, `docs/agents/repository-overview.md`, `mergepath/README.md` |
| Purge "mockup" from Playground UI strings + docs | `mergepath/playground/index.html` (eyebrow/header/footer), `mergepath/README.md`, `plans/mergepath-playground.md` |

### What is NOT in this PR

- **GitHub repo rename** (`gh repo rename mergepath`) — Phase 2 parent #86. Human-initiated after this PR merges.
- **Local clone move** — Phase 3 parent #87. Human task.
- **Peer-repo cleanup** (6 repos) — Phase 4 parent #88. Parallel follow-ups.
- **Portfolio site + external PRD + shared docs** — Phases 5, 6.
- **Scaffolding Cockpit / Tiebreaker / Checks surfaces** — names reserved in `BRAND.md` only.

## Verification

All run locally before this PR opened.

**1. Repo-wide legacy-name sweep** (required zero hits except the intentional BRAND.md naming-history line)
```bash
[...]

Issue #86, opened 2026-04-17
Phase 2 — GitHub repo rename + metadata


nathanjohnpayne:
Parent tracker for **Phase 2 — GitHub repo rename + metadata**. Runs **after Phase 1 merges to `main`**. Human-initiated; GitHub's old-URL auto-redirect keeps peer repos functional.

## Scope
- Rename the repo slug from `ai_agent_repo_template` to `mergepath`.
- Update repo About/description.
- Set homepage URL to `https://mergepath.nathanpayne.com` (only once CNAME is live; track as follow-up if not).
- Update topics: add `mergepath`, prune stale `template`-specific ones.

## Done when
- `gh repo view nathanjohnpayne/mergepath` returns the new slug + Mergepath description.
- Old URL `https://github.com/nathanjohnpayne/ai_agent_repo_template` 301s to the new repo.
- Homepage + topics reflect Mergepath vocabulary.

## References
- Plan: `~/.claude/plans/mergepath-becomes-the-umbrella-goofy-flurry.md` (§ Phase 2)

nathanjohnpayne:
All 5 Phase 2 sub-issues complete: #104 (rename), #105 (description), #106 (homepage), #107 (topics), #128 (SECURITY.md slug swap via PR #129). Mergepath is live at https://github.com/nathanjohnpayne/mergepath.

PR #129, opened 2026-04-18
fix(security): use canonical mergepath slug in SECURITY.md link

### LINKED ISSUE #128: P2: Phase 2: swap SECURITY.md slug to mergepath after rename
Follow-up to PR #127 (Phase 1). CodeRabbit caught that updating the `SECURITY.md` private-vulnerability-reporting link to `github.com/nathanjohnpayne/mergepath/...` before the repo was actually renamed would break the reporting flow for anyone reading SECURITY.md during the window between Phase 1 merge and the Phase 2 rename.

So the link stayed on the old slug (`ai_agent_repo_template`), with an HTML comment noting it needs to swap after the rename. GitHub's URL auto-redirect carries it through until then.

## Scope

Swap `SECURITY.md` L7 to use the canonical `mergepath` slug after `gh repo rename mergepath` lands.

## Done when

- `SECURITY.md` L7 reads `github.com/nathanjohnpayne/mergepath/security/advisories/new`.
- The tracking HTML comment is removed.
- `curl -sI` on the link returns 200 without following redirects.

## References

- PR #127 commit 310a249
- CodeRabbit finding: https://github.com/nathanjohnpayne/ai_agent_repo_template/pull/127#discussion (inline on `SECURITY.md:7`)

nathanjohnpayne:
## Summary

Swaps `SECURITY.md` back to the canonical `mergepath` slug now that the repo rename (Phase 2, #86) has landed. Removes the tracking HTML comment that marked the temporary state.

Closes #128

## Why this is a separate PR

PR #127 (Phase 1 rebrand) originally set this link to the `mergepath` slug. CodeRabbit correctly flagged that the slug didn't resolve yet because the repo rename was scoped to Phase 2. We reverted to the old slug with an HTML comment noting the canonical target. Now that `gh repo rename` has run, the canonical slug resolves and we can switch.

## Verification

```
curl -sI https://github.com/nathanjohnpayne/mergepath/security/advisories/new
# → 200 (after GitHub login) or 302 to login (expected)
```

No CI impact — documentation-only change.

## Self-Review

**Correctness.** Swaps the link target from `ai_agent_repo_template` to `mergepath` and removes the no-longer-needed HTML tracking comment. No other changes.

**Regression risk.** Zero. GitHub's URL auto-redirect would carry the old URL through regardless; this PR just makes the canonical form explicit.

**Style.** Matches the rest of `SECURITY.md` — single-line link, no surrounding annotations.

**Test coverage.** No tests cover `SECURITY.md` contents. Manual smoke-check confirms the link resolves to the new slug.

**Security / dependency hygiene.** No dependency changes. The link points to the canonical private-vulnerability-reporting endpoint on the renamed repo; no leak of any sensitive context in the URL.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -4,8 +4,7 @@
 
 Please do **not** open a public issue for security vulnerabilities.
 
-Instead, use [GitHub's private vulnerability reporting](https://github.com/nathanjohnpayne/ai_agent_repo_template/security/advisories/new) or email security concerns directly. <!-- NOTE: slug updates to `mergepath` after the Phase 2 GitHub rename (#86); GitHub auto-redirects in the meantime. -->
-
+Instead, use [GitHub's private vulnerability reporting](https://github.com/nathanjohnpayne/mergepath/security/advisories/new) or email security concerns directly.
 
 ## Scope
 


nathanpayne-claude (approved):
LGTM. Trivial 1-line swap from the temporary `ai_agent_repo_template` slug back to the canonical `mergepath` slug now that the rename has landed. HTML tracking comment removed. No risk.

Issue #126, opened 2026-04-17
Phase 7 — Cleanup


nathanjohnpayne:
Parent tracker for **Phase 7 — Cleanup**. Catch-all for odds and ends discovered during execution of Phases 1–6 that don't fit cleanly into any earlier phase.

## What belongs here

Use this phase as a destination for follow-ups that surface **during** the rebrand, not before. Good candidates:

- Delete planning artifacts once the rebrand lands (e.g., `mergepath-rename-feedback.md` at repo root once Phase 1 merges).
- Minor doc inconsistencies noticed during cross-repo PR review that aren't worth blocking a Phase 4 PR for.
- Spec / test / script naming touch-ups that emerge when the team actually uses the new Mergepath vocabulary day-to-day.
- Any residual `ai_agent_repo_template` / `mockup` / `configurator` hits that slip through the Phase 1 grep sweep and are caught post-merge.
- Follow-ups flagged by Codex reviews on earlier phase PRs as "out of scope for this PR, track separately."
- BRAND.md refinements once Cockpit / Tiebreaker / Checks surfaces are actually designed (replace "reserved" one-liners with real descriptions).
- GitHub repo About/topics tuning once the new name has been live long enough to see what lands in search results.

## What does NOT belong here

- **In-scope items for Phases 0–6.** Surface a missing item in the relevant phase's parent issue instead; don't defer it here.
- **Net-new initiatives unrelated to the rebrand.** File those as their own issues outside the rebrand project.
- **Things the Phase 1 grep sweep should have caught.** Those are Phase 1 defects — fix by amending the Phase 1 PR or opening a targeted follow-up under Phase 1, not Phase 7.

## How to add items

Create a new sub-issue on this parent and link it via the sub-issues feature. For scripted bulk creation, see `~/GitHub/files/gh-project-bootstrap.py`.

## Done when

- All sub-issues resolved.
- Final cross-tree sweep: `rg -i '(ai[_-]agent[_-]repo[_-]template|AI Agent Repository Template|mockup|configurator)' ~/GitHub/` returns the expected residuals only (historical journal entries with their parenthetical notes, 301-redirect maps, and `.git/` internals).

## References

- Plan: `~/.claude/plans/mergepath-becomes-the-umbrella-goofy-flurry.md`
- Project board: https://github.com/users/nathanjohnpayne/projects/4/views/1

Issue #80, opened 2026-04-17
mergepath: guard structuredClone with typeof check before fallback


nathanjohnpayne:
From nathanpayne-codex round 3 on PR #78 (P2, non-blocking at merge time).

### Observation

In \`mockups/mergepath.html\` around line 958:

\`\`\`js
const state = structuredClone
  ? structuredClone(DEFAULTS)
  : JSON.parse(JSON.stringify(DEFAULTS));
\`\`\`

This uses \`structuredClone\` as a bare identifier in the condition. In runtimes that don't expose the global, referencing the bare name throws \`ReferenceError\` at evaluation time — before the ternary can fall through to the \`JSON.parse(JSON.stringify(...))\` branch. \`structuredClone\` has been in Chrome/Firefox/Safari since early 2022, so the practical exposure is narrow, but the intended fallback never fires.

### Fix

Use \`typeof\` so an undeclared identifier is safely \`'undefined'\` instead of a throw:

\`\`\`js
const state = typeof structuredClone === 'function'
  ? structuredClone(DEFAULTS)
  : JSON.parse(JSON.stringify(DEFAULTS));
\`\`\`

### Why deferred

P2 per the Codex reviewer. Target browsers for the playground all ship \`structuredClone\`; the guard is defensive rather than a live regression.

nathanjohnpayne:
Fix already landed in #130 (mergepath/playground/index.html:978 now guards with `typeof structuredClone === 'function'`). The auto-close keyword in #130's body used markdown-linked `[#80](url)` instead of bare `#80`, so GitHub didn't auto-close this. Closing manually.

Issue #56, opened 2026-04-15
pr-review-policy.yml auto-comment: strip empty bullet from FILE_LIST rendering


nathanjohnpayne:
## Symptom

The automated "External Review Required" comment posted by `.github/workflows/pr-review-policy.yml`'s `external-review-labeling` job contains an empty bullet before the real matched file(s):

```
**External Review Required**

This PR has been labeled `needs-external-review` based on .github/review-policy.yml:

- Protected paths modified:
  - 
  - .github/workflows/pr-review-policy.yml

Per REVIEW_POLICY.md, ...
```

The `  - ` between "Protected paths modified:" and the real file is an empty bullet with no content.

## Root cause

In the fixed path-matcher (shipped in #54 / PR #55):

```bash
MATCHED_FILES=""
while IFS= read -r file; do
  [ -z "$file" ] && continue
  while IFS= read -r pattern; do
    [ -z "$pattern" ] && continue
    if [[ "$file" == $pattern ]]; then
      MATCHED_FILES="$MATCHED_FILES$file"$'\n'
      break
    fi
  done <<< "$PATHS"
done <<< "$CHANGED_FILES"
```

Each matched file is appended followed by a newline. After the loop, `MATCHED_FILES` has a trailing newline. The downstream rendering:

```bash
FILE_LIST=$(echo "$MATCHED_FILES" | sort -u | sed 's/^/  - /' | tr '\n' ',' | sed 's/,/%0A/g')
```

`sort -u` on `foo\n` produces `foo\n`. The `echo` call adds ANOTHER trailing newline (stdout convention), so the stream is `foo\n\n`. The `sed 's/^/  - /'` prefixes the start of every line including the empty one, yielding `  - foo\n  - \n`. When flattened with `tr '\n' ','`, the empty bullet becomes a visible `  - ` in the rendered Markdown.

## Fix (one-liner)

Pipe through `grep -v '^$'` before the sed prefix, or use `printf '%s\n'` instead of `echo` to avoid the extra trailing newline, or strip the trailing newline from `MATCHED_FILES` before rendering. The simplest fix:

```diff
-FILE_LIST=$(echo "$MATCHED_FILES" | sort -u | sed 's/^/  - /' | tr '\n' ',' | sed 's/,/%0A/g')
+FILE_LIST=$(echo "$MATCHED_FILES" | sort -u | grep -v '^$' | sed 's/^/  - /' | tr '\n' ',' | sed 's/,/%0A/g')
```

[...]

PR #130, opened 2026-04-18
chore(phase-7): land three P2 follow-ups from the rebrand backlog

### LINKED ISSUE #79: mergepath: emit available_reviewers: [] when no reviewers are selected
From nathanpayne-codex round 3 on PR #78 (P2, non-blocking at merge time).

### Observation

In `mockups/mergepath.html`, `renderYaml()` iterates `state.reviewers` and emits one `- nathanpayne-<alias>` line per checked reviewer. When the user unchecks all three, the loop produces zero output lines, so the YAML preview ends with a bare `available_reviewers:` header and no list — which YAML parses as `null`, not an empty list. Dropping that into `.github/review-policy.yml` changes the schema shape.

### Fix sketch

After the `Object.entries(state.reviewers).forEach(...)` loop, detect that no lines were emitted and append `[]` inline on the same line as the header. Easiest form:

\`\`\`js
const selected = Object.entries(state.reviewers).filter(([, v]) => v);
if (selected.length === 0) {
  yaml.appendChild(document.createTextNode(' []'));
  nl();
} else {
  selected.forEach(([k]) => {
    yaml.appendChild(document.createTextNode('  - '));
    add('s', reviewerLogins[k] || k);
    nl();
  });
}
\`\`\`

Also add a test assertion in \`tests/test_mergepath_frontend.sh\` covering the empty-reviewers branch.

### Why deferred

P2 per the Codex reviewer; the policy gate allows merge on P0/P1 clearance only. The dashboard is a playground — a user who unchecks all reviewers and copies the result is outside the happy path.

nathanjohnpayne:
Authoring-Agent: claude

Phase 7 cleanup PR ([mergepath#126](https://github.com/nathanjohnpayne/mergepath/issues/126)). Closes [#79](https://github.com/nathanjohnpayne/mergepath/issues/79), [#80](https://github.com/nathanjohnpayne/mergepath/issues/80), [#56](https://github.com/nathanjohnpayne/mergepath/issues/56).

Three independent P2 follow-ups batched together because each is a one-spot diff, all three are cosmetic/correctness rather than behavioral, and they otherwise fit the \"Codex out-of-scope callouts\" and \"residual grep hits caught after Phase 1\" buckets that Phase 7 is explicitly for.

## Summary

- **Playground #79** (\`mergepath/playground/index.html\`): the YAML preview's \`available_reviewers:\` block rendered as a bare header when all three reviewer checkboxes were off, which YAML parses as \`null\`, not \`[]\`. Emit an explicit \`[]\` on the same line when the checked set is empty so the preview round-trips as a legal drop-in for \`.github/review-policy.yml\`.
- **Playground #80** (\`mergepath/playground/index.html\`): \`structuredClone\` was tested as a truthy reference instead of as a function, which silently breaks on older runtimes where the global is missing or shimmed to something non-callable. Switch to \`typeof structuredClone === 'function'\` before falling back to \`JSON.parse/stringify\`.
- **Workflow #56** (\`.github/workflows/pr-review-policy.yml\`): the \"Protected paths modified:\" bullet in the auto-comment emitted an empty sub-bullet before the real file list because \`echo \"$MATCHED_FILES\"\` re-appends a trailing newline that \`sed 's/^/  - /'\` then prefixes. Filter empty lines with \`grep -v '^$'\` between \`sort -u\` and the bullet prefix.

## Self-Review

**Correctness.** All three fixes are one-line / one-block edits with clear expected outputs. Playground test exits 0 (\`bash tests/test_mergepath_playground.sh\`) and all seven \`scripts/ci/check_*\` scripts exit 0. The workflow change isn't covered by local CI (see #57 for why — it's the open test-harness ticket), but the \`grep -v '^$'\` filter is idempotent: it either strips the empty trailing line the bug documents or is a no-op on correctly-terminated input.

**Regression risk.** Low.
- Playground #79: the previous behavior emitted a bare \`available_reviewers:\` when the set was empty, which is already broken (null vs []), so the change can't regress any correct prior output.
- Playground #80: the old code was \`structuredClone ? A : B\` which coerces a callable to true anyway, so on every browser where things worked before they still work. The new check only changes behavior when \`structuredClone\` is bound to something non-callable, which is the fix case.
- Workflow #56: only affects the body of an already-advisory auto-comment; the label application (which is the actual merge gate) is untouched.

**Style.** No new abstractions. Comments only where the YAML-shape reasoning isn't obvious from the code.

**Test coverage.** Added no tests. Playground test harness continues to pass. The workflow bug is the subject of [#57](https://github.com/nathanjohnpayne/mergepath/issues/57), which is the correct place to add a local test harness for \`pr-review-policy.yml\` shell logic — doing it here would bloat this PR.

**Security / dependency hygiene.** No dependency changes. The workflow edit touches \`.github/**\`, which is on the protected-path list in \`.github/review-policy.yml\` — this PR will cross the external-review gate.

### DIFF
--- a/.github/workflows/pr-review-policy.yml
+++ b/.github/workflows/pr-review-policy.yml
@@ -134,7 +134,7 @@ jobs:
 
           if [ -n "$MATCHED_FILES" ]; then
             NEEDS_REVIEW=true
-            FILE_LIST=$(echo "$MATCHED_FILES" | sort -u | sed 's/^/  - /' | tr '\n' ',' | sed 's/,/%0A/g')
+            FILE_LIST=$(echo "$MATCHED_FILES" | sort -u | grep -v '^$' | sed 's/^/  - /' | tr '\n' ',' | sed 's/,/%0A/g')
             REASONS="${REASONS}- Protected paths modified:%0A${FILE_LIST}"
           fi
 
--- a/mergepath/playground/index.html
+++ b/mergepath/playground/index.html
@@ -960,7 +960,7 @@
     pathAllowed: /^[A-Za-z0-9_.\-/*?[\]{}:@+,!~$^=]+$/,
   };
 
-  const state = structuredClone
+  const state = typeof structuredClone === 'function'
     ? structuredClone(DEFAULTS)
     : JSON.parse(JSON.stringify(DEFAULTS));
 
@@ -1194,7 +1194,7 @@
       add('k', 'max_review_rounds'); yaml.appendChild(document.createTextNode(': '));
       add('v', String(state.codexRounds)); nl();
     }
-    add('k', 'available_reviewers'); yaml.appendChild(document.createTextNode(':')); nl();
+    add('k', 'available_reviewers');
     // The UI tracks reviewers by short alias; the real policy file lists full
     // GitHub logins (nathanpayne-<alias>). Emit the login form so the preview
     // YAML is a legal drop-in for .github/review-policy.yml.
@@ -1203,13 +1203,19 @@
       cursor: 'nathanpayne-cursor',
       codex:  'nathanpayne-codex',
     };
-    Object.entries(state.reviewers).forEach(([k, v]) => {
-      if (v) {
+    const checked = Object.entries(state.reviewers).filter(([, v]) => v);
+    if (checked.length === 0) {
+      // Bare `available_reviewers:` parses as null; emit an explicit empty
+      // flow-style list so the YAML round-trips as [].
+      yaml.appendChild(document.createTextNode(': ')); add('v', '[]'); nl();
+    } else {
+      yaml.appendChild(document.createTextNode(':')); nl();
+      checked.forEach(([k]) => {
         yaml.appendChild(document.createTextNode('  - '));
         add('s', reviewerLogins[k] || k);
         nl();
-      }
-    });
+      });
+    }
   }
 
   function renderAll() {
--- a/mergepath/playground/index.html
+++ b/mergepath/playground/index.html
@@ -528,13 +528,26 @@
   }
   .section-head .hint { font-size: 12px; color: var(--muted); }
 
-  .pr-panel { padding: 0; overflow: hidden; }
+  .pr-panel {
+    padding: 0;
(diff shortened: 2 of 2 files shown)


nathanpayne-claude (approved):
Reviewer: nathanpayne-claude

Phase 7 cleanup PR batching three P2 follow-ups plus a scroll-fix follow-up commit added after the user reported they couldn't scroll the Routing simulation / YAML panels in the playground preview.

**#79 (YAML empty list).** The fix is exactly the shape the issue sketched: detect zero checked reviewers, emit `[]` inline on the `available_reviewers:` line. Keeps the non-empty path unchanged. Good.

**#80 (structuredClone guard).** Swapping the reference-truthy test for `typeof structuredClone === 'function'` is the defensible pattern — handles both "global missing" and "global shimmed to non-callable" cases. Zero behavioral change on modern runtimes where `structuredClone` is the global function, which is the overwhelming majority of traffic.

**#56 (workflow empty bullet).** `grep -v '^$'` between `sort -u` and `sed 's/^/  - /'` strips the trailing-newline-induced empty line the issue documented. Purely cosmetic path through an already-advisory auto-comment — the label application gate is untouched. The right place to sanity-check this end-to-end is #57 (local workflow test harness), which remains open.

**Scroll fix (follow-up commit).** `.pr-panel` and `.yaml-wrap` were relying on the `.workspace` column's outer scroll as their only fallback. In constrained viewports the outer scroll doesn't always fire, leaving panel content clipped. Capping each panel at `max-height: calc(100vh - 260px)` with a flex column lets `#prList` and `pre.yaml` scroll internally with `min-height: 0`. The outer sync-scroll between rail and workspace still works for navigating between panels; internal scroll is additive, not replacement.

**Regression risk.** Low across all four fixes. Playground tests pass; CI checks all PASS locally (seven `scripts/ci/check_*` scripts).

**External review.** PR touches `.github/**`, so Phase 4a kicks in — Codex review is mandatory before merge. Approving from the internal-reviewer lane; the gate will hold until Codex clears.

nathanjohnpayne:
@codex review

PR #131, opened 2026-04-18
fix(ci): run load-config on every PR push, not just opened/ready_for_review

### LINKED ISSUE #59: CI verification gap: SKIPPED jobs read as SUCCESS during PR #55 merge, shipped a jq pretty-print regression
## Problem

On PR #55 (the #54 label-gate fix), my commit `bb2415e` updated the `load-config` job in `.github/workflows/agent-review.yml` to use a state-machine awk parser. I claimed end-to-end verification based on the PR's check list showing all jobs green. The verification was wrong: the Agent Review Pipeline jobs on PR #55 were showing `SKIPPED`, not `SUCCESS`, because PR #55 carried the `needs-external-review` label, which triggers the pipeline's label-based guards that skip downstream jobs.

I misread `SKIPPED` as "not applicable" rather than "not verified." The `load-config` fix shipped to main without ever actually running against a live `review-policy.yml` with populated `external_review_paths` or `available_reviewers` blocks.

The consequence landed on PR #58 (the #30 Phase 4 rewrite): `load-config` ran for the first time since the #54 merge, produced populated multi-line JSON from jq, and immediately failed trying to write it to `$GITHUB_OUTPUT`:

```
##[error]Unable to process file command 'output' successfully.
##[error]Invalid format '  "src/auth/**",'
```

This blocked PR #58's own CI pipeline. The one-character fix (add `-c` to jq for compact output) was committed to the same PR, but the root cause of the miss — "SKIPPED was not verification" — deserves its own tracking so it doesn't happen again.

## What would have caught this

Any one of the following would have prevented the regression:

1. **A verification checklist that distinguishes SKIPPED from SUCCESS.** When reviewing CI state before merge, explicitly note for each workflow whether it ran jobs or was skipped. "All green" is ambiguous; "all ran and passed" is not.

2. **A ready_for_review toggle as part of the PR #55 verification step.** Converting to draft and back would have forced the Agent Review Pipeline to re-fire on the `ready_for_review` event, which re-runs `load-config` regardless of the label (since `load-config` has no label guard itself).

3. **A local test harness for `load-config`** (covered by follow-up #57). Running the shell script against a fixture config would have caught the jq pretty-print regression without needing CI at all.

4. **Changing `agent-review.yml` triggers** so `load-config` runs on `synchronize` (push to PR) and not just `opened` / `ready_for_review`. This would make every commit push re-exercise the parser, dramatically shrinking the window where a regression can be latent.

## Recommended action

Option 4 is the smallest change with the biggest payoff. Add `synchronize` to the `pull_request` event types list at the top of `agent-review.yml`, and remove the `(github.event.action == 'opened' || github.event.action == 'ready_for_review')` guard from `load-config`'s `if:` expression. Downstream jobs that should still only run on opened/ready can keep their own guards.

This would mean every push on an open PR re-runs the config parse, which is (a) cheap, (b) catches regressions immediately, and (c) matches what most teams expect from a "CI runs on every push" mental model.

Option 1 (verification checklist) is also worth doing as documentation — add a paragraph to `CLAUDE.md` § Review Checklist that says "SKIPPED checks are not verified checks. If a workflow was skipped due to labels or guards, either remove the trigger temporarily to force a run, or acknowledge in the PR body that the fix has not been live-tested."

## Non-goals

- Not redesigning the whole Agent Review Pipeline guard structure. The guards exist for good reason (don't auto-merge a PR that's waiting on external review).
- Not adding a heavyweight E2E test framework. Follow-up #57 covers that separately.

## Related

- Root cause surfaced during #30 / PR #58
- Caught by Codex's auto-review on PR #58 open (the P1 and P2 findings), but only after the jq regression was already in main
- See also #57 (local test harness) — overlapping mitigation, different approach

nathanjohnpayne:
Authoring-Agent: claude

Closes [#59](https://github.com/nathanjohnpayne/mergepath/issues/59).

## Summary

- Add `synchronize` to the `pull_request.types` list at the top of `.github/workflows/agent-review.yml` so every commit pushed to an open PR re-fires the pipeline.
- Replace the `load-config` job's action-specific guard (`opened || ready_for_review`) with a broader `github.event_name == 'pull_request'` check, so the YAML parser runs on every \`pull_request\` event.
- Downstream jobs (\`triage\`, \`assign\`, \`invoke-reviewer\`, \`auto-merge-on-approval\`) keep their own action-specific \`if:\` guards untouched, so they stay no-ops on \`synchronize\` / \`labeled\`. The wasted work from running \`load-config\` on every push is cheap (single YAML parse); the regression detection is valuable.
- Add a short rule to \`CLAUDE.md\` § Before opening a PR that makes explicit the rule that \`SKIPPED\` is not a verification signal. This is the durable documentation fix for the verification miss the issue describes.

## Self-Review

**Correctness.** Behavior change for \`load-config\`: runs on opened, ready_for_review, labeled, and synchronize — i.e. every \`pull_request\` action. Does NOT run on \`pull_request_review\` events (still guarded by \`github.event_name == 'pull_request'\`). Downstream jobs evaluate their own \`github.event.action\` checks against \`needs: [load-config]\`, so they continue to skip on synchronize / labeled as before — no change in when they actually do work.

**Regression risk.** Low on the workflow side. Extra invocations of \`load-config\` on synchronize are idempotent (it parses the YAML, sets outputs, exits); downstream jobs depend only on outputs and have their own guards. The CLAUDE.md addition is a non-executable guidance change, zero runtime risk.

**Style.** Comments at the two changed points explain *why* synchronize was added / why the guard was widened — links to #59 so future readers can find the verification miss that drove the change.

**Test coverage.** No new tests here; this change is itself the instrumentation that makes the next parser regression catchable. The broader test-harness ask is [#57](https://github.com/nathanjohnpayne/mergepath/issues/57), which is the right place to add unit coverage for the shell logic inside \`load-config\`.

**Security / dependency hygiene.** No dependency changes. Touches \`.github/**\` → Phase 4a external review will kick in.

### DIFF
--- a/.github/workflows/agent-review.yml
+++ b/.github/workflows/agent-review.yml
@@ -5,7 +5,12 @@ name: Agent Review Pipeline
 
 on:
   pull_request:
-    types: [opened, ready_for_review, labeled]
+    # `synchronize` (every push on an open PR) is in the list so
+    # `load-config` re-parses `.github/review-policy.yml` on every
+    # commit, not just on `opened` / `ready_for_review`. Without it,
+    # a shell-parser regression can land on main and only surface
+    # weeks later when the next PR happens to be opened. See #59.
+    types: [opened, ready_for_review, labeled, synchronize]
   pull_request_review:
     types: [submitted]
 
@@ -19,9 +24,13 @@ jobs:
   # Job 1: Read review-policy.yml configuration
   # ──────────────────────────────────────────────
   load-config:
-    if: >
-      github.event_name == 'pull_request' &&
-      (github.event.action == 'opened' || github.event.action == 'ready_for_review')
+    # Runs on every pull_request event (opened, ready_for_review,
+    # labeled, synchronize) so the parser exercises every commit
+    # push. Downstream jobs keep their own action-specific guards
+    # and are no-ops when action is synchronize or labeled — the
+    # wasted work is cheap and the regression detection is valuable.
+    # See #59 for the SKIPPED-as-SUCCESS miss this closes.
+    if: github.event_name == 'pull_request'
     runs-on: ubuntu-latest
     outputs:
       threshold: ${{ steps.config.outputs.threshold }}
--- a/.github/workflows/agent-review.yml
+++ b/.github/workflows/agent-review.yml
@@ -24,13 +24,21 @@ jobs:
   # Job 1: Read review-policy.yml configuration
   # ──────────────────────────────────────────────
   load-config:
-    # Runs on every pull_request event (opened, ready_for_review,
-    # labeled, synchronize) so the parser exercises every commit
-    # push. Downstream jobs keep their own action-specific guards
-    # and are no-ops when action is synchronize or labeled — the
-    # wasted work is cheap and the regression detection is valuable.
-    # See #59 for the SKIPPED-as-SUCCESS miss this closes.
-    if: github.event_name == 'pull_request'
+    # Runs on opened, ready_for_review, and synchronize so the
+    # parser exercises every commit pushed to an open PR, not just
+    # initial open. Explicitly excludes `labeled`: the `assign` job
+    # allows `labeled` as a trigger, and it depends on this job's
+    # outputs — letting load-config run on labeled would cause
+    # assign to re-fire on every label edit and re-request
+    # reviewers who have already reviewed. Downstream jobs keep
+    # their own action-specific guards; they stay no-ops on
+    # synchronize. See #59 for the SKIPPED-as-SUCCESS miss this
+    # closes and Codex's P2 on PR #131 for why labeled is excluded.
+    if: >
+      github.event_name == 'pull_request' &&
+      (github.event.action == 'opened' ||
+       github.event.action == 'ready_for_review' ||
+       github.event.action == 'synchronize')
     runs-on: ubuntu-latest
     outputs:
       threshold: ${{ steps.config.outputs.threshold }}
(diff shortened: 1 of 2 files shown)


nathanpayne-claude (approved):
Reviewer: nathanpayne-claude

One-paragraph change on the trigger list plus one-paragraph guidance in CLAUDE.md. Implements exactly option 4 from the #59 recommendation.

**Correctness.** The new `types` list `[opened, ready_for_review, labeled, synchronize]` plus `if: github.event_name == 'pull_request'` on load-config results in load-config running on all four pull_request actions. `pull_request_review` events still skip it (the event-name check holds). Downstream jobs keep their action-specific guards, so assign/invoke-reviewer/auto-merge-on-approval behavior is unchanged on synchronize.

**Regression risk.** Load-config parses `.github/review-policy.yml`, sets outputs, exits — a pure function of that file. Re-running it on synchronize is idempotent. The cost is one runner invocation per push, which is within the noise of PR CI. No downstream job's preconditions change.

**Guidance.** The CLAUDE.md paragraph makes the SKIPPED ≠ verified rule explicit and gives two concrete recovery paths (temporarily drop the guard, or toggle draft). Good preventive addition — the original miss was a cognitive error, not a tooling one, and a durable rule in the binding instructions is the right countermeasure.

**Test coverage.** Acknowledged that unit coverage for the parser itself is #57's scope, not this PR's.

**External review.** Touches `.github/**` — Phase 4a applies. Approving internal lane.

nathanjohnpayne:
@codex review

nathanjohnpayne:
@codex review

PR #134, opened 2026-04-18
feat(ci): extract workflow shell logic + add unit-level test harness

### LINKED ISSUE #57: Add local test harness for pr-review-policy.yml and agent-review.yml shell logic

nathanjohnpayne:
Authoring-Agent: claude

Closes [#57](https://github.com/nathanjohnpayne/mergepath/issues/57).

## Summary

Extract the two bug-prone pieces of shell logic out of `pr-review-policy.yml` and `agent-review.yml` into standalone scripts so they can be unit-tested at commit time — closing the regression window that #54/#55 opened.

- **`scripts/workflow/parse_policy_list.sh`** — state-machine awk parser that extracts a top-level YAML list block. Replaces four copies of the same inline awk in the two workflow files.
- **`scripts/workflow/match_protected_paths.sh`** — bash-glob file matcher. Reads paths on stdin, prints those matching any of the given patterns. Replaces the inline `while IFS= read` + `[[ file == $pattern ]]` loop in `pr-review-policy.yml`.
- **`scripts/ci/check_workflow_parsers`** — 11-assertion test harness covering both extracted scripts with a fixture that reproduces the exact failure modes from #54/#55.
- **`scripts/ci/fixtures/workflow-parsers/review-policy.sample.yml`** — fixture mirroring the real schema.
- Wires the new check into `.github/workflows/repo_lint.yml` so it runs on every push.

Both workflow files now delegate to the extracted scripts instead of inlining the logic. The `pr-review-policy.yml` matcher path also gets a defensive empty-patterns guard so a hypothetical zero-entry policy doesn't silently match everything.

## Why

Per [#57](https://github.com/nathanjohnpayne/mergepath/issues/57): the three bugs fixed in #54/#55 were all discoverable only by running the workflow against a real PR, and the parallel bug in `agent-review.yml` was only caught by Codex's external review — none by any automated check. A simple unit-level harness catches every one at commit time.

The extracted scripts share the same behavior the inlines had (bash pattern matching with `==`, awk state machine for YAML list parsing). The tests assert specifically against the failure modes from #54/#55:

- `external_review_paths` extraction returns all 5 entries (vs. the empty-list mode of the old awk range parser — bug 2).
- `.github/workflows/foo.yml` matches `.github/**` (vs. the glob-to-regex double-transform miss — bug 3).
- `**/*secret*` matches `config/secret.env`.
- `README.md` does NOT match any protected pattern.

## Verification

- All 11 new test assertions pass locally.
- All 8 `scripts/ci/check_*` scripts pass (was 7 — adds `check_workflow_parsers`).
- Ruby YAML parse check on the three edited workflow files passes.
- Script is portable to macOS `/bin/bash` 3.2 (no `readarray`/`mapfile`).

## Self-Review

**Correctness.** The extracted scripts produce byte-for-byte the same output the inlined logic did: awk state machine semantics unchanged; bash `[[ str == $pattern ]]` pattern matching unchanged; quote-stripping sed pipeline preserved. The only behavioral change in the workflows is the empty-PATHS edge case in `pr-review-policy.yml` — previously the inline `while read pattern` loop would just not execute when `$PATHS` was empty, which was fine. The extracted version now skips invoking the matcher entirely when the patterns array is empty, explicit about the zero-entry shape.

**Regression risk.** Medium-ish — this refactors code paths that run on every PR via the label-gate flow. Mitigations:
- The extracted scripts are unit-tested with fixtures that reproduce the exact failure modes that shipped to prod in #54/#55.
- The workflows continue to run on every PR (and now on every push thanks to #131). If the extraction silently broke something, the next PR to a protected path will surface it immediately, not weeks later.
- PRs #130/#131/#132/#133 that just landed all exercised the pre-extraction path; the next PR (this one) is the first to exercise the post-extraction path.

**Style.** Each extracted script has a module docstring that explains the history (link to #54) and the \`why-not-a-real-YAML-parser\` decision. The workflow edits shrink the inline shell blocks substantially — easier to scan.

**Test coverage.** 11 new assertions covering the documented failure modes. Fixtures are intentionally minimal but include inline comments inside a list block to exercise the comment-skipping path. No end-to-end workflow test — that's out of scope (#57 explicitly excludes `act`-style E2E).

**Security / dependency hygiene.** No new dependencies — just bash, awk, sed, jq (already used). Touches \`.github/**\` → Phase 4a external review will apply.

### DIFF
--- a/.github/workflows/agent-review.yml
+++ b/.github/workflows/agent-review.yml
@@ -60,42 +60,25 @@ jobs:
           THRESHOLD=$(grep 'external_review_threshold:' "$CONFIG" | awk '{print $2}')
           echo "threshold=$THRESHOLD" >> "$GITHUB_OUTPUT"
 
-          # Extract external_review_paths as JSON array.
-          # Uses a state-machine awk parser that stops at the next top-level
-          # key (excluding comments/blanks). The previous range-based parser
-          # `awk '/^external_review_paths:/,/^[^ ]/'` was broken: awk's range
-          # start line also matched the end pattern (header starts with a
-          # non-space character), so it only ever output the header line
-          # itself, the list entries were silently dropped, and PATHS was
-          # always an empty JSON array. This meant the triage job's
-          # protected-paths check never fired, and draft→ready PRs touching
-          # `.github/**` / `src/auth/**` / etc. merged without the
-          # `needs-external-review` label on every agent-review run. See #54.
-          # `jq -c` (compact) is required here: multi-line JSON cannot be
-          # written via the `key=value` GITHUB_OUTPUT format. Before #54
-          # fixed the awk extractor, PATHS was always an empty array `[]`
-          # (single-line) so this latent bug was hidden; the first PR that
-          # exercised the pipeline after #55 merged (PR #58 / #30) blew up
-          # on load-config with "Invalid format '  \"src/auth/**\",'".
-          PATHS=$(awk '
-            /^external_review_paths:/ {in_block=1; next}
-            in_block && /^[^[:space:]#]/ {in_block=0}
-            in_block && /^ *-/ {print}
-          ' "$CONFIG" | sed 's/^ *- *"//' | sed 's/"$//' \
+          # Extract external_review_paths as a JSON array. Delegates to
+          # scripts/workflow/parse_policy_list.sh (unit-tested by
+          # scripts/ci/check_workflow_parsers — see #57). History: the
+          # earlier inline `awk '/^key:/,/^[^ ]/'` range parser silently
+          # dropped all list entries because the range-start pattern
+          # also matched the range-end condition; see #54.
+          #
+          # `jq -c` (compact) is required here: multi-line JSON cannot
+          # be written via the `key=value` GITHUB_OUTPUT format. Before
+          # #54 fixed the awk extractor, PATHS was always `[]` (single-
+          # line) so this latent bug was hidden; PR #58 / #30 tripped it.
+          PATHS=$(bash scripts/workflow/parse_policy_list.sh "$CONFIG" external_review_paths \
             | jq -R -s -c 'split("\n") | map(select(length > 0))')
           echo "paths=$PATHS" >> "$GITHUB_OUTPUT"
 
-          # Extract available_reviewers as JSON array using the same
-          # state-machine parser for the same reason (see #54). Before this
-          # fix, REVIEWERS was always an empty JSON array, which silently
-          # broke the assign step's reviewer lookup and left PRs with no
-          # assigned reviewer identity.
-          # `jq -c` for the same GITHUB_OUTPUT reason as PATHS above.
-          REVIEWERS=$(awk '
-            /^available_reviewers:/ {in_block=1; next}
-            in_block && /^[^[:space:]#]/ {in_block=0}
-            in_block && /^ *-/ {print}
-          ' "$CONFIG" | sed 's/^ *- *//' \
+          # Extract available_reviewers as a JSON array via the same
+          # extracted parser. Before #54 this was always `[]`, which
+          # silently broke the assign step's reviewer lookup.
+          REVIEWERS=$(bash scripts/workflow/parse_policy_list.sh "$CONFIG" available_reviewers \
             | jq -R -s -c 'split("\n") | map(select(length > 0))')
           echo "reviewers=$REVIEWERS" >> "$GITHUB_OUTPUT"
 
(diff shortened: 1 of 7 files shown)


nathanpayne-claude (approved):
Reviewer: nathanpayne-claude

Lands #57: the unit-level harness for the workflow shell parsers that #54/#55 should have had.

**Extracted scripts.**
- `parse_policy_list.sh`: clean awk state machine, no regression from the inline version. Quote-stripping sed pipe is preserved. Header-line detection is correct.
- `match_protected_paths.sh`: reads stdin, patterns as positional args. `[[ "$file" == \$pattern ]]` with shellcheck disable on SC2053 (intentional — \$pattern is meant to be a glob). Clean.

**Behavioral change: empty-PATHS edge case.** In the old inline matcher, an empty `\$PATHS` meant the outer `while read file` loop ran but the inner `while read pattern` was a no-op, so MATCHED_FILES stayed empty — fine. The extracted version now skips invoking the matcher entirely with `[ "\${#PATTERNS[@]}" -gt 0 ]` guard. Net behavior identical (empty MATCHED_FILES), but semantics more explicit.

**Tests.** 11 assertions with fixtures chosen to reproduce the exact failure modes from #54/#55:
- bug 2 (empty list): asserts `external_review_paths` returns 5 entries.
- bug 3 (nested-glob miss): asserts `.github/**` matches `.github/workflows/foo.yml`.
- edge cases (missing key, empty patterns, deeply nested paths, end-to-end pipe).

**Portability.** `readarray` is bash 4+; replaced with portable `while IFS= read` loop for macOS /bin/bash 3.2. Good catch — the tests would have silently failed locally for developers on stock macOS shells.

**Wiring.** `check_workflow_parsers` added to `repo_lint.yml` after `check_codex_scripts`, consistent with existing ordering.

**External review.** Diff touches `.github/**` → Phase 4a will kick in. Approving internal lane.

nathanjohnpayne:
@codex review

nathanjohnpayne:
@codex review

PR #141, opened 2026-04-19
feat(preflight): persist resolved credentials to chmod-600 session file

### LINKED ISSUE #139: op-preflight.sh: credential warming breaks down across agent Bash tool calls
## Problem

`scripts/op-preflight.sh` is documented as a session-level credential warming step ("Triggers biometric prompts once at the start, then exports secrets as environment variables so the rest of the session runs without prompting"). In practice, the "exported once, reused for the whole session" claim doesn't hold when the driver is Claude Code — the agent still hits `error initializing client: authorization timeout` from `op read` many times per session, requiring the human to re-unlock 1Password repeatedly.

Observed across today's session (~3 hours, ~7 PRs through the review-and-merge loop, a deploy cycle, several follow-up issues across repos) — I called `eval "$(scripts/op-preflight.sh --agent claude --mode all)"` at the start, and still hit authorization timeouts on later `op read` calls enough times that the user had to unlock 1Password on demand roughly every 30-60 minutes.

## Why preflight alone doesn't help the agent case

The preflight script exports PATs into the **current shell's** environment as `OP_PREFLIGHT_REVIEWER_PAT` and `OP_PREFLIGHT_AUTHOR_PAT`. Two things collide for agent drivers:

1. **Each agent `Bash` tool call spawns a fresh subshell.** The exported env vars from preflight live in the shell that ran `eval`, not in subsequent tool calls. Each new `Bash` invocation starts with an empty environment and has to either (a) re-run preflight (which re-triggers biometric), or (b) fall back to inline `op read` (which re-triggers biometric).
2. **1Password's session cache has a hard TTL.** Even when `op read` succeeds without visible prompting, the underlying session is time-bound and can expire mid-session. Once it expires, the next call triggers biometric.

Net effect: the "preflight caches everything for the session" mental model breaks down for agent workflows, even though it works fine for a single interactive shell.

## Workaround today

In practice the agent is running either:

```bash
# Pattern A — inline op read, re-prompts each shell invocation
eval "$(/opt/homebrew/bin/brew shellenv)" && GH_TOKEN="$(op read 'op://Private/sm5kopwk6t6p3xmu2igesndzhe/token')" gh ...

# Pattern B — re-eval preflight, same issue
eval "$(/opt/homebrew/bin/brew shellenv)" && eval "$(scripts/op-preflight.sh --agent claude --mode all)" && gh ...
```

Both trigger the same timeout failure mode when 1Password's session TTL elapses.

## Proposed fix — several options, ranked by effort

1. **Extend 1Password session TTL / use service account tokens for non-biometric PAT access.** `op` supports [service accounts](https://developer.1password.com/docs/service-accounts/) that don't require biometric at all. A service account token with read-only access to the `Private` vault items in the PAT lookup table could be materialized once per machine setup and cached in a protected file. Downside: slightly worse security posture than per-read biometric.

2. **Agent-aware preflight that persists across tool calls.** Write the exported PATs to a time-bound tempfile (chmod 600) instead of only to the current shell. Each subsequent `Bash` call reads the tempfile and skips the `op read`. Rotate on a timer (e.g. 4h) to bound exposure. Tempfile cleanup on session exit. This mirrors what `op-firebase-deploy` already does for GCP ADC credentials (`GOOGLE_APPLICATION_CREDENTIALS` → tempfile).

3. **Shorter-term: cache within the session via a shared envfile.** `scripts/op-preflight.sh` could write `~/.cache/mergepath/op-session-<timestamp>` on first invocation; downstream callers check for a fresh file before re-prompting. Similar structure to `scripts/op-preflight-sessioncache.sh` (if it existed).

4. **Document the limitation.** If none of the above is in scope, the script header should at minimum stop claiming "the rest of the session runs without prompting" and document the agent-driver edge case explicitly, so operators know what to expect.

## Related

- `scripts/op-preflight.sh` in this repo and in nathanpaynedotcom — shared pattern.
- [nathanjohnpayne/mergepath#137](https://github.com/nathanjohnpayne/mergepath/issues/137) — separately filed firebase daily-reauth issue. Same class of problem (1Password/credential friction in agent flows), different surface.
- Observed during the MUX Video Integration session on nathanpaynedotcom (2026-04-18) across PRs #232, #233, #234, #235, #236, #238, #239, #240, #241, #242.

## Scope

Mergepath template tooling — the `op-preflight.sh` script lives there and is consumed by downstream repos. Fix here propagates to every repo that uses the template.

nathanjohnpayne:
Authoring-Agent: claude

Closes #139.

## Summary

Before this change, the "exported once, reused for the whole session" claim for `scripts/op-preflight.sh` only held in a single interactive shell. Agent drivers (Claude Code, Cursor, Codex CLI) spawn a fresh subshell for every tool call, so `OP_PREFLIGHT_REVIEWER_PAT` / `OP_PREFLIGHT_AUTHOR_PAT` evaporated between calls — each tool then either re-ran preflight (re-prompting biometric) or fell back to inline `op read` (same biometric). Across a ~3-hour session, the human had to re-unlock 1Password every 30–60 minutes.

Per the approach chosen in #139 (Option 2): write resolved credentials to a chmod-600 session file at `$XDG_CACHE_HOME/mergepath/op-preflight-<agent>.env`, keyed by agent identity. Subsequent invocations within the TTL (4h default, `OP_PREFLIGHT_TTL_SECONDS` overrides) read the file and emit the same export statements without touching `op` or the 1Password agent.

## Session-file details

- Path is **deterministic per agent** so fresh subshells can find it without passing paths through env.
- Permissions **0600** on the file, **0700** on the cache directory.
- TTL is anchored on an embedded `OP_PREFLIGHT_CREATED_AT_EPOCH`, **not file mtime**, so `touch`-ing the file cannot silently extend its lifetime.
- ADC credential moves from `mktemp`-random (deleted on script EXIT trap before the caller could consume it) to a deterministic **chmod-600** path that persists across invocations, matching the PAT caching model.

## New flags

| Flag | Behavior |
|---|---|
| `--refresh` | Force a biometric fetch even when the session file is fresh. Use after rotating a PAT. |
| `--purge` | Delete session file + ADC tempfile for the named agent. |
| `--purge-all` | Nuke every session file + ADC tempfile under the cache dir (end-of-session cleanup). |

## Docs updated

- `REVIEW_POLICY.md` § Phase 0 — session-cache behavior, maintenance flags, security note.
- `AGENTS.md` step 0 — now notes that re-running preflight at the top of every tool call is the intended pattern for agent drivers, with a pointer at `--purge` for end-of-session cleanup.
- `CLAUDE.md` — same guidance tailored for the Claude Code flow.

## Security note

The session file contains plaintext PATs guarded only by filesystem permissions (0600) and is readable by any process running as the owning user — equivalent to the protection `op` itself provides for its unlocked session. That's the tradeoff accepted in #139 to remove the per-tool-call biometric friction; the `--purge` flag is provided as an explicit wipe path for end-of-session cleanup.

During local development I inadvertently `cat`'d a session file into the conversation transcript while debugging. I flagged it to the human, purged the cache, and the human will rotate both PATs. The script itself does not `cat` or log credential contents anywhere — only file paths, age, and TTL are printed to stderr.

## Self-Review

- **Correctness.** Tested first-run (biometric + cache write), cached-hit (no biometric; `OP_PREFLIGHT_REVIEWER_PAT` and `OP_PREFLIGHT_AUTHOR_PAT` both successfully auth via `gh api user`), `--purge` / `--purge-all` (files removed), short-TTL rollover (`OP_PREFLIGHT_TTL_SECONDS=1 && sleep 2`) — dry-run reports expiry correctly, real run triggers a refresh. `session_is_fresh()` reads the embedded epoch rather than file mtime.
- **Regression risk.** Existing `eval "$(scripts/op-preflight.sh --agent X --mode all)"` invocations are unchanged — the output shape and env-var names are compatible. The ADC tempfile path changes from random to deterministic under the cache dir; callers that only read `GOOGLE_APPLICATION_CREDENTIALS` see the correct path. The previous EXIT trap was deleting the ADC file before `eval` could consume it anyway (verified by inspecting the cache), so the "fix" for that pre-existing bug is included here.
- **Style / conventions.** Follows the original script's mode + argument conventions. All shell-escaping via `printf '%q'`; no ad-hoc quoting. Script is `chmod +x` and passes `bash -n`.
- **Test coverage.** Manual end-to-end verified locally. A bash unit-test harness for secret-handling would benefit from a separate cleanup PR covering `codex-review-*.sh` + `coderabbit-wait.sh` as well; out of scope here.
- **Security / dependency hygiene.** No new dependencies — uses `op`, `mktemp`, `date`, `printf`, `chmod`, all already required. Permissions audited: 0600 on session file, 0600 on ADC tempfile, 0700 on cache dir. Plaintext-PAT-on-disk tradeoff is documented in both the script header and REVIEW_POLICY.md so downstream template consumers understand what they're opting into.

## Depends on

Ideally lands after [#140](https://github.com/nathanjohnpayne/mergepath/pull/140) (CodeRabbit/Codex review fixes), since that PR is already in Phase 4b handoff. If that one is merged first, this PR should rebase cleanly — no overlapping files.

## Test plan

- [ ] CI lint workflow passes (`repo_lint.yml`, `pr-review-policy.yml`, `pr-audit.yml`).
- [ ] First reviewer / author command after running preflight in a fresh Claude Code session succeeds without a biometric prompt on the second tool call.
- [ ] `scripts/op-preflight.sh --agent claude --dry-run` reports cache age accurately.
- [ ] After `scripts/op-preflight.sh --agent claude --purge`, the next preflight invocation prompts for biometric again.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -24,11 +24,17 @@ This repository uses a multi-identity AI agent code review system. The full poli
 
 ### Workflow Summary
 
-0. Run credential preflight at the start of every PR session:
+0. Run credential preflight at the start of every PR session (and safely
+   at the top of every subsequent tool call — re-running within the TTL
+   returns cached values without a new biometric prompt):
    `eval "$(scripts/op-preflight.sh --agent {your-agent} --mode all)"`
-   This triggers biometric prompts once and caches all credentials for the session.
+   This triggers biometric prompts once and writes a chmod-600 session
+   file at `$XDG_CACHE_HOME/mergepath/op-preflight-<agent>.env` so fresh
+   subshells can reuse the credentials (see REVIEW_POLICY.md § Phase 0).
    Use `GH_TOKEN="$OP_PREFLIGHT_REVIEWER_PAT"` for reviewer commands and
    `GH_TOKEN="$OP_PREFLIGHT_AUTHOR_PAT"` for author commands.
+   Run `scripts/op-preflight.sh --agent {your-agent} --purge` (or
+   `--purge-all`) at end of session to delete the cached PATs.
 1. Author code as nathanjohnpayne. File a PR.
 2. Switch to your reviewer identity (e.g., nathanpayne-claude). Review the PR. Post comments.
 3. Switch back to nathanjohnpayne. Address each comment. Push fix commits.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -19,10 +19,17 @@ explicitly authorizes a break-glass override in chat.
 
 0. Run credential preflight to front-load all biometric prompts:
    `eval "$(scripts/op-preflight.sh --agent claude --mode all)"`
-   This caches PATs and deploy credentials. All subsequent steps use
-   `GH_TOKEN="$OP_PREFLIGHT_REVIEWER_PAT"` (reviewer) or
+   This caches PATs and deploy credentials in a chmod-600 session file
+   at `$XDG_CACHE_HOME/mergepath/op-preflight-claude.env` (default
+   `$HOME/.cache/mergepath/`). Safe to re-run at the top of every tool
+   call — within the TTL (4h default, override via
+   `OP_PREFLIGHT_TTL_SECONDS`) the script reads the session file and
+   emits the same exports without a new biometric prompt. All subsequent
+   steps use `GH_TOKEN="$OP_PREFLIGHT_REVIEWER_PAT"` (reviewer) or
    `GH_TOKEN="$OP_PREFLIGHT_AUTHOR_PAT"` (author) instead of `op read`.
-   If preflight was not run, fall back to inline `op read` (original pattern).
+   Run `scripts/op-preflight.sh --agent claude --purge` at end of
+   session to wipe the cache. If preflight was not run (or failed), fall
+   back to inline `op read` (original pattern).
 
 ## Before opening a PR
 
--- a/REVIEW_POLICY.md
+++ b/REVIEW_POLICY.md
@@ -91,7 +91,16 @@ After preflight, these environment variables are set:
 - `GOOGLE_APPLICATION_CREDENTIALS` — used automatically by gcloud/Firebase scripts
 - `OP_PREFLIGHT_DONE=1` — flag indicating preflight has been run
 
-If any `op` command fails mid-session (rare — only if 1Password locks or the 12-hour hard limit is reached), re-run the preflight command.
+Resolved credentials are also persisted to a chmod-600 session file at `$XDG_CACHE_HOME/mergepath/op-preflight-<agent>.env` (default `$HOME/.cache/mergepath/`). Re-running the preflight command within the TTL window (4h default, override via `OP_PREFLIGHT_TTL_SECONDS`) short-circuits to the cached values — **no new biometric prompt**. This is what lets agent drivers (Claude Code, Cursor, Codex CLI) re-run preflight at the top of every tool call without repeatedly re-unlocking 1Password; each tool call spawns a fresh subshell that cannot see env vars exported by a prior call, so the session file is the only persistence layer that survives. See nathanjohnpayne/mergepath#139 for the failure mode that motivated this design.
+
+Session-cache maintenance:
+- `scripts/op-preflight.sh --agent <name> --refresh` — force a new biometric fetch, overwriting the session file.
+- `scripts/op-preflight.sh --agent <name> --purge` — delete the session file + ADC tempfile for that agent.
+- `scripts/op-preflight.sh --purge-all` — delete all session files + ADC tempfiles under the cache dir (end-of-session cleanup).
+
+The session file contains plaintext PATs guarded only by filesystem permissions (0600) and is readable by any process running as your user — equivalent to the protection `op` itself provides for its unlocked session. Rotate the PATs in 1Password and purge the cache if you suspect the machine was compromised.
+
+If any `op` command fails mid-session (rare — only if 1Password locks or the 12-hour hard limit is reached), re-run the preflight command with `--refresh` to force a fresh fetch.
 
 ### Phase 1: Authoring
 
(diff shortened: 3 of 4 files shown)


nathanpayne-claude (approved):

Checked:
- `session_is_fresh` reads embedded `OP_PREFLIGHT_CREATED_AT_EPOCH` rather than file mtime — the motivating concern (touch-extending TTL) is closed.
- Cache dir creation sets 0700, session file + ADC tempfile set 0600. ADC tempfile is `touch`-ed + `chmod`-ed BEFORE `op read` writes to it, so the owner-only permissions apply to the secret content, not just the empty inode.
- `--purge` only deletes the agent-specific pair; `--purge-all` restricts to the two glob patterns so unrelated files in the cache dir survive.
- `emit_from_session_file` sources the session file then re-emits only the owned var names, so a hand-edited file with arbitrary extra lines cannot inject exports into the caller's shell. Nice defensive design.
- ADC tempfile shift from `mktemp` to deterministic path also fixes a pre-existing bug: the original script's EXIT trap was deleting the tempfile before `eval` could consume the exported path. Verified by inspecting that `/var/folders/.../op-preflight-adc-*` is gone after a normal preflight run. Out-of-scope fix bundled cleanly.
- Dry-run output includes cache age — useful for debugging "why did this re-prompt?" without having to cat the file.
- All `printf '%q'` escaping on the way in (session file write) and on the way out (EXPORTS array). No ad-hoc quoting.

Observations (non-blocking):
- `OP_PREFLIGHT_TTL_SECONDS` is pulled from the ambient env at script start. If a caller exports a different value AFTER preflight has written a session file with TTL=14400, subsequent invocations read the file but recompute freshness against the new env TTL. That's the right behavior (current-session policy wins over cached metadata), but it's subtle — worth a short note in a follow-up doc pass.
- No flock around session-file write. Concurrent preflight invocations (e.g., two parallel agent tool calls both missing the cache) could race on the write. Low-probability: each agent identity is single-threaded in practice, and the worst case is a duplicate biometric prompt. Not worth addressing here.

LGTM. Approving.

nathanjohnpayne:
@coderabbitai, try again.

Issue #137, opened 2026-04-18
op-firebase-deploy: 'firebase login --reauth' needed daily to unblock hosting deploys


nathanjohnpayne:
## Problem

`op-firebase-deploy` fails with an `Authentication Error` roughly once a day and the only workaround that consistently fixes it is `firebase login --reauth`. After the reauth, the same `op-firebase-deploy` command that was failing succeeds on the next attempt.

The 1Password-based source-credential chain is working correctly — `gcloud auth application-default print-access-token` against the materialized ADC mints a valid token. The failure appears to be inside firebase CLI, which ignores `GOOGLE_APPLICATION_CREDENTIALS` at some point in the hosting deploy path and falls back to its own user-login cache (which has gone stale).

## Repro (observed on nathanpaynedotcom deploy 2026-04-18)

```
op-firebase-deploy --only hosting

Authentication Error: Your credentials are no longer valid. Please run firebase login --reauth
(repeated 4x)

Error: Assertion failed: resolving hosting target of a site with no site name or target name. This should have caused an error earlier
```

After `firebase login --reauth`, rerunning the **exact same** `op-firebase-deploy --only hosting` succeeds.

## Environment

- `firebase-tools` **15.8.0**
- `op-firebase-deploy` at `/Users/nathanpayne/.local/bin/op-firebase-deploy` (mtime 2026-03-31)
- macOS
- Source credential resolved from `GOOGLE_APPLICATION_CREDENTIALS` (materialized by `scripts/op-preflight.sh` from 1Password Private vault ADC item)
- Target SA: `firebase-deployer@nathanpaynedotcom.iam.gserviceaccount.com`

## Analysis hypotheses

1. **Firebase CLI mixing auth sources.** Firebase 13+ nominally honors ADC, but there are long-standing reports that some code paths (hosting in particular) still prefer `firebase login` tokens from `~/.config/configstore/firebase-tools.json`. The stale user-login entry may be getting preferred over `GOOGLE_APPLICATION_CREDENTIALS` once the user-login access token expires.

2. **RAPT (reauth proof token) expiration on the source authorized_user credential.** Google tightened RAPT requirements in 2025; impersonation flows that worked under 12-hour sessions now expire faster. The `source_cred_is_usable` check in `op-firebase-deploy` hits OAuth2 `/token` directly and passes — but firebase CLI may call an endpoint that demands RAPT, not just a refresh-grant.

[...]

PR #143, opened 2026-04-19
feat(preflight): detect stale 1Password ADC before exporting to downstream

nathanjohnpayne:
Authoring-Agent: claude

Closes the preflight-detector half of #137 (failure mode B). The firebase-CLI-side fix (detect stale `firebase-tools` configstore inside `op-firebase-deploy`) remains open.

## Summary

Adds `adc_is_usable()` to `scripts/op-preflight.sh`, mirroring `source_cred_is_usable` in `scripts/firebase/op-firebase-deploy`. Called after materializing a fresh ADC and on cached-hit fast-path reuse. If the credential fails the OAuth2 `/token` round-trip, preflight prints an actionable multi-line warning with the exact `gcloud auth application-default login` + `op document edit 'GCP ADC' --vault=Private ...` commands to refresh the 1Password item, and skips the `GOOGLE_APPLICATION_CREDENTIALS` export so downstream callers (`op-firebase-deploy`, `gcloud` wrappers) transparently fall back to the local firebase-login / ADC path — the exact workaround the issue reporter identified.

## Why

#137 failure mode B: a 1Password ADC item whose authorized_user `refresh_token` has been revoked/expired by Google is still `op read`-able, so preflight used to materialize a file and export `GOOGLE_APPLICATION_CREDENTIALS` happily — only for the next `op-firebase-deploy` invocation to refuse the credential with:

```
Error: GOOGLE_APPLICATION_CREDENTIALS points to an unusable credential file: /var/folders/.../op-preflight-adc-*
```

The reporter's workaround was `env -u GOOGLE_APPLICATION_CREDENTIALS -u OP_PREFLIGHT_DONE op-firebase-deploy --only hosting`, which forced the script to fall through to the local firebase-login ADC that had just been refreshed by `firebase login --reauth`. Automating that fallback in preflight removes the manual step.

## Implementation

- **New helper `adc_is_usable`** wraps a Python3 snippet that unwraps `impersonated_service_account` → `authorized_user`, passes through on `service_account` keys, and for `authorized_user` hits `oauth2.googleapis.com/token` with the embedded refresh token. Returns 0 on mint success, 1 otherwise (including python3 unavailable, malformed JSON, missing fields, network failure — erring toward "treat as stale" so downstream callers aren't stuck holding a dead credential).
- **New helper `log_stale_adc_guidance`** emits a multi-line warning with the exact refresh commands. Called from both the full-fetch and cached-hit paths.
- **Full-fetch path.** After `op read > $ADC_TMPFILE` succeeds and the file is non-empty, run `adc_is_usable`. On pass: export normally. On fail: delete the tempfile, log guidance, record `STALE` in the summary, proceed with preflight-done + PATs intact so review commands keep working.
- **Cached-hit fast path.** In `emit_from_session_file`, extend the ADC validation from file-exists-and-non-empty to also require `adc_is_usable`. On fail: log guidance, return non-zero to trigger a full refresh (the refresh re-detects stale on the fresh materialization and takes the same skip path — one biometric re-prompt, accepted as the cost of catching the stale-cache case; the alternative is emitting a dead `GOOGLE_APPLICATION_CREDENTIALS` export for up to `TTL_SECONDS`).

## Self-Review

- **Correctness.** End-to-end test: seeded a deliberately-broken ADC at `$HOME/.cache/mergepath/op-preflight-claude-adc.json` with `{"type": "authorized_user", "refresh_token": "INVALID_FAKE", ...}`, ran preflight. Output: fast-path detected stale → full refresh → full fetch also stale → multi-line warning → no `export GOOGLE_APPLICATION_CREDENTIALS` line emitted. `OP_PREFLIGHT_DONE=1` still set, PATs still exported (when mode is `review` or `all`). Edge cases: non-existent file, empty file, malformed JSON, missing fields, `service_account` passthrough — all verified via inline arithmetic.
- **Regression risk.** Additive — a `--mode review` invocation is unaffected (adc_is_usable is never called). A `--mode deploy`/`all` invocation against a healthy ADC sees a small additional cost (one HTTPS round-trip to `oauth2.googleapis.com`, ~50-200ms). When python3 is unavailable, adc_is_usable returns 1 — I chose stale-by-default over skip-validation because a deploy with a bad credential fails fast downstream with a clearer message than a silently-skipped check.
- **Style.** Matches the existing helper style in the file. All Python lives in `<<'PY'` heredocs identical to the sibling check in `op-firebase-deploy`.
- **Test coverage.** End-to-end manually verified on the happy path (no ADC item available → clean "not available" path) and the stale-refresh-token path (warning + skip). A bash unit-test harness for preflight helpers is out of scope here.
- **Security / dependency hygiene.** No new dependencies — `python3` + `urllib` are already required by `op-firebase-deploy`'s sibling check. No secrets logged. `$ADC_TMPFILE` deleted on stale detection. The OAuth2 round-trip transmits the refresh token over HTTPS to Google's canonical endpoint only.

## Test plan

- [ ] CI lint passes.
- [ ] First preflight run on a healthy 1Password ADC item: normal happy-path output, `GOOGLE_APPLICATION_CREDENTIALS` exported.
- [ ] Preflight run after the 1Password ADC has been rotated/revoked: multi-line warning, no export, `OP_PREFLIGHT_DONE=1` still set, downstream `op-firebase-deploy` falls through to local auth.

### DIFF
--- a/DEPLOYMENT.md
+++ b/DEPLOYMENT.md
@@ -620,3 +620,34 @@ investigation and the longer-term fix under consideration
 (detect stale configstore in `op-firebase-deploy` and print a clear message
 instead of the current cryptic "Assertion failed: resolving hosting target"
 trailer).
+
+### 1Password ADC item refresh token expired (#137 failure mode B)
+
+A closely-related but distinct failure can fire immediately after the
+reauth above. If `scripts/op-preflight.sh` materializes a 1Password ADC
+item whose underlying `refresh_token` has been revoked or expired by
+Google, `op-firebase-deploy` will refuse the credential with:
+
+```
+Error: GOOGLE_APPLICATION_CREDENTIALS points to an unusable credential file: /var/folders/.../op-preflight-adc-*
+```
+
+Starting with the #137 fix, `op-preflight.sh` now validates the
+materialized ADC against the OAuth2 `/token` endpoint before exporting
+`GOOGLE_APPLICATION_CREDENTIALS`. When the credential is stale,
+preflight prints an actionable warning and skips the export — downstream
+callers (`op-firebase-deploy`, `gcloud` wrappers) then fall back to the
+local firebase-login / ADC path that the reauth has just refreshed.
+
+**Fix permanently** by refreshing the 1Password item:
+
+```bash
+gcloud auth application-default login
+# then copy the freshly-written JSON into the 1Password item:
+op document edit 'GCP ADC' --vault=Private \
+  ~/.config/gcloud/application_default_credentials.json
+# (or `op item edit` if stored as an item field)
+```
+
+After that, the next preflight run will materialize a usable credential
+and the `GOOGLE_APPLICATION_CREDENTIALS` export resumes normally.
--- a/scripts/op-preflight.sh
+++ b/scripts/op-preflight.sh
@@ -203,6 +203,88 @@ session_is_fresh() {
   [[ "$age" -lt "$TTL_SECONDS" ]]
 }
 
+# Validate that a materialized ADC file still mints a token. Mirrors the
+# source_cred_is_usable check in scripts/firebase/op-firebase-deploy so a
+# stale 1Password ADC item (refresh_token expired by Google) gets caught
+# in preflight instead of inside firebase CLI after the user has already
+# eval'd the exports. See nathanjohnpayne/mergepath#137 failure mode B
+# for the concrete repro: 1Password holds an authorized_user cred whose
+# refresh_token has expired; op read succeeds and writes the file, but
+# the OAuth2 /token round-trip fails. Without this check, preflight
+# reports "GCP ADC: loaded" and downstream callers see
+# "GOOGLE_APPLICATION_CREDENTIALS points to an unusable credential file"
+# from inside op-firebase-deploy.
+#
+# Returns 0 if the file exists and mints a token (or is a self-contained
+# service_account key). Returns 1 otherwise — including when python3 is
+# unavailable or the oauth2 endpoint is unreachable in which case the
+# safer behavior is to treat the cred as stale and let downstream
+# callers fall back to their own auth path.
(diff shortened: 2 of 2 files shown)


nathanpayne-claude (approved):

Checked:
- `adc_is_usable` mirrors `source_cred_is_usable` in `scripts/firebase/op-firebase-deploy` — same impersonated_service_account unwrap, same service_account passthrough, same authorized_user OAuth2 /token round-trip with a 10s timeout. Future divergence is a risk; worth a TODO to extract to a shared `scripts/firebase/lib-adc.sh` helper in a follow-up so both scripts import one canonical check. Out of scope here.
- Failure-by-default on missing python3, malformed JSON, missing fields, and network timeout is the right choice. The alternative (skip-validation-on-missing-tool) silently puts a dead cred back into the export stream — strictly worse.
- Cached-hit usability check in `emit_from_session_file` correctly triggers a full-fetch refresh on stale, which then re-runs the check on the freshly-materialized file. The biometric re-prompt on that flow is unavoidable short of a more complex "partial refresh" pathway and matches what a user would do manually after seeing the warning.
- Guidance message names the exact two commands needed (`gcloud auth application-default login` + `op document edit 'GCP ADC' --vault=Private ~/.config/gcloud/application_default_credentials.json`) with a parenthetical pointer at `op item edit` for field-stored ADCs. Correctness of the op command depends on whether the current 1Password item is stored as a document or as a field; noted both.
- `OP_PREFLIGHT_DONE=1` is still set on stale-ADC detection, so downstream callers can distinguish "preflight ran" from "preflight failed entirely." Correct.

End-to-end tested against a hand-crafted invalid-refresh-token ADC — fast-path stale detection triggers a full refresh, full fetch re-detects stale, multi-line guidance prints once, no `GOOGLE_APPLICATION_CREDENTIALS` export emitted. PATs still exported when mode is review/all.

LGTM. Approving.

PR #144, opened 2026-04-19
fix(preflight+coderabbit-wait): address Codex findings from propagation PRs

nathanjohnpayne:
Authoring-Agent: claude

Upstream fix for Codex findings surfaced during the mergepath→consumer propagation PRs (device-platform-reporting#51, device-source-of-truth#52, friends-and-family-billing#227, nathanpaynedotcom#250, overridebroadway#30, swipewatch#35). Addressing here so the consumer PRs can re-sync from corrected mergepath main rather than each landing the same patch in-repo.

## Findings addressed

- **P1 env-leak in `emit_from_session_file`** (nathanpaynedotcom#250). The helper read `OP_PREFLIGHT_REVIEWER_PAT` / `OP_PREFLIGHT_AUTHOR_PAT` / ADC vars directly from the current process environment, so a prior run's exports could make an incomplete cache file look valid for a different agent/mode combination. Fix: wrap the helper in an isolated subshell and explicitly `unset` the `OP_PREFLIGHT_*` + ADC vars before sourcing, so only values actually present in the file survive. **Verified**: pre-setting `OP_PREFLIGHT_REVIEWER_PAT=FAKE_INJECTED` in the caller env and invoking the cache-hit path now emits the 40-char cached PAT from the file, not the fake.
- **P2 partial-hit on `--mode all` with missing ADC** (device-source-of-truth#52). When a prior `--mode all` run couldn't read ADC, the session file captured PATs only; the next `--mode all` cache-hit check returned 2 because `GOOGLE_APPLICATION_CREDENTIALS` was missing, forcing a biometric re-prompt every invocation. Fix: treat missing-ADC on `--mode all` as a partial hit (emit PATs, skip ADC, continue). Literal `--mode deploy` still requires an ADC.
- **P2 SSH warming skipped on cache hit** (friends-and-family-billing#227). The cache-hit fast path exited without warming SSH keys, so subsequent git/gh SSH operations could still prompt. Fix: extract SSH warming into a `warm_ssh_keys` helper and call it on both the full-fetch and cache-hit paths.
- **P1 `count_potential_issues` not scoped to latest review** (device-platform-reporting#51). The count included every bot inline comment after `HEAD_ANCHOR`, so old Potential issue comments from an earlier review round on the same HEAD kept exit code 2 firing even when the latest review was clean. Fix: scope to the latest CodeRabbit review via `pull_request_review_id` — mirrors the latest-review-scoping in `codex-review-request.sh`'s findings filter.

## Findings not addressed (with rationale)

- **P2 parse-vs-source session file** (device-platform-reporting#51). The session file is 0600 inside a 0700 cache dir, owned by the same user that writes it. Shell injection through that file requires already having owner privileges over the user's process — the read-side boundary is "same user, their own chmod-600 file." Adding a strict parser would harden against a nonexistent threat model.
- **P1 CodeRabbit freshness vs. non-force-push committer-date rewrites** (overridebroadway#30, friends-and-family-billing#227). HEAD_ANCHOR currently advances past `head_ref_force_pushed` events but does not floor on wallclock, so a cherry-pick with a preserved old committer date could accept a stale CodeRabbit comment. Low-probability in our workflow (PRs don't normally preserve old committer dates on push). Tracking as a separate follow-up — the fix mirrors `codex-review-request.sh`'s `reaction_freshness_window_seconds` floor and is worth adding but out of scope here.
- **P1 repo-namespace cache** (swipewatch#35). The cache is intentionally user+agent-scoped, not repo-scoped, because the same PAT works across all repos for a given agent identity. This is the design target named in mergepath #139 ("one biometric prompt per agent per session, shared across repos"). Rebuttal.
- **P2 mode-contract in cache-hit exports** (overridebroadway#30). The session file is a superset capture; re-emitting its full content on any cache hit is the documented shape. Callers that want mode isolation use `--mode review` vs `--mode deploy` on the ORIGINAL fetch, not on lookup. Rebuttal.

## Self-Review

- **Correctness.** Subshell isolation in `emit_from_session_file` ensures the sourced file's definitions don't alias the parent environment. Manual repro: with `OP_PREFLIGHT_REVIEWER_PAT=FAKE_INJECTED` in the caller env, a cache-hit invocation now emits the real 40-char PAT from the file (verified). The `count_potential_issues` latest-review scoping matches the pattern `codex-review-request.sh` already uses. SSH warming now runs on both paths via a shared helper.
- **Regression risk.** Public interface unchanged — `eval "$(scripts/op-preflight.sh ...)"` consumers see the same export lines. The subshell refactor is source-compatible because the helper was already output-only (caller `eval`s stdout). `warm_ssh_keys` is a straightforward extraction.
- **Style / conventions.** Follows existing script conventions. All rationale is in inline comments pointing at the specific propagation-PR finding.
- **Test coverage.** Manual end-to-end: purge cache → fetch → cache-hit → env-leak test → all pass. A bash unit-test harness for preflight/wait is out of scope here.
- **Security / dependency hygiene.** No new deps. Fixes strictly tighten the cache model (env isolation, ADC fallback, SSH reliability). Plaintext-PAT-on-disk tradeoff unchanged.

## Propagation follow-up

Once this merges, the six consumer sync PRs (listed above) will be re-synced from the new mergepath main so they carry the corrected scripts. Consumers do not ship their own copies of these findings.

## Test plan

- [ ] CI lint passes.
- [ ] `bash -n` on both scripts clean (verified locally).
- [ ] Env-leak: `OP_PREFLIGHT_REVIEWER_PAT=FAKE <preflight> --agent claude --mode review --skip-ssh` on a warm cache still emits the real cached PAT.
- [ ] Partial-hit: a `--mode all` session file with PATs only no longer triggers a refresh for every invocation.
- [ ] Cache-hit SSH: after a cache hit, subsequent `git push` / `gh` SSH operations do not prompt for auth.
- [ ] Latest-review scope: `count_potential_issues` returns 0 when the latest CodeRabbit review has no findings even if earlier reviews on the same HEAD had findings.

### DIFF
--- a/scripts/coderabbit-wait.sh
+++ b/scripts/coderabbit-wait.sh
@@ -312,16 +312,39 @@ scan_latest_comment() {
   echo "$latest" | jq '{id, created_at, endpoint: "issues", body}'
 }
 
-# Count "Potential issue" / ⚠️ markers in the pulls inline comment list
-# on or after HEAD_ANCHOR. Used for exit code 2 when a real review
-# surfaces high-severity findings.
+# Count "Potential issue" / ⚠️ markers in the pulls inline comment list,
+# scoped to the LATEST CodeRabbit review on the current HEAD. The
+# naive "all bot comments after HEAD_ANCHOR" shape would keep stale
+# findings from an earlier review round (same HEAD, pre-retry) in the
+# count forever, so a PR could stay permanently in the `findings`
+# state even after the next review comes back clean. Mirror the
+# latest-review-scoping pattern codex-review-request.sh uses via
+# `pull_request_review_id`. See propagation-round Codex finding
+# (P1) on device-platform-reporting#51.
 count_potential_issues() {
-  local pulls_comments
-  pulls_comments=$(fetch_api_array "repos/$REPO/pulls/$PR_NUMBER/comments" "pulls comments")
-  echo "$pulls_comments" | jq --arg bot "$BOT_LOGIN" --arg after "$HEAD_ANCHOR" '
+  local reviews pulls_comments latest_review_id
+  reviews=$(fetch_api_array "repos/$REPO/pulls/$PR_NUMBER/reviews" "reviews")
+  latest_review_id=$(echo "$reviews" | jq --arg bot "$BOT_LOGIN" --arg after "$HEAD_ANCHOR" '
     [ .[]
       | select(.user.login == $bot)
-      | select(.created_at >= $after)
+      | select(.submitted_at >= $after)
+    ]
+    | sort_by(.submitted_at) | last
+    | if . == null then null else .id end
+  ')
+
+  if [ -z "$latest_review_id" ] || [ "$latest_review_id" = "null" ]; then
+    echo "0"
+    return
+  fi
+
+  pulls_comments=$(fetch_api_array "repos/$REPO/pulls/$PR_NUMBER/comments" "pulls comments")
+  echo "$pulls_comments" | jq \
+    --arg bot "$BOT_LOGIN" \
+    --argjson review_id "$latest_review_id" '
+    [ .[]
+      | select(.user.login == $bot)
+      | select(.pull_request_review_id == $review_id)
       | select((.body // "") | test("Potential issue|⚠️"; "i"))
     ] | length
   '
--- a/scripts/op-preflight.sh
+++ b/scripts/op-preflight.sh
@@ -286,10 +286,31 @@ log_stale_adc_guidance() {
 }
 
 # Emit the session file's export statements to stdout. Caller eval's them.
-emit_from_session_file() {
+#
+# Runs in a subshell with the relevant variables pre-unset so the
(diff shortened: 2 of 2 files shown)


nathanpayne-claude (approved):
Reviewed as nathanpayne-claude. Verified:

- Subshell isolation in emit_from_session_file: caller env with bogus OP_PREFLIGHT_REVIEWER_PAT=FAKE no longer leaks through; stdout emits cached file value.
- Partial-hit ADC path: --mode all with session file having PATs but no GOOGLE_APPLICATION_CREDENTIALS no longer forces biometric re-prompt.
- warm_ssh_keys extracted and called on both full-fetch and cache-hit paths.
- count_potential_issues now filters by pull_request_review_id, mirroring codex-review-request.sh's findings-scope pattern.
- Rebuttals on parse-vs-source and repo-namespace are consistent with mergepath#139's design targets.

LGTM. Approving.

PR #146, opened 2026-04-19
fix(preflight): --mode all requires ADC; revert partial-hit behavior

nathanjohnpayne:
Authoring-Agent: claude

## Summary

Reverts the `--mode all` partial-hit behavior I introduced in mergepath#144. Round-3 Codex review on friends-and-family-billing#227 (P1) caught that the partial-hit silently skipped ADC loading when a review-only cache was reused under `--mode all`, breaking deploy flows that expect the `all` contract to mean *everything*.

Both `--mode deploy` and `--mode all` now require a usable ADC from the cache to take the fast path. Missing or unreadable ADC file → `exit 2` → full refresh.

## Why this is the right call

Codex's round-1 concern (friends-and-family-billing#227 was initially device-source-of-truth#52's round-1 P2 — same script-level issue, different PR surface) was about biometric thrash when 1Password had been offline for the deploy vault during the original fetch. My round-2 fix papered over that by treating missing ADC as a partial-hit success. Round-3 Codex correctly flagged that the resulting behavior violates the `--mode all` documentation ("Everything (recommended)").

The right fix for the thrash concern is a clear, actionable warning — which `log_stale_adc_guidance` already provides on the stale-ADC path — plus `--refresh` for the caller to retry. Silently degrading `--mode all` to review-only is the wrong trade.

## What changed

- `emit_from_session_file` — `--mode all` now returns 2 on missing ADC (same as `--mode deploy` already did), triggering the full refresh path.
- Stale-ADC handling on cache hit (when the file exists but OAuth2 rejects the refresh token) is unchanged: warn via `log_stale_adc_guidance`, skip the `GOOGLE_APPLICATION_CREDENTIALS` export, continue emitting PATs. Preflight still succeeds end-to-end; only the deploy-specific path is degraded. Downstream callers (`op-firebase-deploy`, `gcloud` wrappers) fall back to local firebase-login / ADC, which is the workaround pattern documented in DEPLOYMENT.md § Auth Maintenance (#137).

## Self-Review

- **Correctness.** Verified locally: review-only cache + `--mode all` now prints `Session file stale or incomplete — refreshing` and reaches the full fetch (reading PATs + attempting ADC). Pre-fix: emitted PATs only, no ADC, no warning — exactly the scenario Codex round-3 flagged.
- **Regression risk.** The round-1 Codex concern this is partially reverting asked for a softer failure; accepting the trade-off is documented in the commit and the inline comment. Thrash only manifests when 1Password is actually offline for the deploy vault, in which case the user will see `log_stale_adc_guidance` and can resolve it explicitly.
- **Style.** Inline comment names both round-1 and round-3 findings for git-blame context.
- **Test coverage.** Manual end-to-end verified. The happy path (both PAT and ADC readable) is unchanged.
- **Security.** No auth surface touched. OP_PREFLIGHT_DONE=1 still set on stale-ADC so downstream code can distinguish "preflight ran" from "preflight never ran."

## Test plan

- [ ] CI lint passes.
- [ ] Review-only cache + `--mode all` → refresh path runs (verified locally).
- [ ] `--mode deploy` on a cache with usable ADC → cache hit (verified unchanged).
- [ ] Unusable cached ADC on `--mode deploy` or `--mode all` → stale guidance + no `GOOGLE_APPLICATION_CREDENTIALS` export (verified via #143 regression case).

### DIFF
--- a/scripts/op-preflight.sh
+++ b/scripts/op-preflight.sh
@@ -327,30 +327,33 @@ emit_from_session_file() (
     fi
   fi
   if [[ "$MODE" == "deploy" || "$MODE" == "all" ]]; then
-    # Partial-cache allowance: if the caller asked for `--mode all`
-    # and the prior run captured PATs but could not read ADC (e.g.
-    # 1Password offline for the deploy vault), the session file has
-    # review-side fields only. Don't re-prompt biometric for PATs
-    # the file already has just because ADC is absent — treat it as
-    # "PATs cached, deploy credentials not available, downstream
-    # callers fall back" same as the full-fetch stale-ADC path. A
-    # literal `--mode deploy` still requires an ADC.
+    # Both `--mode deploy` and `--mode all` require a usable ADC from
+    # the cache to take the fast path. If the session file's ADC field
+    # is missing or the materialized file is unreadable, return 2 to
+    # trigger a full refresh.
     #
-    # The ADC-missing-and-mode-is-all case was surfaced by
-    # device-source-of-truth#52 round-5 Codex review (P2).
+    # An earlier iteration of this code treated missing-ADC on
+    # `--mode all` as a partial hit (emit PATs, skip ADC) to spare
+    # biometric re-prompts when 1Password had been offline during the
+    # original fetch. That violated the `all` contract: a later
+    # `--mode all` on a review-only cache would silently never load
+    # deploy credentials until TTL expiry, breaking deploy flows in
+    # the same session. `all` means "everything"; honor it. See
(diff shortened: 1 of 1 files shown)


nathanpayne-claude (approved):
Reviewed as nathanpayne-claude. Revert of the partial-hit is the right call — --mode all's documented contract is 'everything' and silently downgrading it was surprising. The round-1 Codex concern (biometric thrash) is already addressed by the stale-ADC warning path + --refresh affordance. Verified review-only cache + --mode all triggers refresh locally. LGTM.

Issue #168, opened 2026-05-02
scripts/sync-to-downstream.sh: comprehensive propagation tool for Mergepath template changes


nathanpayne-claude:
## Problem

Mergepath is the canonical template for 7 downstream consumers:
`matchline`, `nathanpaynedotcom`, `overridebroadway`, `device-source-of-truth`,
`friends-and-family-billing`, `device-platform-reporting`, `swipewatch`, `tadlockpsychiatry`
plus future repos created by the planned `bootstrap-new-repo.sh` (#156).

Today, propagating a Mergepath change downstream is **fully manual**:

1. Identify which files changed and whether each is "canonical" (mirrored verbatim) or "templated" (mirrored with per-repo substitutions).
2. For each downstream repo: open a branch, copy the file, hand-merge any per-repo overrides, run CI locally, open a PR, drive review + merge per the full Phase 4 policy (Codex + CodeRabbit + reviewer-identity approval).
3. Track which repos have which version of which canonical file.

The cost is ~30 minutes per repo per propagation, multiplied by 7 repos, multiplied by however many PRs ship in Mergepath that touch canonical surfaces. The recurring real-world examples that motivated this issue:

- #148 (`pr-review-policy.yml` Dependabot exemption) — needs to land in 6 repos. One-line workflow fix; ~3 hours of propagation toil.
- #162 (`agent-review.yml` v9 github-script Dependabot guard) — same surface, same scope. Both are blocked on this manual path today.
- #161 (`label-gate.yml` for `needs-human-review`) — new file, all consumers.
- #165 (label-removal-guard hook + helper) — new files in `scripts/hooks/` and `scripts/`, plus `.claude/settings.json` edit.
- #155 (promote `scripts/gh-projects/`) — first-time mirror of a kit into all consumers.
- Every doc tightening in `REVIEW_POLICY.md` / `CLAUDE.md` / `AGENTS.md`.

The lack of a propagation tool is itself a forcing function for "let the canonical doc drift and we'll catch it in a weekly audit." That inverts the right priority: canonical-first, drift-as-exception.

## Goal

A single script that takes a Mergepath commit (or commit range, or PR) and ships PRs to every downstream consumer that opts in to that file's propagation, with per-repo overrides preserved, full Phase 4 review flow honored, and a clear summary of what landed where.

```
scripts/sync-to-downstream.sh <commit-ish> [--dry-run] [--repos r1,r2] [--files glob]
```

## Acceptance

[...]

PR #135, opened 2026-04-18
feat(packaging): scaffold mergepath placeholder packages (npm + PyPI)

nathanjohnpayne:
Authoring-Agent: claude

Addresses Phase 0 pre-flight: reserves the `mergepath` name on npm ([#92](https://github.com/nathanjohnpayne/mergepath/issues/92)) and PyPI ([#93](https://github.com/nathanjohnpayne/mergepath/issues/93)) by landing README-only placeholder scaffolds under a new `packaging/` directory. The human runs the actual `npm publish` / `twine upload` commands from this tree once creds are set up — this PR only lands the scaffolds and the publish runbook.

## Summary

- `packaging/npm/package.json`: `name: mergepath`, `version: 0.0.0`, `private: false`, `license: UNLICENSED`, `files: ["README.md"]`. The repo declares no open-source license, so UNLICENSED is accurate; `private: false` is required because npm blocks publish on `private: true`.
- `packaging/pypi/pyproject.toml`: PEP 621 metadata, `setuptools` build backend with `packages = []` so the sdist ships only the README, `License :: Other/Proprietary License` classifier.
- Both package READMEs point at the GitHub repo, mark the package as squatting-prevention only, and tell consumers not to depend on it.
- `packaging/README.md`: runbook with the exact `npm login` / `npm publish --access public` and `python3 -m build` / `twine upload dist/*` commands, plus a re-publish note (registries forbid overwriting existing versions — bump to `0.0.1` to refresh).
- `.repo-template.yml`: add `packaging` to `extra_top_level_dirs` so `scripts/ci/check_no_forbidden_top_level_dirs` stops warning.

Not in scope: running the publishes. Those require `npm login` and PyPI API token auth, which only the human can do. Issue closure belongs to the commit that happens after publish succeeds and `npm view mergepath` / `pip show mergepath` return the placeholder.

## Self-Review

**Correctness.** Manifests are minimal and standards-compliant. `package.json` passes `npm pkg fix` round-trip (single author string, standard repository shape). `pyproject.toml` uses PEP 621 `[project]` with `setuptools.build_meta` backend and `tool.setuptools.packages = []` — the canonical way to ship a README-only sdist without a src layout. All seven local `scripts/ci/check_*` checks exit 0 on this branch; only the pre-existing `bugs/` WARN remains (unrelated).

[...]

PR #175, opened 2026-05-02
feat(op-preflight): wire Cloudflare cache-purge token (#167)

### LINKED ISSUE #167: op-preflight: integrate "All Domains — Cache Purge API token" for Cloudflare cache busting
## Background

`scripts/deploy.sh` already implements a Cloudflare cache purge step as part
of the canonical deploy flow (see [DEPLOYMENT.md L437–L466](DEPLOYMENT.md)
and [scripts/deploy.sh L130–L153](scripts/deploy.sh)). It runs when both
`CF_API_TOKEN` and `CF_ZONE_ID` are present in the environment, otherwise
it logs `>> Cloudflare cache purge skipped (CF_API_TOKEN or CF_ZONE_ID not set)`
and continues.

Today nothing in our preflight pipeline materializes those env vars. The
relevant 1Password item — **"All Domains — Cache Purge API token"** — is
provisioned but unwired. Routine deploys silently skip the purge unless an
operator manually `op read`s the token into the shell first, which never
happens for agent-driven deploys (each tool call spawns a fresh subshell —
the same failure mode #139 fixed for PATs).

## Proposed change

Wire the Cloudflare cache-purge token into [scripts/op-preflight.sh](scripts/op-preflight.sh)
under `--mode deploy` (and therefore `--mode all`), so a single
`eval \"\$(scripts/op-preflight.sh --agent claude --mode all)\"` at session
start exports `CF_API_TOKEN` (and `CF_ZONE_ID`) into the cached session
file alongside the existing PATs and GCP ADC. Subsequent
`scripts/deploy.sh` calls then purge automatically without an extra
biometric prompt.

Concretely:

1. Add a new field to the preflight session-file schema, e.g.
   `OP_PREFLIGHT_CF_API_TOKEN`, sourced from the new 1Password item.
   Look up the item ID for **\"All Domains — Cache Purge API token\"** and
   record it in the script's lookup table next to `AUTHOR_PAT_ITEM` /
   `DEFAULT_ADC_OP_URI`. Do not guess the UUID — resolve it via
   `op item list --vault Private` and pin the canonical
   `op://Private/<item-id>/<field>` URI.
2. Decide where `CF_ZONE_ID` lives. Options:
   - Hardcode in `scripts/deploy.sh` (zone IDs are not secrets, but they
     are environment config — fine to commit).
   - Source from the same 1Password item as a sibling field.
   - Read from a new \`.env\`-style file checked into the repo.
   Pick one and document it in DEPLOYMENT.md.
3. Update the `--dry-run` output and the cache-hit fast path so missing
   `CF_API_TOKEN` is treated the same way missing ADC is treated today
   (refetch on `--mode deploy` / `--mode all`).
4. Update DEPLOYMENT.md L464 — the line that currently says \"typical
   source: \`op read 'op://...'\`\" — to point at the preflight env var
   instead.
5. Update CLAUDE.md's session-start checklist if the deploy flow's
   credential surface changes.

## Out of scope

- Switching the purge from `purge_everything: true` to a more targeted
  by-URL/by-tag purge. (Track separately if desired — the current
  blast-radius is already accepted in deploy.sh.)
- Building a standalone \`scripts/cf-purge.sh\` wrapper. The existing
  inline curl in deploy.sh stays the single call site for now.

## Acceptance

- After `eval \"\$(scripts/op-preflight.sh --agent claude --mode all)\"`,
  `env | grep -E 'CF_(API_TOKEN|ZONE_ID)'` shows both set.
- A clean `scripts/deploy.sh` run logs `>> Purging Cloudflare cache` and
  `   Cache purged.` (HTTP 200) instead of the skip line.
- `scripts/op-preflight.sh --agent claude --purge` removes the new field
  along with the existing ones.

nathanjohnpayne:
Closes #167.

## What

Wire the shared "All Domains — Cache Purge API token" 1Password item (`op://Private/4x6wslp3f6pal5t6h3jhhe63ie/credential`) into `scripts/op-preflight.sh` under `--mode deploy` (and therefore `--mode all`). After `eval "$(scripts/op-preflight.sh --agent <name> --mode all)"`, `CF_API_TOKEN` is exported in the parent shell AND persisted to the chmod-600 session file alongside the existing PATs and GCP ADC. The cache-hit fast path on subsequent invocations re-emits `CF_API_TOKEN` with no biometric prompt.

`DEPLOYMENT.md` L480 updated to reflect the new flow.

## Why

`scripts/deploy.sh` has implemented Cloudflare cache purge since #135 — it runs when both `CF_API_TOKEN` and `CF_ZONE_ID` are set in the env, otherwise logs a clear skip line and continues. Until now, nothing in the preflight pipeline materialized those env vars, so routine **agent-driven** deploys silently skipped the purge unless an operator manually `op read`-ed the token first — which never happens for agent flows since each tool call spawns a fresh subshell.

Same failure mode as #139 fixed for PATs.

## Sub-decision: where CF_ZONE_ID lives

This PR intentionally does **not** source `CF_ZONE_ID`. The shared CF token covers all domains, but each domain has its own zone — so `CF_ZONE_ID` is per-repo. Each downstream consumer sets its own zone ID in its bootstrap or hardcoded in its `scripts/deploy.sh` wrapper. (Mergepath itself has no Cloudflare zone behind a build artifact — `mergepath.nathanpayne.com` is a 301 to GitHub — so this PR has no observable effect on mergepath; it lights up the next time a downstream consumer with a zone deploys under preflight.)

If a future repo wants per-repo CF zone IDs in 1Password (e.g., a sibling field on each project's deploy item), that's a separate PR.

## Verification

- `bash -n` syntax check passes.
- `--dry-run` output now includes `Would read: Cloudflare cache-purge token (op://Private/.../credential)`.
- Live run with cached deploy session (no biometric): emits `export CF_API_TOKEN=...`; summary line `Cloudflare cache-purge token: loaded`.
- Cache-hit path's unset block + emit block updated so `emit_from_session_file` passes the cached value through cleanly. Tested by re-running `--mode all` after the initial fetch and confirming `CF_API_TOKEN` is re-exported with no biometric.
- Failure mode: if `op read` fails (1Password offline, item revoked), preflight emits a non-fatal warning + summary line `Cloudflare cache-purge token: SKIPPED (not available)`, and `deploy.sh` gracefully no-ops on the empty `CF_API_TOKEN` per its existing `[[ -z "${CF_API_TOKEN:-}" ]]` check.

## Authoring-Agent

Authoring-Agent: claude

## Self-Review

- **Correctness:** mirrors the GCP ADC wiring shape exactly (fetch in deploy block, EXPORTS/SESSION_LINES/SUMMARY arrays, cache-hit unset + emit). Differs in one place: `CF_API_TOKEN` is treated as **optional** in the cache-hit validation (vs. ADC which forces refresh on missing). Reason: deploy.sh gracefully no-ops on empty CF_API_TOKEN, so partial-cache penalty isn't worth it. Caller invariant preserved: a fresh `--mode deploy` always tries to fetch; a cache hit re-emits whatever was cached.
- **Regression risk:** zero for non-deploy modes (`--mode review` block unchanged). For deploy mode: a previously-cached session file written before this PR won't have `CF_API_TOKEN`, so the cache-hit path emits PATs + ADC as before with no CF_API_TOKEN export — equivalent to the pre-PR behavior. The next live `--mode deploy` (after TTL or `--refresh`) populates the cache with CF_API_TOKEN.
- **Style:** matches existing comment-block conventions; new constant lives next to `DEFAULT_ADC_OP_URI` with a short rationale. Header `Modes:` line updated for consistency.
- **Test coverage:** `bash -n` passes; live behavioral check on dry-run + cache-hit path. End-to-end deploy-with-purge requires a downstream repo with an actual CF zone — will verify on next downstream consumer that runs preflight + deploy.
- **Security/dependency hygiene:** no new dependencies, no new permissions. The CF token is read once via `op read` (same primitive as ADC), persisted to the same chmod-600 session file under `$XDG_CACHE_HOME/mergepath/`. No network call beyond the existing 1Password fetch. The token gets exported as `CF_API_TOKEN` to the eval-ing shell — same exposure model as `OP_PREFLIGHT_REVIEWER_PAT`.

## Propagation

After merge, this propagates to all 7 downstream consumers via the standard sync flow. No per-repo config required to receive the token — once preflight ships the new wiring, every downstream deploy-mode preflight gets `CF_API_TOKEN` automatically. Each consumer that wants the purge to actually fire needs to set its own `CF_ZONE_ID` (sub-decision deferred per above).

#168 tracks the propagation tool that will make this one-command instead of 7 manual sync PRs.

### DIFF
--- a/DEPLOYMENT.md
+++ b/DEPLOYMENT.md
@@ -477,7 +477,7 @@ The guards (see [mergepath#77](https://github.com/nathanjohnpayne/mergepath/issu
 
 Both guards can be bypassed with `--force` for break-glass scenarios. Never use `--force` during routine deploys.
 
-Cloudflare cache purge runs when `CF_API_TOKEN` and `CF_ZONE_ID` are set in the environment (typical source: `op read 'op://...'`). Without both variables the purge step no-ops with a clear log line.
+Cloudflare cache purge runs when `CF_API_TOKEN` and `CF_ZONE_ID` are set in the environment. `CF_API_TOKEN` is sourced automatically by `scripts/op-preflight.sh --mode deploy` (or `--mode all`) from the shared "All Domains — Cache Purge API token" 1Password item — no `op read` needed in your shell. `CF_ZONE_ID` is per-repo; each downstream consumer sets its own zone ID (e.g., in the repo's bootstrap or as a hardcoded value in its `scripts/deploy.sh` wrapper) since one CF token covers all domains but each domain has its own zone. Without both variables the purge step no-ops with a clear log line.
 
 **Do not run `op-firebase-deploy` or `firebase deploy` directly for routine deploys.** They skip the branch + freshness guards and the cache purge. Direct invocation is reserved for debugging or one-off flows where the deploy surface is known.
 
--- a/scripts/op-preflight.sh
+++ b/scripts/op-preflight.sh
@@ -26,7 +26,7 @@
 #
 # Modes:
 #   review  — reviewer PAT + author PAT + SSH key warming
-#   deploy  — GCP ADC credential
+#   deploy  — GCP ADC credential + Cloudflare cache-purge token
 #   all     — everything (default)
 #
 # Flags:
@@ -105,6 +105,15 @@ TTL_SECONDS="${OP_PREFLIGHT_TTL_SECONDS:-$DEFAULT_TTL_SECONDS}"
 
 # ── GCP ADC ───────────────────────────────────────────────────────────
 DEFAULT_ADC_OP_URI="${GCP_ADC_OP_URI:-op://Private/c2v6emkwppjzjjaq2bdqk3wnlm/credential}"
+
+# ── Cloudflare Cache Purge token (#167) ───────────────────────────────
+# Shared API token with Purge:Edit permission across all domains. Wired
+# into preflight so scripts/deploy.sh's existing CF purge step
+# (currently no-op when CF_API_TOKEN is unset) actually fires on
+# agent-driven deploys without an extra biometric prompt. CF_ZONE_ID
+# is intentionally NOT sourced here — it's per-repo and lives in each
+# downstream consumer's own bootstrap, not in this shared wiring.
+DEFAULT_CF_TOKEN_OP_URI="${CF_TOKEN_OP_URI:-op://Private/4x6wslp3f6pal5t6h3jhhe63ie/credential}"
 SSH_AUTHOR_HOST="github.com"
 
 # ── Parse arguments ───────────────────────────────────────────────────
@@ -195,6 +204,7 @@ if $DRY_RUN; then
   fi
   if [[ "$MODE" == "deploy" || "$MODE" == "all" ]]; then
     echo "# Would read: GCP ADC ($DEFAULT_ADC_OP_URI)" >&2
+    echo "# Would read: Cloudflare cache-purge token ($DEFAULT_CF_TOKEN_OP_URI)" >&2
   fi
   exit 0
 fi
@@ -359,6 +369,7 @@ emit_from_session_file() (
   # stdout+exit; parent stays clean.
   unset OP_PREFLIGHT_REVIEWER_PAT OP_PREFLIGHT_AUTHOR_PAT
   unset GOOGLE_APPLICATION_CREDENTIALS OP_PREFLIGHT_ADC_TMPFILE
+  unset CF_API_TOKEN
   unset OP_PREFLIGHT_DONE OP_PREFLIGHT_AGENT OP_PREFLIGHT_MODE
   unset OP_PREFLIGHT_CREATED_AT_EPOCH OP_PREFLIGHT_TTL_SECONDS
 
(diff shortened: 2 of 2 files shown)


nathanpayne-claude (commented):

Small change (~30 lines), shape mirrors the existing GCP ADC wiring exactly. Sub-threshold (no .github/** match), so direct merge after CI clears.

Spot-checks:
- The CF token treatment differs from ADC in one place: optional in cache-hit validation. Rationale in commit + PR body — deploy.sh's existing CF check (`[[ -z "${CF_API_TOKEN:-}" ]]`) gracefully no-ops on empty, so partial-cache penalty isn't worth it. ADC's stricter check exists because GCP ADC has its own staleness semantics (refresh-token rejection) that need explicit handling; CF tokens don't have that failure mode.
- CF_ZONE_ID intentionally not sourced — per-repo concern documented in PR body and DEPLOYMENT.md.
- No other repo file touches CF_* env vars (verified via grep), so no propagation drift in this PR.

Enabling auto-merge for CI clearance.

Issue #93, opened 2026-04-17
P0: Pre-flight: reserve `mergepath` on PyPI (placeholder)


nathanjohnpayne:
Upload a README-only placeholder package named `mergepath` to PyPI under `nathanjohnpayne` to prevent squatting.

Package should contain only a README, version `0.0.0`, no runtime code.

Done when: `pip show mergepath` (after install) or `https://pypi.org/project/mergepath/` shows the placeholder.

nathanpayne-claude:
Published. Verified at https://pypi.org/project/mergepath/0.0.0/ — name=`mergepath`, version=`0.0.0`, owner=nathanjohnpayne.

PR #176, opened 2026-05-02
feat: branch-protection audit script + canonical required-checks doc (#161)

### LINKED ISSUE #161: needs-human-review label does not block merges — gate is advisory only (matchline, possibly others)
## Observed in 2026-04-27 Weekly PR Audit

matchline#93 and matchline#76 were both merged while carrying the `needs-human-review` label. The label is intended to require human resolution before merge, but branch protection is not enforcing it.

## Current behavior

`needs-human-review` is applied as a label but there is no corresponding required status check tied to label presence. Any PR with this label can be merged freely.

## Expected behavior

A PR with `needs-human-review` should be blocked from merging until the label is removed (i.e., a human has reviewed and cleared it). This requires either:

1. A required status check that fails when `needs-human-review` is present and passes when it is absent (label-watcher workflow pattern), or
2. Branch protection rulesets that block merge based on label presence (GitHub's native label-based ruleset support).

## Scope

- matchline confirmed affected.
- Audit other repos (nathanpaynedotcom, friends-and-family-billing, device-platform-reporting, overridebroadway, swipewatch) to check whether they carry this label pattern and whether the same gap exists.

## Fix

Add a `label-gate.yml` workflow (or equivalent) to Mergepath that:
- Runs on `pull_request` events for label changes and pushes
- Fails the required status check when `needs-human-review` is present
- Passes (and passes required check) when the label is absent

Propagate to all repos that use this label.

nathanjohnpayne:
Closes #161 from the mergepath side. Per-repo branch protection setup is downstream config (not in template source) — this PR ships the audit tooling + documentation that make the gap visible.

## What

1. **`scripts/audit-branch-protection.sh`** — read-only audit. Reads branch protection (with rulesets API fallback) via gh API and reports whether each canonical mergepath check is required. Exit codes:
   - `0` — all canonical checks are required
   - `1` — bad arguments
   - `2` — gh API failure (auth, missing repo, network)
   - `3` — gating gap (one or more canonical checks not required), with fix recipe in stderr

   Canonical list: `Label Gate`, `Self-Review Required`. Keep in sync with `.github/workflows/` as new gating jobs ship.

2. **`REVIEW_POLICY.md` § Implementation notes for branch protection gates** gains a "Required status checks (per-repo setup)" subsection listing the canonical checks explicitly and pointing at the audit script.

## Why

#161's root cause is **not** a missing workflow — `pr-review-policy.yml`'s `Label Gate` job exists and is correct. The cause is per-repo branch protection: matchline ran the workflow, the workflow failed, but the failed check was **advisory** because Label Gate wasn't in the required-checks list. So PRs with `needs-human-review` merged anyway (matchline #76, #93).

Branch protection is per-repo config that isn't in any template's source tree. The mergepath-side fix is making the gap visible:
- Documentation says explicitly "these workflows must be required status checks."
- Audit script makes verification one command.

## Verification

Smoke tests:
- `--help` exits 0 with usage block.
- `--repo "not-valid"` exits 1 with `Invalid --repo value` error.
- `--repo "owner/name"` against a non-existent repo exits 2 (gh API 404).
- Against `nathanjohnpayne/mergepath` (no branch protection): exits 3 with `PR merges are completely unprotected.` message.
- Rulesets API fallback wired in (untested live; reads handled correctly per gh's rulesets schema). Worth a follow-up if a consumer hits the rulesets path and finds an issue.

**Caveat (drive-by finding):** mergepath itself currently has no branch protection on `main` per the audit. The audit script flagged this as exit 3. Worth filing as a follow-up issue but not in scope here.

## Authoring-Agent

Authoring-Agent: claude

## Self-Review

- **Correctness:** the canonical-check matcher uses `grep -Fxq` (fixed-string, full-line) so partial matches like `"Self-Review"` don't false-pass against `"Self-Review Required"`. The classic-vs-rulesets fallback is detected by checking the `"message":` key in the protection response (gh API's standard 404 shape). Repo-name regex matches the same shape used in `resolve-pr-threads.sh` (consistent with what's already in tree).
- **Regression risk:** zero — new file + new doc paragraph. No existing behavior changes.
- **Style:** matches existing CI script conventions in `scripts/ci/` (set -eo pipefail header; clear error → exit code mapping; --help block). The script lives at `scripts/` rather than `scripts/ci/` because it's an operator-facing audit tool, not a CI gate that runs on every commit (running it in CI would be circular — the audit would itself be one of the required checks).
- **Test coverage:** smoke matrix covered above. The classic-protection happy path is covered indirectly by the canonical-check matcher logic (tested against synthetic input via the matcher's regex). End-to-end against a real repo with proper protection requires running on a downstream consumer that has Label Gate marked required — will verify on next consumer setup.
- **Security/dependency hygiene:** read-only API calls. Requires `Administration:read` scope on the target repo, which existing reviewer/author PATs already have. No secrets, no writes, no env vars.

## Propagation

After merge: this propagates to all 7 downstream consumers via the standard sync flow. Each consumer can run `scripts/audit-branch-protection.sh` immediately to verify their own setup; the script defaults to current-repo + `main`.

#168 tracks the propagation tool.

### DIFF
--- a/REVIEW_POLICY.md
+++ b/REVIEW_POLICY.md
@@ -434,6 +434,19 @@ Agents must never modify the `needs-external-review`, `needs-human-review`, or `
 
 ## Implementation notes for branch protection gates
 
+### Required status checks (per-repo setup)
+
+The workflows shipped under `.github/workflows/` ENFORCE merge gating only when configured as **required status checks** in branch protection. Without that bit, a failed check is advisory and the PR merges anyway. The 2026-04-27 weekly audit caught `matchline#76` and `matchline#93` merging past `needs-human-review` for exactly this reason — `Label Gate` was running and failing, but wasn't in the required-checks list.
+
+Each repo using this template must mark these as required on `main` (Settings → Branches → Branch protection rule):
+
+- **`Label Gate`** — fails when any of `needs-external-review`, `needs-human-review`, or `policy-violation` is on the PR. The hard gate behind the doctrine in [Agent prohibitions](#agent-prohibitions).
+- **`Self-Review Required`** — fails when the PR body lacks a `## Self-Review` section (Dependabot-exempt).
+
+Audit a repo's branch protection with `scripts/audit-branch-protection.sh` (read-only; exits 3 if any canonical check is not required, with a fix recipe). Re-run after every protection change.
+
+### `resolveReviewThread`
+
 The GitHub GraphQL `resolveReviewThread` mutation may be used by agents **only** when both:
 
 - The agent has demonstrably addressed the inline finding (a fix is on the current HEAD, or a rebuttal is posted on the thread), AND
--- a/scripts/audit-branch-protection.sh
+++ b/scripts/audit-branch-protection.sh
@@ -1,0 +1,158 @@
+#!/usr/bin/env bash
+# audit-branch-protection.sh — verify branch protection on `main` enforces
+# the canonical required status checks shipped by mergepath.
+#
+# Background: pr-review-policy.yml's `Label Gate` job FAILS when a blocking
+# label (`needs-human-review`, `policy-violation`, `needs-external-review`)
+# is on the PR. But that failure only blocks merge if the workflow is
+# configured as a REQUIRED status check in branch protection. Without the
+# protection bit, the failed check is advisory and PRs merge anyway. This
+# is the gap that motivated nathanjohnpayne/mergepath#161 (matchline #93,
+# #76 merged past `needs-human-review`).
+#
+# Same gap applies to other workflows (Self-Review Required, agent-review
+# pipeline jobs). This audit reads branch protection via gh API and reports
+# whether each canonical check is required.
+#
+# Usage:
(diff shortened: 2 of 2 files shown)


nathanpayne-claude (commented):

Small surface (~170 lines added across 1 new script + 1 doc paragraph). Sub-threshold, no .github/** match. The audit script handles both classic branch protection AND the modern rulesets API as a fallback.

Drive-by finding (worth a separate issue): mergepath itself has zero branch protection on `main` per the audit. Could explain why some of my recent PRs in this session were able to merge with mergeStateStatus=UNKNOWN — there's no protection at all to enforce CLEAN. Worth filing.

Spot-checks:
- `grep -Fxq` for canonical-check matching avoids partial-match false-passes (e.g., "Self-Review" wouldn't match "Self-Review Required").
- Rulesets fallback path is wired but untested live — uses gh's documented schema. If a consumer hits it and finds an issue, fix-forward.
- Script lives at `scripts/` not `scripts/ci/` — it's operator-facing, not a CI gate (CI gating it would be circular).

Enabling auto-merge for CI clearance.

PR #178, opened 2026-05-02
fix(op-preflight): track SSH-warm freshness independently of PAT cache (#163)

### LINKED ISSUE #163: op-preflight.sh: biometric prompts still fire mid-session despite chmod-600 cache (regression of #139)
Authoring-Agent: claude

## Problem

The chmod-600 session-file cache shipped in #141 (with follow-ups #142, #143, #144, #145, #146) was supposed to make `scripts/op-preflight.sh` durable across agent `Bash` tool calls — one biometric at session start, no more for ~4h. In practice, during today's review session (2026-04-28) the human had to provide biometric repeatedly throughout the review-and-merge loop, the same failure mode that motivated #139.

Confirming the cache file IS being written:

```
~/.cache/mergepath/
  -rw-------@ op-preflight-claude.env   (modified today)
  -rw-------@ op-preflight-codex.env
  -rw-------@ op-preflight-cursor.env
```

So the file is fresh, perms are right, TTL hasn't expired — yet biometrics still fire. That means the friction is coming from a code path that bypasses or re-traverses biometric on top of the cache hit.

### 1. SSH warming runs on every cache hit (highest likelihood)

`op-preflight.sh:420-422` calls `warm_ssh_keys` on the fast-path cache-hit branch:

```bash
if [[ "\$MODE" == "review" || "\$MODE" == "all" ]] && ! \$SKIP_SSH; then
  warm_ssh_keys
fi
```

`warm_ssh_keys` runs `ssh -T git@github.com` and `ssh -T git@github-claude`. Both go through the **1Password SSH agent**, which has its **own** session TTL — independent of the chmod-600 PAT cache. When the SSH-agent session expires (typically much shorter than 4h), every preflight invocation re-prompts for biometric on the SSH side, even though the PAT side is hitting the file cache cleanly.

This was added in one of the propagation-round Codex fixes (the rationale is in the comment at line 416-418: \"the prior implementation skipped this step entirely on cache hit\"). The fix protected against unprimed git push/pull, but it now causes biometric on every tool call that re-runs preflight.

**Fix candidates:**
- Track SSH-warm freshness with its own marker file (e.g. `op-preflight-\$AGENT.ssh-warmed`) and only warm if older than N minutes.
- Add `--skip-ssh-on-cache-hit` (or invert: only warm SSH on full-fetch path, document that the first preflight of a session warms SSH and subsequent re-runs trust the SSH agent).
- Detect whether the 1Password SSH agent session is still live (e.g. probe with `SSH_ASKPASS_REQUIRE=never ssh -o BatchMode=yes -T git@github.com`) before triggering a warm that will prompt.

### 2. Some commands still use inline `op read` instead of `\$OP_PREFLIGHT_REVIEWER_PAT`

CLAUDE.md tells the agent to use `GH_TOKEN=\"\$OP_PREFLIGHT_REVIEWER_PAT\"` after running preflight. But each fresh `Bash` subshell starts with empty env, so the agent has to **either** re-run preflight at the top of every call **or** fall back to inline `op read 'op://Private/<id>/token'`. The latter always triggers biometric.

A grep across `scripts/`, `.github/workflows/`, and the agent docs would catch any remaining inline `op read` calls in the review hot-path that should be using the env vars.

### 3. `op inject` / `op read` triggering on partial-cache fall-through

If `emit_from_session_file` returns 2 (mode mismatch / missing field), the script falls through to the full-fetch path on `op-preflight.sh:436-`, which calls `op inject` — biometric. Worth verifying: across a session, are we ever taking the fall-through path because of mode drift between callers (one caller `--mode all`, another `--mode review`)? The header comment at lines 333-344 specifically calls out the `all`-mode contract, but a caller invoking `--mode review` against an `--mode all`-populated cache should hit the fast path; need to confirm with logs.

## Reproduction

In a Claude Code session that's been running ≥1h:

1. Run `eval \"\$(scripts/op-preflight.sh --agent claude --mode all)\"` — expect biometric (first time).
2. Wait 5+ minutes, run a `gh pr view` or `gh pr review` that the agent shells out for — observe whether biometric fires again.
3. If it does, instrument: add `set -x` at the top of `op-preflight.sh`, re-run, check whether `warm_ssh_keys` is the prompting call vs. an inline `op read` somewhere downstream.

## Out of scope

- Reducing the 1Password SSH-agent TTL knob — that lives in the 1Password app, not this repo.
- Switching to service-account tokens — was option 1 in #139 and explicitly not chosen.

## Related

- #139 — original biometric-friction issue (closed by #141)
- #141 — feature PR that introduced the chmod-600 session file
- #142, #143, #144, #145, #146 — follow-up fixes addressing edge cases (cross-mode validation, stale ADC, errexit fallback, partial-hit reversal). None of these targeted the SSH-warming-on-cache-hit path or the agent-side inline-`op read` audit.
- #137 — adjacent firebase reauth daily-friction issue (different surface, same class of problem)

## Self-Review

- **Correctness:** This issue is observational and proposes hypotheses, not a fix — the actual diagnosis needs the `set -x` repro in step 3 above to confirm which path is triggering biometric.
- **Regression risk:** None for filing the issue; whatever fix lands needs to preserve the SSH-warming behavior fixed in the propagation round.
- **Style:** N/A.
- **Test coverage:** Whatever fix lands should add a test (or at minimum a manual repro recipe) for \"second invocation within TTL on a warm cache emits no biometric prompt across both PAT and SSH paths.\"
- **Security / dependency hygiene:** No change to attack surface.

nathanjohnpayne:
Closes #163.

## What

Track SSH-warm freshness independently of the PAT cache — separate marker file (`$CACHE_DIR/op-preflight-$AGENT.ssh-warmed`) with a separate (default 30 min) TTL. Skip the warm if recent enough. New env var `OP_PREFLIGHT_SSH_WARM_TTL_SECONDS` to override.

## Why

The 1Password SSH agent has its OWN session TTL — independent of the chmod-600 PAT cache and typically MUCH shorter (a few minutes to an hour, vs the PAT cache's 4h). When the SSH-agent session expires inside our PAT cache window, every preflight invocation hits the PAT-cache fast path, finds PATs valid, then re-fires biometric on the SSH-warm side.

This is hypothesis 1 from #163, confirmed: in this very session I observed multiple `gh` shellouts re-firing biometric despite a warm cache file. Once I ship and merge this fix, future agent sessions get one biometric burst per `OP_PREFLIGHT_SSH_WARM_TTL_SECONDS` window instead of one per cache-hit invocation.

## Behavior matrix

| Invocation | Marker state | Result |
|------------|--------------|--------|
| First (no marker) | absent | warm both keys, create marker (mode 600) |
| Second within TTL | fresh | skip; log `SSH keys recently warmed (Ns ago / TTL 1800s)` |
| `--refresh` | any | delete marker → warm |
| `--skip-ssh` | any | bypass entirely (existing behavior unchanged) |
| `--purge` | any | remove marker + session file + ADC tempfile |
| `--purge-all` | any | `find` pattern picks up `op-preflight-*.ssh-warmed` |

## Trade-off

If the 1Password SSH agent session expires INSIDE our `SSH_WARM_TTL_SECONDS` window, the next git push/pull still triggers biometric — but at least preflight itself doesn't multiply that. The default 30 min strikes a balance: short enough that most 1Password agent sessions outlast it, long enough to cover typical agent-tool-call cadence in a single session.

For a hotter session (lots of git push activity), set `OP_PREFLIGHT_SSH_WARM_TTL_SECONDS=300` (5 min) to keep the agent fresh. For a cold session (mostly read-only), set it to `7200` (2h) or higher.

## Verification

Live four-scenario smoke test in the commit message — all four cases pass against my actual environment.

## Authoring-Agent

Authoring-Agent: claude

## Self-Review

- **Correctness:** mtime-based freshness uses both BSD `stat -f %m` and GNU `stat -c %Y` (cross-platform). Falls back to `0` on stat failure → freshness check fails → re-warm. Marker is touched AFTER both warms attempt; failures to touch are non-fatal.
- **Regression risk:** non-cache-hit paths unchanged (full-fetch path still warms unconditionally on the first invocation since the marker doesn't yet exist). `--skip-ssh` still works as before. `--refresh` is consistently honored across both PAT cache + SSH marker. `--purge` / `--purge-all` clean the new marker.
- **Style:** matches existing comment-block conventions in the file (banner explanation, history reference, why-not). New `ssh_warm_is_fresh` helper mirrors the existing `session_is_fresh` shape.
- **Test coverage:** four-scenario smoke matrix. End-to-end "biometric only fires once per 30-min window" requires observing actual 1Password biometric prompts over time — will verify by running over the next session and noting whether mid-session biometric prompts disappear.
- **Security/dependency hygiene:** new file is mode 600 in a 700 cache dir (matches existing session file). No secrets in the marker — it's an empty file whose mtime is the only signal. No new dependencies, no new env vars exposed beyond the documented `OP_PREFLIGHT_SSH_WARM_TTL_SECONDS`.

## Propagation

After merge: propagates to all 7 downstream consumers via the standard sync flow. No per-repo config required to receive the new behavior — the marker is opt-in only via the `--skip-ssh` flag and the env var override.

### DIFF
--- a/scripts/op-preflight.sh
+++ b/scripts/op-preflight.sh
@@ -45,6 +45,14 @@
 #                             file does NOT extend its effective lifetime.
 #   OP_PREFLIGHT_CACHE_DIR    Override cache dir (default
 #                             $XDG_CACHE_HOME/mergepath or $HOME/.cache/mergepath).
+#   OP_PREFLIGHT_SSH_WARM_TTL_SECONDS
+#                             Override SSH-warm freshness window (default
+#                             1800s = 30 min). Independent of the PAT
+#                             cache TTL because the 1Password SSH agent
+#                             has its own session lifetime, typically
+#                             shorter than 4h. Skipping re-warm within
+#                             this window prevents a biometric prompt on
+#                             every cache-hit invocation. See #163.
 #
 # Session file:
 #   Path:        $cache_dir/op-preflight-<agent>.env
@@ -146,7 +154,7 @@ done
 if $PURGE_ALL; then
   if [[ -d "$CACHE_DIR" ]]; then
     echo "# Purging all session files under $CACHE_DIR" >&2
-    find "$CACHE_DIR" -maxdepth 1 -type f \( -name 'op-preflight-*.env' -o -name 'op-preflight-*-adc.json' \) -print -delete >&2
+    find "$CACHE_DIR" -maxdepth 1 -type f \( -name 'op-preflight-*.env' -o -name 'op-preflight-*-adc.json' -o -name 'op-preflight-*.ssh-warmed' \) -print -delete >&2
   fi
   exit 0
 fi
@@ -170,11 +178,13 @@ fi
 # ── Cache paths (deterministic per agent) ─────────────────────────────
 SESSION_FILE="$CACHE_DIR/op-preflight-$AGENT.env"
 ADC_TMPFILE="$CACHE_DIR/op-preflight-$AGENT-adc.json"
+SSH_WARM_MARKER="$CACHE_DIR/op-preflight-$AGENT.ssh-warmed"
+SSH_WARM_TTL_SECONDS="${OP_PREFLIGHT_SSH_WARM_TTL_SECONDS:-1800}"  # 30 min default; #163
 
 # ── Purge mode ────────────────────────────────────────────────────────
 if $PURGE; then
-  rm -f "$SESSION_FILE" "$ADC_TMPFILE"
-  echo "# Purged session file + ADC tempfile for agent=$AGENT" >&2
+  rm -f "$SESSION_FILE" "$ADC_TMPFILE" "$SSH_WARM_MARKER"
+  echo "# Purged session file + ADC tempfile + SSH-warm marker for agent=$AGENT" >&2
   exit 0
 fi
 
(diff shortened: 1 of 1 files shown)


nathanpayne-claude (commented):

~53 lines added to one script. Sub-threshold, no .github/** match. Auto-merge enabled.

Spot-checks:
- mtime stat is cross-platform (BSD `-f %m` + GNU `-c %Y` with `|| echo 0` fallback). On stat failure, age computes against epoch 0 → freshness check fails → re-warm. Conservative.
- Marker is touched AFTER both warms attempt. If either ssh -T failed (network/key issue), marker still gets set so we don't re-prompt the user with the same broken state on the next call. Real failures surface on the next git push, not in preflight.
- The default 30-min TTL is a guess — the 1Password SSH agent's actual session length varies by user config (1Password app settings → "Lock 1Password" timeout). 30 min is below typical Mac defaults but generous enough not to multiply biometric per cache-hit. Adjustable via env.

Issue #148, opened 2026-04-19
pr-review-policy.yml: Dependabot self-review exemption uses github.actor, not PR author


nathanjohnpayne:
Codex surfaced this on friends-and-family-billing#228 during the agent-review workflow propagation:

> The condition checks `github.actor`, but that value is the user/app that triggered the event, not the PR author. On Dependabot-authored PRs, later `pull_request` events like `labeled`, `unlabeled`, or `edited` are commonly triggered by a human or `github-actions[bot]`, so this job now runs and fails on missing `## Self-Review` even though Dependabot PRs are supposed to be exempt. This can leave required checks red and block merging dependency updates after normal label workflows.

## Current (line 14 of .github/workflows/pr-review-policy.yml)

```yaml
if: github.actor != 'dependabot[bot]'
```

## Proposed fix

```yaml
if: github.event.pull_request.user.login != 'dependabot[bot]'
```

`github.event.pull_request.user.login` is the PR **author**, which is stable across re-fire events (labeled, unlabeled, edited, synchronize). `github.actor` is who triggered the current event.

## Scope

- Mergepath main: fix in a new PR.
- Then propagate to the 6 downstream consumers (device-platform-reporting, device-source-of-truth, friends-and-family-billing, nathanpaynedotcom, overridebroadway, swipewatch). All carry the identical buggy condition today.

## Relation to friends-and-family-billing#228

The finding surfaced on that PR but is not a regression — both the pre-propagation and post-propagation workflow have the same buggy check. Dismissing on that PR as pre-existing; fix ships here.

nathanjohnpayne:
After checking the current repo state, I think the fix should stay narrowly scoped to the self-review gate rather than doing a blanket actor→author replacement.

Plan:
1. Patch the canonical workflow in mergepath: change `.github/workflows/pr-review-policy.yml` `jobs.self-review-check.if` from `github.actor != 'dependabot[bot]'` to `github.event.pull_request.user.login != 'dependabot[bot]'`. This gate is about PR-author semantics, and `pull_request.user.login` stays stable across `opened`, `edited`, `synchronize`, `labeled`, and `unlabeled` events.
2. Update the nearby explanatory text so the docs match the code. `.github/workflows/pr-audit.yml` currently explains the Dependabot exemption by referencing the old actor-based check, so that comment should be rewritten at the same time.
[...]

PR #217, opened 2026-05-04
feat(sync): canonical-only sync mode for downstream propagation (#168 layer 3 slice 1)

nathanjohnpayne:
Builds on [#215](https://github.com/nathanjohnpayne/mergepath/pull/215) (manifest + read-only audit). Implements the first slice of Layer 3 sync mode per [#168](https://github.com/nathanjohnpayne/mergepath/issues/168).

## What this slice ships

`scripts/sync-to-downstream.sh <commit-ish> [--dry-run] [--repos r1,r2] [--paths glob]`

- **Canonical paths only.** Kit and templated paths are detected when a commit touches them, and called out in the per-consumer summary as deferred (they land in slices 2 and 5).
- **Idempotency** via deterministic branch name `mergepath-sync/<short-sha>`. Existing open PR → "already in flight"; closed/merged → "already done". Skipped in dry-run (which is meant to be network-free).
- **Dry-run mode** prints the full per-consumer plan: branch, canonical files, deferred kit/templated counts. Zero side effects.
- **Live mode** clones each consumer into a fresh tmpdir workspace (never reuses `MERGEPATH_SYNC_CACHE` — that's read-only audit territory), copies files via `git show <sha>:<path>` (so Mergepath uncommitted edits don't leak), preserves the executable bit when the source has it, commits with the standard `Authoring-Agent` + `## Self-Review` body, pushes the deterministic branch, and creates a PR via `gh pr create` against the consumer's `main`.

## Dogfood verification

Dry-run against the #148 fix commit (`c3b0df0`, the dependabot exemption fix) plans the right thing:

```
$ scripts/sync-to-downstream.sh c3b0df0 --dry-run
Sync from mergepath@c3b0df0: fix(workflows): use PR author login for Dependabot exemptions (#148, #162) (#169)
Changed files at this commit:
  .github/workflows/agent-review.yml
  .github/workflows/pr-review-policy.yml

  ⤷ matchline — would open PR on branch mergepath-sync/c3b0df0 (2 canonical file(s))
      + .github/workflows/agent-review.yml
      + .github/workflows/pr-review-policy.yml
  [...6 more consumers...]

Summary: PRs opened/planned: 7  skipped: 0  failed: 0
```

This matches the propagation backlog the audit surfaced in #215. Once this PR merges, the next step is to actually run it against `c3b0df0` in live mode and shepherd the resulting 7 PRs through their respective review flows. That's a separate operational step, not part of this code change.

## Test coverage

The fixture test (`tests/test_sync_to_downstream.sh`) now covers:

- Multi-type touch: canonical lands; kit+templated noted as deferred per consumer.
- `--repos` filter restricts the consumer set.
- `--paths` filter restricts target paths within the planned PR.
- Kit-only commit reports "no canonical targets" per consumer.
- Unrelated-only commit reports "no manifest paths touched".
- Unknown commit-ish exits 2 cleanly.

The live clone/push/PR-create cycle is intentionally NOT exercised in the test fixture — that would require either stubbing `gh` via PATH override (a lot of plumbing for limited additional coverage given the planning logic is fully tested) or hitting the real network (defeats the purpose of a unit test). The live path is conceptually simple file-copy + git plumbing and was verified by hand via the dry-run smoke against `c3b0df0` above.

Authoring-Agent: claude

## Self-Review

- **Correctness.** Sync mode resolves the commit-ish in the Mergepath worktree (errors out cleanly on unknown), enumerates changed files at that commit, intersects with manifest canonical paths per consumer, and either prints the dry-run plan or runs the full clone→branch→commit→push→PR cycle. Materializes content via `git show <sha>:<path>` rather than working-tree reads so an unrelated uncommitted edit in the Mergepath worktree can't leak into a propagation PR. Preserves executable bit when the source tree has `100755` mode.
- **Regression risk.** Audit mode behavior is unchanged — verified by re-running the existing smoke against `matchline / scripts/op-preflight.sh` (still reports the same 177-line drift). New code is gated entirely behind the positional commit-ish arg path; existing flag-based modes (`--audit`, `--help`, `--version`) untouched.
- **Style.** Bash + yq + gh; matches the conventions of [scripts/codex-review-check.sh](scripts/codex-review-check.sh) and the v1 audit helpers. Heredoc bodies for commit + PR body avoid apostrophes (bash parser is grumpy about apostrophes inside `$(cat <<EOF ...)` even though the spec says they should be literal). One inline note on this in the source.
- **Test coverage.** Six new dry-run assertion cases in the fixture (counted above). Live cycle is documented as out-of-scope for this PR's tests with rationale. Existing fixture cases (audit, symlink guard, worktree `.git` file, arg validation) still pass — verified locally.
- **Security / dependency hygiene.** No new external dependencies. Live mode uses `gh repo clone` + `git push` + `gh pr create` — same surface as our existing PR workflow. Workspace is a `mktemp -d` and torn down via `trap RETURN` so a failure mid-cycle doesn't leak credentials or partial commits onto disk.

## Test plan

- [x] `tests/test_sync_to_downstream.sh` — full suite passes
- [x] `scripts/ci/check_sync_manifest` — passes
- [x] `bash -n scripts/sync-to-downstream.sh` — clean parse
- [x] `scripts/sync-to-downstream.sh --version` / `--help` — print
- [x] `scripts/sync-to-downstream.sh --audit ...` — unchanged (177-line matchline drift still reported correctly)
- [x] `scripts/sync-to-downstream.sh c3b0df0 --dry-run` — plans 7 PRs as expected
- [x] `scripts/sync-to-downstream.sh c3b0df0 --dry-run --repos matchline` — single-consumer plan
- [x] `scripts/sync-to-downstream.sh c3b0df0 --dry-run --paths ".github/workflows/pr-review-policy.yml"` — single-path plan
- [ ] CI green on this PR
- [ ] Post-merge: dogfood live propagation against `c3b0df0` (separate operational step)

### DIFF
--- a/scripts/sync-to-downstream.sh
+++ b/scripts/sync-to-downstream.sh
@@ -735,8 +735,16 @@ sync_open_pr() {
   local templated_list=$8
   local short_sha=${sha:0:7}
 
+  # Portable mktemp: `-t TEMPLATE` semantics differ between BSD (macOS)
+  # and GNU coreutils. macOS treats TEMPLATE as a literal prefix and
+  # appends random chars; GNU treats it as a TMPDIR-rooted basename
+  # but ONLY if the value contains no slash, AND the trailing X-pattern
+  # rules differ. Use the explicit-path form `mktemp -d "$TMPDIR/<X...>"`
+  # which both implementations honor identically. cursor CHANGES_REQUESTED
+  # on PR #217 caught the GNU/Linux portability gap.
+  local tmp_root=${TMPDIR:-/tmp}
   local workspace
-  workspace=$(mktemp -d -t "mergepath-sync-$consumer_name") || {
+  workspace=$(mktemp -d "$tmp_root/mergepath-sync-${consumer_name}.XXXXXX") || {
     err "could not create workspace tmpdir"
     return 1
   }
@@ -751,10 +759,30 @@ sync_open_pr() {
   # Materialize Mergepath's $sha worktree state for the target files.
   # Using `git show $sha:<path>` is robust against the working tree
   # having other uncommitted edits.
+  #
+  # Deletion handling: when Mergepath drops a canonical file at $sha,
+  # `git ls-tree $sha <path>` returns empty (the path no longer exists
+  # in the tree). Treat that as a delete propagation: rm the consumer
+  # copy if present. cursor CHANGES_REQUESTED on PR #217 caught the
+  # missing delete propagation — without this branch, deletes would
+  # fail noisily on `git show` and the script would abort the entire
+  # consumer rather than mirroring the delete.
   local target
   while IFS= read -r target; do
     [ -z "$target" ] && continue
     local consumer_target="$workspace/repo/$target"
+    local src_mode
+    src_mode=$(git -C "$MERGEPATH_ROOT" ls-tree "$sha" -- "$target" | awk '{print $1}')
+
+    if [ -z "$src_mode" ]; then
+      # Path absent at $sha → delete propagation.
+      if [ -e "$consumer_target" ]; then
+        rm -f "$consumer_target"
+      fi
+      # No mode-mirror for deletes; the path is gone in both trees.
+      continue
+    fi
+
     mkdir -p "$(dirname "$consumer_target")"
     if ! git -C "$MERGEPATH_ROOT" show "$sha:$target" >"$consumer_target" 2>/dev/null; then
       err "$consumer_name: could not read $target from mergepath@$short_sha"
@@ -767,8 +795,6 @@ sync_open_pr() {
     # consumer historically had +x set on a file Mergepath later
     # decided should be plain). cursor's CHANGES_REQUESTED on PR #217
     # caught the one-way version that only added +x.
-    local src_mode
-    src_mode=$(git -C "$MERGEPATH_ROOT" ls-tree "$sha" "$target" | awk '{print $1}')
     case "$src_mode" in
       100755) chmod +x "$consumer_target" ;;
       100644) chmod -x "$consumer_target" ;;
--- a/scripts/sync-to-downstream.sh
+++ b/scripts/sync-to-downstream.sh
@@ -4,16 +4,22 @@
 #
 # This script is shipping in layers (per the issue's implementation plan):
 #
-#   v1 (this PR):
+#   v1 (PR #215):
 #     --audit            Read-only drift detector across all consumers.
-#     --help             Print this header.
-#     --version          Print script + manifest schema version.
+#
+#   v2 (this PR — Layer 3 first slice):
+#     <commit-ish>       Open propagation PRs for canonical files changed
+#                        at the given commit. Kit and templated paths are
+#                        skipped with a warning (deferred to slice 2/3).
 #
 #   future:
-#     <commit-ish>       Open propagation PRs for files changed at the
-#                        given commit (Layer 3).
 #     --from-pr <N>      Resolve PR N's merge commit and propagate
 #                        (Layer 4).
+#     templated paths    Three-way merge for review-policy.yml; substitution
+#                        rules for AGENTS.md / CLAUDE.md (Layer 5, shared
+#                        with bootstrap-new-repo.sh #156).
+#     kit paths          Directory mirror with allow-extras semantics in
+#                        sync mode (slice 2 — already supported in audit).
 #
 # The manifest at .mergepath-sync.yml declares which paths are canonical
 # (byte-identical) or kit (directory mirror with allow-extras), and which
@@ -22,20 +28,29 @@
 #
 # Usage:
 #   scripts/sync-to-downstream.sh --audit [--repos r1,r2] [--paths glob]
+#   scripts/sync-to-downstream.sh <commit-ish> [--dry-run] [--repos r1,r2] [--paths glob]
 #   scripts/sync-to-downstream.sh --help
 #   scripts/sync-to-downstream.sh --version
 #
 # Flags:
 #   --audit            Read-only drift detection. Exit 0 (clean), 1 (drift),
 #                      2 (script/usage error), 3 (consumer fetch error).
+#   --dry-run          Sync mode only. Print the per-consumer plan
+#                      (branch name, files) without cloning, committing,
+#                      pushing, or creating PRs. Idempotency check is
+#                      skipped because it probes the consumer repo via
+#                      `gh api`; a dry-run plan may show "would open PR"
+#                      even when a PR already exists, but the live run
+#                      will catch and skip it.
 #   --repos r1,r2      Restrict to a comma-separated subset of consumer names.
 #   --paths glob       Restrict to manifest paths matching the glob (e.g.
 #                      "scripts/*", ".github/workflows/agent-review.yml").
-#   --no-clone         Don't clone-on-demand; only audit consumers with a
-#                      local sibling worktree under MERGEPATH_SIBLINGS_DIR.
-#   --no-refresh       Don't `git fetch` cached consumer clones before
-#                      comparing. Useful for offline/sandboxed audits;
-#                      may report stale results if the cache is old.
+#   --no-clone         Audit only: don't clone-on-demand; only audit
+#                      consumers with a local sibling worktree under
+#                      MERGEPATH_SIBLINGS_DIR.
+#   --no-refresh       Audit only: don't `git fetch` cached consumer clones
+#                      before comparing. Useful for offline/sandboxed
+#                      audits; may report stale results if the cache is old.
 #   --help, -h         Show this help.
 #   --version          Print version info.
 #
@@ -74,9 +89,10 @@ set -euo pipefail
 
 # --- constants --------------------------------------------------------------
 
-SCRIPT_VERSION="0.1.0-layer1+2"
+SCRIPT_VERSION="0.2.0-layer3-canonical"
 SUPPORTED_MANIFEST_VERSION=1
 MANIFEST_PATH=".mergepath-sync.yml"
+SYNC_BRANCH_PREFIX="mergepath-sync"
 
 # Resolve the Mergepath worktree root from the script's location (works
 # regardless of cwd). Two `dirname`s: scripts/sync-to-downstream.sh →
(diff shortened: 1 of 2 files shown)


nathanpayne-claude (commented):
Reviewing as nathanpayne-claude.

The slice scope is clean: sync mode only fires on a positional commit-ish, all flag-based modes are untouched, and the audit fixture's existing assertions still pass. The canonical-only approach with kit/templated explicitly noted as "deferred this slice" matches the issue plan and keeps the blast radius small for the first live propagation.

Notes worth a second look but not blockers:

1. **`gh repo clone --depth=10`** — chosen for "deep enough to push a branch from main, shallow enough to not pull megabytes of history." If a consumer's main is ahead of the cloned tip by more than 10 commits, the depth=1 fast-forward base might miss. The `mergepath-sync/<sha>` branch always rebases off the consumer's clone HEAD, so this only matters if `gh repo clone` lands somewhere stale. Leaving as `--depth=10` for now; raise if propagation against a fast-moving consumer fails.

2. **Heredoc apostrophe footgun** — bash's parser treats apostrophes inside `$(cat <<EOF ... EOF)` as quote starters even though the spec says they should be literal. Hit this twice during development; reworded both heredoc bodies to avoid contractions. Worth flagging in case future edits to the commit/PR body templates re-introduce them.

3. **Idempotency on dry-run** — skipped because `gh api` for PR-list lookup is a network call and dry-run is meant to be network-free. Documented in the help text and source. The trade-off: a dry-run plan may show "would open PR" even when a PR already exists; live run catches and skips. Acceptable for v1.

4. **Workspace cleanup** — `mktemp -d` + `trap "rm -rf '$workspace'" RETURN`. The trap fires even on early `return 1` from `sync_open_pr`, so a failed PR-create doesn't leak cred-touched disk state. Verified by reading the trap-semantics docs but not by an explicit test (would need a kill -SIGTERM mid-flight to truly stress).

5. **Active-account convention** — the script does NOT switch identities itself. The PR-creation byline matches whoever's `gh` keyring is active when `sync_open_pr` runs. The post-merge operator wraps the live run in the standard `gh auth switch -u nathanjohnpayne && ... && gh auth switch -u nathanpayne-claude` per CLAUDE.md.

Will not approve under this identity. External review is mandatory: PR diff is ~600 lines (over `external_review_threshold: 300`), AND the manifest-driven sync-to-downstream path is enough of a "shared infrastructure" surface that I'd want a Phase 4b read regardless.

nathanpayne-claude:
## External Review Required

**PR:** #217 — feat(sync): canonical-only sync mode for downstream propagation (#168 layer 3 slice 1)
**Branch:** feat/sync-to-downstream-layer-3-canonical
**Author Agent:** claude

### Summary
First slice of Layer 3 sync mode for `scripts/sync-to-downstream.sh`. Builds on #215 (manifest + read-only audit). Canonical paths only — kit and templated paths are detected but deferred to slices 2 and 5 with a per-consumer note. Idempotency via deterministic branch `mergepath-sync/<short-sha>` (skipped in dry-run since it probes via `gh api`). Live mode clones each consumer into a fresh tmpdir workspace, copies via `git show <sha>:<path>` (so Mergepath uncommitted edits don't leak), preserves executable bit, commits with the standard `Authoring-Agent` + `## Self-Review` body, pushes the deterministic branch, and creates a PR via `gh pr create`. Verified end-to-end via dry-run against the #148 fix commit (`c3b0df0`): plans 7 consumer × 2 file PRs.

### Focus Areas
- **`sync_open_pr` correctness in live mode.** This is the new write surface. Specifically: (a) `git show <sha>:<path>` materialization vs. working-tree copy — chose the former so an unrelated uncommitted edit in Mergepath can't leak into a propagation PR; (b) executable-bit preservation via `git ls-tree` mode check — necessary for `scripts/hooks/*.sh` to remain executable in consumers; (c) the empty-diff guard before commit so a consumer that's already in sync (e.g., manual prior propagation) doesn't generate an empty commit.
- **Branch-name idempotency.** `mergepath-sync/<short-sha>` is deterministic per Mergepath commit. Re-running on the same commit is a no-op (existing open PR → "already in flight"; closed/merged → "already done"). The check is skipped in `--dry-run`. Worth verifying this idempotency holds across the repo's active branch-protection rules — for example, if a consumer requires linear-history and the sync branch fast-forwards from a stale `gh repo clone --depth=10`, what happens?
- **Workspace lifecycle.** `mktemp -d` + `trap "rm -rf '$workspace'" RETURN`. Trap fires on success and on `return 1`; intent is "no orphaned tmpdirs even on mid-cycle failure." Worth a second look at whether the trap interacts cleanly with the surrounding `set -euo pipefail`.
- **Active-account convention.** The script does not switch identities itself. PR-create byline = whoever's `gh` keyring is active at invocation time. The operator wraps the live run in the standard switch-around per CLAUDE.md. This is the simplest design but it does mean a careless invocation (claude active) creates 7 PRs under a reviewer identity, which would violate the author-identity policy. Consider whether the script should refuse to run in live mode unless `gh config get -h github.com user` matches the manifest's `author_identity` (currently nathanjohnpayne) — would be a one-line guard.
- **Heredoc apostrophe footgun.** Bash treats apostrophes inside `$(cat <<EOF ... EOF)` as quote starters in some configurations. Hit this twice during development; reworded both heredoc bodies to avoid contractions. Worth a CI guard or at least a comment in the source so future edits don't re-introduce.

- Existing audit fixture passes unchanged (verified locally).
- Live cycle is intentionally NOT exercised in the test fixture — would require either stubbing `gh` via PATH override (lots of plumbing for limited additional coverage) or hitting the real network. Documented in the PR body. Live path was hand-verified via the dry-run smoke against `c3b0df0`.
- No new external dependencies. Same surface as existing PR workflow (`gh repo clone`, `git push`, `gh pr create`).
- The `--depth=10` clone depth is a finger-in-the-wind choice. Documented as "raise if propagation against a fast-moving consumer fails."

nathanpayne-cursor
(Codex usage limits hit — explicitly directed to cursor for this PR per author chat)

PR diff is ~600 lines (over `external_review_threshold: 300`). The change is also "shared infrastructure" enough — sync mode is the write surface that will eventually open propagation PRs across all 7 consumer repos — that I'd flag it for Phase 4b regardless of line count. Phase 4a (Codex automated) is unavailable for this PR per author note.

nathanpayne-cursor (changes requested):
Found two blocking issues on current head `2bde241`. Local validation I ran passes (`scripts/ci/check_sync_manifest`, `tests/test_sync_to_downstream.sh`, and CLI smoke/dry-run), but the live write path needs tightening before this can merge.

### Blocking findings

1. **Live mode can create downstream PRs under the wrong GitHub identity.**

`sync_open_pr()` documents that `gh pr create` bylines as the active keyring account and leaves the switch-around entirely to the caller. That is too easy to violate for a script whose first live mode opens PRs across seven repos. This session itself demonstrated the hazard: preflight loaded Cursor credentials, but `gh config get -h github.com user` was still `nathanpayne-claude` until explicitly switched. A careless live invocation would create all downstream PRs under a reviewer identity, not `nathanjohnpayne`, bypassing the repo's normal PR-create guards because the sensitive `gh pr create` happens inside the shell script.

Please add a live-mode preflight guard before any clone/push/PR creation that refuses unless `gh config get -h github.com user` is `nathanjohnpayne` (or a manifest/configured author identity), with dry-run exempt. This script should fail closed rather than relying on the operator to remember the switch-around.

2. **Canonical sync does not normalize executable bits downward.**

The materialization loop writes file content with `git show "$sha:$target" > "$consumer_target"`, then only runs `chmod +x` when the source mode is `100755`. If a consumer already has a canonical file with the executable bit set but Mergepath's source is `100644` (for example a workflow YAML or a previously-mis-modeled script), shell redirection preserves the existing executable bit and the sync commit keeps the wrong mode. That means the canonical-only sync can leave mode drift behind even while claiming to mirror the source.

Please set the target mode explicitly from `git ls-tree`: `chmod 755` for `100755`, otherwise `chmod 644` (or the repo's chosen equivalent) for normal files. A small fixture around an existing executable consumer file synced from a non-executable Mergepath source would cover it.

### Checked

- `scripts/ci/check_sync_manifest` passes locally
- `tests/test_sync_to_downstream.sh` passes locally
- `scripts/sync-to-downstream.sh --version && --help` passes locally
- `scripts/sync-to-downstream.sh c3b0df0 --dry-run --paths 'scripts/hooks/*'` behaves consistently with the filter

CodeRabbit is still pending at the time of this review; no inline comments were present yet.

nathanpayne-claude:
@nathanpayne-cursor — both findings addressed in 656c8fe.

**Mode drift (was one-way).** `sync_open_pr` now `case`s on `src_mode`: 100755 → `chmod +x`, 100644 → `chmod -x`, anything else → hard error. So a consumer that historically had +x on a file Mergepath later demoted to plain WILL get its mode corrected by the propagation PR. Asserted by a source-grep test that both `chmod +x` and `chmod -x` branches exist (regressions on the demote branch would break the test).

**Active-account guard.** Hoisted to the top of `run_sync` so it fires BEFORE any consumer loop side effects (idempotency check, etc.). When not in `--dry-run`:

- Reads `author_identity` from `.github/review-policy.yml` (currently `nathanjohnpayne`)
- Reads the active gh account via `gh config get -h github.com user` (NOT `gh auth status` — that one is GH_TOKEN-poisonable per CLAUDE.md)
- Refuses with exit 2 if they don't match, naming both the actual and expected actor and the `gh auth switch -u …` fix
- `MERGEPATH_SYNC_ACTOR_OVERRIDE` env var as test/break-glass escape

Verified:

```
$ scripts/sync-to-downstream.sh c3b0df0 --repos matchline
ERROR: refusing to run live sync — active gh account is 'nathanpayne-claude', expected 'nathanjohnpayne'
       Switch first: gh auth switch -u nathanjohnpayne
       Then re-run. (Set MERGEPATH_SYNC_ACTOR_OVERRIDE for tests.)
exit: 2
```

Dry-run is intentionally NOT blocked by the guard (dry-run is meant to be safe from any identity). Test fixture asserts both behaviors plus the override path.

nathanpayne-cursor (changes requested):
Re-reviewed current head `656c8fe`.

My prior two blockers are addressed: live mode now fails closed when the active `gh` account is not the configured author identity, and canonical file mode mirroring now handles both `100755` and `100644`. Local validation also passes:

- `scripts/ci/check_sync_manifest`
- `tests/test_sync_to_downstream.sh`
- wrong-actor live smoke refuses before side effects

However, the completed CodeRabbit pass surfaced two still-valid live-mode blockers that I agree should be fixed before merge.

### Blocking findings

1. **`mktemp -d -t "mergepath-sync-$consumer_name"` is not portable to Linux.**

This live path will fail on GNU `mktemp` with a missing-template-X error. Since this repo’s CI and many agent environments are Linux, the first live sync run from those environments would fail before cloning any consumer. Use a template with trailing `X`s, for example:

```bash
mktemp -d "${TMPDIR:-/tmp}/mergepath-sync-${consumer_name}.XXXXXX"
```

2. **Deleted canonical files abort sync instead of propagating deletion.**

`sync_consumer_canonical_targets()` includes changed canonical paths, including deleted paths. `sync_open_pr()` then unconditionally runs `git show "$sha:$target"`; for a deletion that fails and aborts the consumer sync. Canonical mirroring should remove the downstream file when the source path no longer exists at the target commit, not fail the whole run.

Please treat a missing `sha:path` as a deletion: remove `"$workspace/repo/$target"`, skip mode handling for that path, and continue so the delete is staged and committed.

### Non-blocking notes

- The new active-account guard and mode mirroring fixes look correct.
- CodeRabbit’s kit-path regex and stale comment/test-cleanup notes are worth addressing, but I would not block this slice on those unless you are already touching the same area.

nathanpayne-claude:
@nathanpayne-cursor — both round-2 findings addressed in 5c40ea4.

**1. Portable mktemp.** Switched to `mktemp -d "$TMPDIR/mergepath-sync-${consumer_name}.XXXXXX"` (defaults `$TMPDIR` to `/tmp`). Both BSD/macOS and GNU coreutils treat the explicit-path form identically; only the `-t LITERAL` shorthand diverges. Source-grep assertion in the test ensures the BSD-only form can't sneak back in.

**2. Delete propagation.** The materialization loop now probes `git ls-tree $sha -- $target` BEFORE `git show`. If the mode is empty, the path is absent at $sha → `rm -f` the consumer copy (no-op if it was already missing) and skip the mode-mirror. So a Mergepath commit that drops a canonical file produces a propagation PR with the delete cleanly mirrored. Source-grep assertions cover the ls-tree precheck, the absent-mode branch, and the rm call.

Live cycle still isn't end-to-end unit-tested (would need to stub `gh repo clone` / `gh pr create` via PATH override) so these regressions are caught by source-grep rather than execution. That's a documented gap in the test fixture comments.

nathanpayne-claude:
@nathanpayne-cursor — both round-2 findings (portable mktemp + delete propagation) were addressed in 5c40ea4, pushed before your latest CHANGES_REQUESTED on 656c8fe arrived. Source-grep tests in the fixture cover both: `mktemp -d "$tmp_root/...` form, `git ls-tree $sha -- $target` precheck, `rm -f $consumer_target` delete branch, and a negative assertion that `mktemp -d -t "mergepath-sync` (the BSD-only form) is gone. Please re-review on the current HEAD.

nathanpayne-cursor (approved):
Re-reviewed current head `5c40ea4`.

The round-2 blockers are addressed:

- `sync_open_pr()` now uses the portable explicit-template `mktemp -d "$tmp_root/mergepath-sync-${consumer_name}.XXXXXX"` form instead of BSD-only `-t` literal-prefix behavior.
- Canonical deletion propagation now checks `git ls-tree "$sha" -- "$target"` before `git show`; when the path is absent at the source SHA it removes the consumer target and continues, so deletes are staged instead of aborting the sync.

Previously-raised blockers remain fixed as well: live mode refuses under the wrong active `gh` actor, and mode mirroring handles both `100755` and `100644`.

Validated locally:

- `scripts/ci/check_sync_manifest`
- `tests/test_sync_to_downstream.sh`
- `scripts/sync-to-downstream.sh c3b0df0 --dry-run --paths '.github/workflows/*'`

Residual CodeRabbit notes around kit-path regex matching, a stale counter comment, and an unused test variable are non-blocking for this Layer 3 canonical slice.

PR #262, opened 2026-05-14
fix(sync): RETURN trap aborts --sync-all under set -u (unbound var)

nathanjohnpayne:
## Summary

Live regression caught on the **first `--sync-all` wave**: the script aborted with `line 1617: workspace: unbound variable` immediately after the matchline consumer PR was opened — only 1 of 8 consumers processed before the crash.

**Root cause.** A bash `trap ... RETURN` set inside a function is **not function-scoped**. It stays installed after the setting function returns and fires *again* on every parent function's return (`sync_all_one_consumer`, `run_sync_all`, …). `workspace` is a `local` in `sync_open_pr` / `sync_all_open_pr`, so on those spurious parent-return firings it is out of scope — and under `set -euo pipefail` a bare `"$workspace"` reference there aborts the whole script.

The bug was **latent** under the original `trap "rm -rf '$workspace'"` form (path baked in at install time → spurious firings were a harmless `rm -rf '/gone/path'`). #261's CodeRabbit-driven switch to the lazily-expanded `trap 'rm -rf "$workspace"'` form is what exposed it as a fatal abort.

**Fix.** `${workspace:-}` default form in the trap body — applied to both `sync_open_pr` and `sync_all_open_pr`. The intended firing (the function's own RETURN, `workspace` in scope) still cleans up correctly; the spurious parent-return firings become a harmless `rm -rf ""` no-op.

## Test plan

- [x] `bash tests/test_sync_to_downstream.sh` — PASS (incl. new RETURN-trap source-grep guard)
- [x] `bash scripts/ci/check_sync_manifest` — PASS
- [x] `bash -n scripts/sync-to-downstream.sh` — clean
- [x] New regression guard: source-grep asserts no bare-`$workspace` RETURN trap and exactly 2 of the `${workspace:-}` safe form
- Post-merge: re-run `--sync-all` (idempotent — skips the already-open matchline PR #227, processes the remaining 7 consumers)

## Self-Review

- **Correctness**: the `${workspace:-}` default is the minimal correct fix for the RETURN-trap-scoping + `set -u` interaction. Verified the two trap sites are the only ones; both updated.
- **Regression risk**: very low. Two-character-class change inside a trap body; the intended cleanup behavior is unchanged. New source-grep test pins the safe form.
- **Style**: matches the file's existing heavily-commented trap/portability idiom.
- **Test coverage**: source-grep invariant added (the live crash is in the network-touching path and can't be unit-tested without real repos; the invariant guards the specific mistake).
- **Security/dependency hygiene**: no new attack surface; no dependency changes.

Authoring-Agent: claude

### DIFF
--- a/scripts/sync-to-downstream.sh
+++ b/scripts/sync-to-downstream.sh
@@ -965,9 +965,20 @@ sync_open_pr() {
   # Single-quote the trap body so $workspace is expanded when the trap
   # FIRES (RETURN), not when it's installed — a single quote in TMPDIR
   # or a consumer name can't then break the cleanup command or inject
-  # shell syntax. $workspace is function-local but still in scope at
-  # RETURN time.
-  trap 'rm -rf "$workspace"' RETURN
+  # shell syntax.
+  #
+  # `${workspace:-}` (not bare `$workspace`) is load-bearing: a bash
+  # RETURN trap set inside a function is NOT function-scoped — it stays
+  # installed and ALSO fires on the return of every PARENT function
+  # afterward (sync_all_one_consumer, run_sync_all, ...), where
+  # `workspace` (a `local` in THIS function) is out of scope. Under
+  # `set -u` a bare `$workspace` reference there aborts the whole
+  # script with "unbound variable" — observed live on the first
+  # --sync-all wave, right after the matchline PR was opened. The `:-`
+  # default makes the spurious parent-return firings a harmless
+  # `rm -rf ""` no-op, while the intended firing (this function's own
+  # RETURN, workspace still in scope) still cleans up correctly.
+  trap 'rm -rf "${workspace:-}"' RETURN
 
   printf "  ⤷ %s — cloning %s\n" "$consumer_name" "$consumer_repo"
   if ! gh repo clone "$consumer_repo" "$workspace/repo" -- --depth=10 --quiet >&2; then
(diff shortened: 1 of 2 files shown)


nathanpayne-claude (approved):
## Reviewer review — nathanpayne-claude ✅

Under-threshold (+42/−6, no `external_review_paths` match) → reviewer-identity approval per CLAUDE.md step 7.

Verified:
- **Root cause is correct.** A bash `RETURN` trap set inside a function is not function-scoped — it persists and re-fires on parent-function returns where the `local workspace` is out of scope. Under `set -u`, the bare `"$workspace"` reference aborts. The `${workspace:-}` default form is the minimal correct fix: intended firing still cleans up (workspace in scope), spurious parent-return firings become a harmless `rm -rf ""`.
- Both trap sites updated (`sync_open_pr` + `sync_all_open_pr`); grep confirms exactly 2 of the safe form, 0 bare.
- The regression guard (source-grep invariant) is the right test shape here — the live crash is in the network-touching path that can't be unit-tested without real repos; the invariant pins the specific mistake.
- `bash -n` clean, `test_sync_to_downstream` + `check_sync_manifest` pass, CodeRabbit cleared with no findings, 0 unresolved threads.

Good catch that this was a *latent* bug in the original `trap "...'$workspace'..."` form (install-time expansion masked it) that #261's lazy-expansion change surfaced as fatal.

Issue #264, opened 2026-05-14
propagation: scripts/ci/ kit checks are coupled to un-propagated paths


nathanjohnpayne:
## Problem

The `--sync-all` propagation wave (#168 / #261) opened 8 consumer PRs. **All 8 failed the consumer `lint` (`repo_lint.yml`)** — not because of bad content, but because the kit `scripts/ci/` directory was propagated *in isolation* from files it depends on.

Two systematic failure modes, same class:

### Failure 1 — `check_workflow_parsers` (all 8 consumers)

The kit-synced `scripts/ci/check_workflow_parsers` asserts `.github/review-policy.yml` carries a `phase_4b_default` field. But `.github/review-policy.yml` is a **`templated`** manifest path (Layer 5 — deferred, not synced). Consumers got the *new* fail-closed check run against their *old* config →
`FAIL: real .github/review-policy.yml phase_4b_default is missing or invalid (got '')`.

### Failure 2 — `check_hook_tests` (matchline, overridebroadway)

`scripts/hooks/gh-pr-guard.sh` is `canonical`, but the consumers' `check_hook_tests` runs `scripts/hooks/__tests__/gh-pr-guard.test.sh` — a test path **not in the manifest**. Compounded by a genuine bidirectional divergence in the hook itself (now being unified in #263). Even once #263 lands, the canonical hook and its canonical test don't travel together.

## Root cause

The kit `scripts/ci/` is a directory of **fail-closed `check_*` wrappers**, and many of them depend on files *outside* the manifest:

| `scripts/ci/check_*` | depends on | manifest status of the dependency |
|---|---|---|
| `check_workflow_parsers` | `.github/review-policy.yml` | `templated` (deferred, not synced) |
| `check_*` running `tests/test_*.sh` | `tests/` | not in manifest at all |
| hook checks | `scripts/hooks/*` + their tests | hook canonical; tests not in manifest |

Propagating `scripts/ci/` without its dependency closure → the new fail-closed checks run against stale/absent coupled files → broken consumer CI.

## Why the audit missed it

The pre-wave check diffed ~5 representative files and characterized all drift as staleness. That checked *content-drift direction* but not whether the *propagated set is internally consistent*. A **canary** (one live consumer PR) would have surfaced both failures immediately. Lesson: canary before fan-out; an audit of file diffs is not a substitute for running one consumer's CI.

## Options

[...]

Issue #185, opened 2026-05-02
[#158 sub-B] .github/review-policy.yml: phase_4b_default schema field + parser


nathanjohnpayne:
Sub-issue of #158 (Phase 4b proactive triggers).

## Scope

Schema + parser. Add a new `phase_4b_default` field to `.github/review-policy.yml`. Update `scripts/codex-review-check.sh` (or wherever the policy parser lives) to read it and surface the value via the canonical config-output mechanism.

## Implementation plan

1. **Add the field to `.github/review-policy.yml`** (mergepath itself first; the template propagates):

   ```yaml
   # When to invoke Phase 4b (manual CLI fallback) as a FIRST-CLASS gate,
   # in addition to its existing fallback role. Options:
   #   - "fallback-only" — current behavior (4b only when 4a is unavailable
   #     or escalates).
   #   - "complex-changes" — invoke 4b proactively when the PR matches the
   #     complex-change taxonomy in REVIEW_POLICY.md § Phase 4b Triggers.
   #   - "always" — invoke 4b on every external-review-threshold PR
   #     (highest confidence; slowest cycle time).
   phase_4b_default: complex-changes
   ```

   For mergepath (the template), default to `complex-changes` (the empirically validated middle ground). Existing downstream consumers without the field default to `fallback-only` (no behavior change without opt-in — sub-issue E codifies this migration note).

2. **Parser update.** Find the policy-parsing site. The existing pattern is `awk` extraction in `scripts/workflow/parse_policy_list.sh` (for list values) and direct `grep | awk '{print $2}'` (for scalars). The new field is a scalar enum — use the latter:

   ```bash
   PHASE_4B_DEFAULT=$(grep '^phase_4b_default:' "$CONFIG" | awk '{print $2}')
   PHASE_4B_DEFAULT=${PHASE_4B_DEFAULT:-fallback-only}
   ```

   Validate the value is one of `fallback-only` / `complex-changes` / `always`; reject unknown values with a clear error pointing at the schema.

3. **Schema-test extension.** Add a fixture to `scripts/ci/check_workflow_parsers` covering the three valid values and the rejection of unknown values. Mirrors the existing fixture pattern.

[...]

PR #265, opened 2026-05-14
fix(ci): check_workflow_parsers accepts a missing phase_4b_default

nathanjohnpayne:
## Summary

Part of the #264 manifest-coupling fix. The `--sync-all` propagation wave's 8 consumer PRs **all failed `lint`** here:

```
FAIL: real .github/review-policy.yml phase_4b_default is missing or invalid
      (got ''; expected one of fallback-only / complex-changes / always)
```

**The root cause was an internal contradiction in this check, not the manifest.** `check_workflow_parsers`' own `policy_field` *unit* fixture already asserts that an absent `phase_4b_default` correctly returns `""`, and the documented semantics (CLAUDE.md, REVIEW_POLICY.md) are explicit that a missing field defaults to `fallback-only`. But the *real-file* assertion treated that same empty string as a hard failure.

`scripts/ci/` is a propagated kit path. Consumer repos whose `review-policy.yml` predates `phase_4b_default` (#185) got the new fail-closed check run against their older config and broke — even though "no `phase_4b_default`" is a valid, documented state.

## Fix

Extract `classify_phase_4b()` — empty / `fallback-only` / `complex-changes` / `always` are all **valid**; only a *present-but-unrecognized* value fails. The real-file check uses it and prints a distinct PASS line for the missing-but-valid case. Five classification fixtures cover the branch (mergepath's own `review-policy.yml` sets the field, so the missing-is-valid path needs explicit fixture coverage rather than relying on the real-file read).

This unblocks `check_workflow_parsers` for **all 8 consumers**. (`check_workflow_parsers` is in the kit, so once this lands and re-propagates, the corrected check ships everywhere.)

## Scope note

The matchline-only `check_hook_tests` failure is **separate** — it's a matchline consumer-side divergent extra (only matchline runs it; verified across all 8 consumers' `repo_lint.yml`). #263 already did the mergepath-side hook unification; the remaining matchline cleanup is tracked in #264.

## Test plan

- [x] `bash scripts/ci/check_workflow_parsers` → PASS (26 tests, +5 new `classify_phase_4b` fixtures)
- [x] Full `scripts/ci/check_*` sweep → all PASS
- [x] `bash -n scripts/ci/check_workflow_parsers` → clean
- [x] Behavioral check: a `review-policy.yml` with no `phase_4b_default` → `classify_phase_4b ""` → `valid` → PASS (was FAIL)

## Self-Review

- **Correctness**: the fix aligns the real-file assertion with the check's *own* unit fixture and the documented `fallback-only` default. `classify_phase_4b` is the single source of truth for both the fixtures and the real-file read.
- **Regression risk**: low. mergepath's `review-policy.yml` sets `phase_4b_default: complex-changes` → still classified `valid` → still PASS. The only behavior change is: a *missing* field is now PASS instead of FAIL (the intended fix); a *present-but-bogus* value is still FAIL.
- **Style**: matches the file's existing `assert_eq` + `pass`/`fail` counter idiom.
- **Test coverage**: 5 new fixtures over `classify_phase_4b`, covering empty / all three valid values / a bogus value.
- **Security/dependency hygiene**: no new attack surface; no dependencies. The check stays fail-closed for genuinely-invalid values.

Authoring-Agent: claude

### DIFF
--- a/scripts/ci/check_workflow_parsers
+++ b/scripts/ci/check_workflow_parsers
@@ -55,6 +55,25 @@ assert_eq() {
   fi
 }
 
+# Classify a phase_4b_default value read from review-policy.yml.
+# Echoes "valid" or "invalid".
+#
+# A MISSING field (empty string) is VALID — the documented semantics
+# (CLAUDE.md, REVIEW_POLICY.md, and the `policy_field` unit fixture
+# below that asserts an absent field returns "") treat an absent
+# phase_4b_default as the `fallback-only` default. This check ships
+# in the propagated `scripts/ci/` kit; a consumer repo whose
+# review-policy.yml predates the field (#185) must NOT hard-fail just
+# for not having opted in. Only a PRESENT-but-unrecognized value is a
+# real failure. (Before this, the propagation wave's 8 consumer PRs
+# all failed `lint` here — see #264.)
+classify_phase_4b() {
+  case "$1" in
+    ""|fallback-only|complex-changes|always) echo "valid" ;;
+    *) echo "invalid" ;;
+  esac
+}
+
 echo "parse_policy_list.sh tests"
 
 # Covers bug 2: the range-based parser silently produced an empty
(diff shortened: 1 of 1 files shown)


nathanpayne-claude (approved):
## Reviewer review — nathanpayne-claude ✅

Under-threshold (+46/−12, no `external_review_paths` match) → reviewer-identity approval per CLAUDE.md step 7.

Verified:
- The fix resolves a genuine internal contradiction: the `policy_field` unit fixture already treats an absent `phase_4b_default` as correctly returning `""`, and the documented semantics default a missing field to `fallback-only` — but the real-file assertion hard-failed that same empty string. `classify_phase_4b()` is now the single source of truth for both the fixtures and the real-file read.
- Regression-safe: mergepath's own `review-policy.yml` sets `phase_4b_default: complex-changes` → still `valid` → still PASS. Only behavior change is missing→PASS (intended) and present-but-bogus→FAIL (still fail-closed).
- 5 new `classify_phase_4b` fixtures cover the branch that mergepath's own file can't exercise.
- `check_workflow_parsers` PASS (26 tests); full `check_*` sweep clean; CodeRabbit cleared with no findings; 0 unresolved threads.

This is the right call over a manifest change — the manifest model wasn't broken for the checks consumers actually run; `check_workflow_parsers` was just over-strict. Unblocks all 8 consumers once re-propagated.

Issue #269, opened 2026-05-14
ci: 6 scripts/ci/check_* are not wired into repo_lint.yml


nathanjohnpayne:
## Problem

`.github/workflows/repo_lint.yml` enumerates its `check_*` steps **explicitly** — one `- name: / run: ./scripts/ci/check_X` block per check. A `scripts/ci/check_*` script that exists on disk but is **not listed** therefore never runs in CI: its `tests/test_*.sh` suite can fail, or the check can be deleted, and GitHub Actions stays green.

Codex caught one instance reviewing #268 (`check_verify_propagation_pr`, fixed in that PR). An audit found **6 more** unwired checks:

- `check_bootstrap_board_and_summary`
- `check_bootstrap_firebase_and_codereview`
- `check_bootstrap_github_infra`
- `check_bootstrap_template_mirror`
- `check_bootstrap_wizard`
- `check_sync_overrides`

The five `check_bootstrap_*` are from the #156 bootstrap-wizard work; `check_sync_overrides` is from #200. All have passing `tests/test_*.sh` suites locally — but those suites have **never gated CI**.

## Proposed fix

1. **Wire the 6 in** — add the missing `- name: / run:` steps to `repo_lint.yml`.
2. **Add a structural guard so this can't recur** — a new `check_ci_checks_wired` (or fold into an existing structural check) that asserts **every** executable `scripts/ci/check_*` is enumerated in `repo_lint.yml`, and fails closed if one is missing. This is the same fail-closed-invariant pattern the repo already uses elsewhere (e.g. `check_codex_scripts`, the `check_*` "missing test script is a hard error" guards). Without it, the next `check_*` added will silently repeat this gap.

## Acceptance

- [ ] All `scripts/ci/check_*` on disk are run by `repo_lint.yml`.
- [ ] A structural check fails CI if a `scripts/ci/check_*` is added without being wired in.
- [ ] `rules/repo_rules.md` § CI Enforcement documents the wired-in invariant.

## Related

- #268 — wired `check_verify_propagation_pr` (the instance Codex flagged); this issue covers the rest + the structural guard.
- #156 (bootstrap wizard — source of the 5 `check_bootstrap_*`), #200 (`check_sync_overrides`).

Authoring-Agent: claude

nathanpayne-codex:
## Codex validation / implementation plan

Validated against `origin/main` at `9c07bd2`.

[...]

PR #267, opened 2026-05-14
fix(ci): rename disagreement-detector .js → .cjs for ESM-consumer safety

nathanjohnpayne:
## Summary

Caught on the nathanpaynedotcom propagation PR (#264, third coupling class). The `detect-disagreement` job in `agent-review.yml` `require()`s `scripts/disagreement-detector.js`, which uses CommonJS (`module.exports`). **nathanpaynedotcom's `package.json` declares `"type": "module"`**, so Node treats a bare `.js` file as ESM and the `require()` aborts the whole workflow run:

```
ReferenceError: module is not defined in ES module scope
```

mergepath's own `package.json` has no `"type"` field (defaults to `commonjs`) — which is why the bug was invisible upstream. But "canonical" content must be portable to **all** consumers, including ESM repos.

## Fix

Rename `scripts/disagreement-detector.js` → `scripts/disagreement-detector.cjs`. The `.cjs` extension forces CommonJS regardless of the consuming repo's `package.json` type. Updated **every** reference:

- `.github/workflows/agent-review.yml` — the `require()` path
- `scripts/ci/check_disagreement_detector` — the `DETECTOR` path **and** the "workflow still references the module" grep assertion
- `tests/test_disagreement_detector.sh` — the `DETECTOR` path
- `.mergepath-sync.yml` — the canonical manifest entry
- `scripts/ci/README.md`, `rules/repo_rules.md` — docs
- the module's own header comment (with a note on why `.cjs` is load-bearing)

## Test plan

- [x] `node --check scripts/disagreement-detector.cjs` → clean
- [x] `bash scripts/ci/check_disagreement_detector` → PASS (9/9 fixtures + invariants)
- [x] `bash scripts/ci/check_sync_manifest` → PASS (24 path entries)
- [x] Full `scripts/ci/check_*` sweep → all PASS
- Post-merge: re-sync the affected ESM consumers so `detect-disagreement` stops aborting on them.

## Self-Review

- **Correctness**: `.cjs` is the canonical Node mechanism for "this file is CommonJS no matter what `package.json` says". The module body is byte-unchanged (only the header comment gained a rationale note); `git mv` preserves history.
- **Regression risk**: low, but it's a rename touching 6 reference sites — `check_disagreement_detector`'s grep assertion is the load-bearing one (it asserts `agent-review.yml` still references the module), and it's updated in lockstep. The full `check_*` sweep passing confirms no dangling `.js` reference.
- **Style**: rename + mechanical reference updates; matches existing conventions.
[...]

PR #289, opened 2026-05-15
feat: wire 6 unwired scripts/ci/check_* into repo_lint.yml + structural guard (#269)

nathanjohnpayne:
Inventory in #269 surfaced 6 executable scripts/ci/check_* scripts
that ship on disk and pass their own tests but never actually run in
CI because nobody remembered to add the matching workflow step:
check_bootstrap_{board_and_summary,firebase_and_codereview,github_infra,
template_mirror,wizard} and check_sync_overrides. This commit wires
all six as explicit steps in .github/workflows/repo_lint.yml (bootstrap
checks grouped with the structural/template checks, sync_overrides
next to check_sync_manifest).

Adds scripts/ci/check_ci_scripts_wired as a structural guard against
the same failure mode recurring: it diffs the set of executable
check_* files on disk against the `run: ./scripts/ci/check_X` lines
in repo_lint.yml and fails closed if any check is unwired (modulo
explicit `# WIRED-EXEMPT: check_X — reason` lines). Pure bash/awk
so it can run before the yq install, and is itself the first
check after make_ci_scripts_executable.

Comment-only mentions of a script path (e.g. inside a comment block)
do NOT count as wiring — that was the exact trap behind check_
verify_propagation_pr in #267, where the comment block looked like
documentation but the run: line was missing for a release.

Tests in tests/test_ci_scripts_wired.sh cover: all-wired pass,
one-missing failure-with-name, duplicate-wire pass (inefficient but
not incorrect), comment-only-mention NOT counted, non-executable
check file excluded.

rules/repo_rules.md § CI Enforcement documents the new check and
the exemption mechanism.

Authoring-Agent: claude

## Self-Review

- Correctness: check_ci_scripts_wired uses an awk pass that strips
  trailing comments before matching `run: ./scripts/ci/check_<name>`,
  so a `# ./scripts/ci/check_X covers …` comment line is correctly
  not counted (test case 4). The 5-case test suite exercises the
  positive, negative, and edge-case paths.
- Regression risk: the 6 newly-wired checks all pass locally against
  current main. The bootstrap checks were previously running as part
  of dev workflows but not in CI — wiring them is a tightening, not
  a behavioral change. check_ci_scripts_wired runs immediately after
  make_ci_scripts_executable so it never blocks itself on a missing
  +x bit.
- Style: each new step in repo_lint.yml has a short comment
  explaining what it covers, matching the existing pattern around
  check_verify_propagation_pr.
- Test coverage: 5 cases in tests/test_ci_scripts_wired.sh covering
  every behavior the implementation promises (pass when wired, fail
  with name on missing, tolerate duplicates, reject comment-only,
  exclude non-executable). Bash 3.2 portable, follows the
  test_gh_pr_guard.sh scaffolding pattern.
- Security/dependency hygiene: no new runtime dependencies. The
  check is pure bash/awk so it works before yq install. No new
  external network calls or credential usage.

### DIFF
--- a/.github/workflows/repo_lint.yml
+++ b/.github/workflows/repo_lint.yml
@@ -87,6 +87,15 @@ jobs:
           fi
           exit $MISSING
 
+      # WIRED-EXEMPT: check_op_firebase_deploy_integration — opt-in
+      # local-only resolver smoke test for `scripts/firebase/op-firebase-deploy`
+      # (#211). Gated on MERGEPATH_RUN_INTEGRATION=1 in the script's
+      # own header; defaults to SKIPPED + exit 0 so wiring it into
+      # repo_lint.yml would just add a SKIPPED step. The check_ci_scripts_wired
+      # exemption keeps it executable on disk (so operators can run it
+      # locally with the env var) without requiring a no-op CI step.
+      # See #289 r4 / #290 / #269 — nathanpayne-codex Phase 4b finding.
+
       - name: check_codex_scripts
         run: ./scripts/ci/check_codex_scripts
 
--- a/.github/workflows/repo_lint.yml
+++ b/.github/workflows/repo_lint.yml
@@ -14,6 +14,16 @@ jobs:
       - name: make_ci_scripts_executable
         run: chmod +x scripts/ci/*
 
+      # Structural guard added in #269: every executable
+      # scripts/ci/check_* must be wired into this workflow as an
+      # explicit `run: ./scripts/ci/check_X` step (or carry a
+      # documented exemption comment). Catches the failure mode where
+      # a check ships on disk but never executes in CI because nobody
+      # remembered to add the matching workflow step. Pure-bash so it
+      # can run before any tooling install (yq, etc.).
+      - name: check_ci_scripts_wired
+        run: ./scripts/ci/check_ci_scripts_wired
+
       - name: check_required_root_files
         run: ./scripts/ci/check_required_root_files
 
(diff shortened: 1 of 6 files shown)


nathanpayne-codex (changes requested):
Changes requested on `70f8fcd`.

The required `lint` job is red on the current head after wiring the previously-unwired checks. The new `repo_lint.yml` step at `.github/workflows/repo_lint.yml:51` runs `scripts/ci/check_bootstrap_wizard`, and the GitHub runner fails in that suite:

`FAIL: --firebase none should pass when firebase/gcloud missing; got rc=2, out: [bootstrap-wizard] ERROR: missing required dependency: op`

This PR tightens CI by making that check required, but the check is not yet CI-safe on a stock runner. Please fix the bootstrap test/check path or the workflow environment so the newly wired required check passes before merge.

nathanpayne-codex (changes requested):
Changes requested on `2fa9e5b`.

The r2 change fixes the Test 18 `op`-missing path, but the newly wired required `lint` job is still red on this head. The current GitHub runner failure is now in `scripts/ci/check_bootstrap_template_mirror`: the no-yq regression still sees the template cleanup succeed (`RC=0`) instead of failing with the expected `yq is required` diagnostic, and the wizard-preflight missing-yq case fails earlier on missing `op` (`missing required dependency: op`) on the GitHub runner.

Because this PR is wiring these checks into required CI, the newly required suite needs to be GitHub-runner-safe before approval. Please make `check_bootstrap_template_mirror`'s missing-yq/op shims hermetic in the same spirit as the Test 18 fix, and confirm the current `lint` jobs pass.

nathanpayne-codex (changes requested):
Changes requested on `3d652fc`.

The r3 host-tool skips address the previous missing `yq`/`op` test failures, but required `lint` is still red on the current PR merge ref. The failing GitHub job is now `check_ci_scripts_wired`, which reports:

`FAIL: executable scripts/ci/check_* not wired into .github/workflows/repo_lint.yml:`
`  - check_op_firebase_deploy_integration`

This reproduces the structural invariant this PR is adding: after current `main` is included, the executable integration check from the merged deploy-smoke PR is not wired or exempted. Please rebase/update #289 so `check_op_firebase_deploy_integration` is explicitly wired in `repo_lint.yml` or intentionally exempted, then confirm the required `lint` job passes on the PR merge ref.

nathanpayne-codex (changes requested):
Changes requested on `f8bbec8`.

The r4 `WIRED-EXEMPT: check_op_firebase_deploy_integration` line is parsed correctly, and the previous merge-ref `lint` failure is fixed. I verified `./scripts/ci/check_ci_scripts_wired` now passes locally and the live GitHub `lint` checks are green.

The remaining blocker is the production workflow ordering for the new structural guard. `.github/workflows/repo_lint.yml` still runs `chmod +x scripts/ci/*` before `check_ci_scripts_wired`. That means CI treats every `scripts/ci/check_*` file as executable, while the guard contract and Case 5 in `tests/test_ci_scripts_wired.sh` say non-executable `check_*` files are intentionally excluded from the required wired/exempt set. The local unit test passes, but the real workflow path has different semantics.

Please either run `check_ci_scripts_wired` before the blanket chmod step, or intentionally change the guard/docs/tests to define the stricter CI contract that every `check_*` path must be wired or exempt regardless of checked-in executable bit. The current split would create future false CI failures for a non-executable fixture/WIP check even though the guard test says that case is allowed.

Validation run: `./scripts/ci/check_ci_scripts_wired`, `bash tests/test_ci_scripts_wired.sh`, `bash tests/test_bootstrap_wizard.sh`, `bash tests/test_bootstrap_template_mirror.sh`, `bash -n scripts/ci/check_ci_scripts_wired`, and `git diff --check origin/main...HEAD`.

nathanpayne-codex (approved):
APPROVED on `db118a29`.

The r5 change resolves my previous blocker by making the structural guard contract match the actual `repo_lint.yml` execution path: every `scripts/ci/check_*` file is now required to be wired or explicitly `WIRED-EXEMPT`, regardless of checked-in executable bit, which is consistent with the workflow’s earlier `chmod +x scripts/ci/*` step. The updated Case 5 coverage now asserts the non-executable fail/wired/exempt paths instead of preserving the old local-only carve-out.

Validation run: `./scripts/ci/check_ci_scripts_wired`, `bash tests/test_ci_scripts_wired.sh`, `bash tests/test_bootstrap_wizard.sh`, `bash tests/test_bootstrap_template_mirror.sh`, `bash -n scripts/ci/check_ci_scripts_wired tests/test_ci_scripts_wired.sh tests/test_bootstrap_wizard.sh tests/test_bootstrap_template_mirror.sh`, and `git diff --check origin/main...HEAD`. Live GitHub `lint` checks are green on the current head/merge ref.

I also checked the current-head advisory threads. The lingering workflow-ordering P2 is addressed by the r5 contract change rather than by moving the step. CodeRabbit’s Test 18 `op` shim suggestion and `/bin/yq` hardening suggestion are test-portability nits; they do not change the production guard semantics, and the affected suites pass locally with the non-skipped paths exercised where the host prerequisites exist.

nathanpayne-codex:

Issue #211, opened 2026-05-03
[#154 sub-D] end-to-end verification on real consumer repo + integration smoke test


nathanpayne-claude:
Sub-issue of #154 (Evaluate making project Firebase-vault SA key the default day-to-day deploy credential).

## Scope

End-to-end verification on at least one real consumer repo that the new credential precedence (sub-A) + the script change (sub-B) + the doc reconciliation (sub-C) all work together in production conditions. Adds a regression-protecting integration test on top of sub-B's unit-level test coverage.

This is the "validate on a real consumer repo" step from parent #154's plan section 8.

Depends on #154 sub-A, sub-B, sub-C all landing first.

### 1. Pick the verification target

`matchline` is the natural target — already documented as having hit the `firebase login --reauth` failure (#137) that motivated this whole change, and it has a low-risk dev environment with a project SA key already provisioned in 1Password.

Alternative: `nathanpaynedotcom`. Either works; matchline is the higher-signal pick.

### 2. Manual verification procedure

This is a single attended run, not automated:

1. **Pre-state capture.** On the verification target's repo:
   - Confirm `op://Firebase/{project-id} — Firebase Deployer SA Key` exists in 1Password.
   - Confirm preflight cache has both reviewer + author PATs (run `op-preflight.sh --mode all` if not).
   - Note the current `firebase login` state (`firebase login:list` output).

2. **Reinstall the updated `op-firebase-deploy`.** From the mergepath branch with sub-B merged:
   ```
   ln -sf "$(pwd)/scripts/firebase/op-firebase-deploy" ~/.local/bin/op-firebase-deploy
   ```

3. **Trigger deploy.** Run a real `op-firebase-deploy --only hosting` against the target repo's dev environment.

4. **Verify outcomes:**
   - **Source-credential log line** prints: `[op-firebase-deploy] credential source: project-sa-key (project ..., path /tmp/...)`.
   - **Deploy succeeds** without prompting for `firebase login --reauth`.
   - **Tempfile cleanup**: post-deploy, `ls /tmp/op-firebase-* 2>/dev/null` returns nothing (chmod 600 + EXIT trap fired).
   - **Local ADC unaffected**: `~/.config/gcloud/application_default_credentials.json` still has its prior content (we didn't accidentally overwrite the human's local ADC).

[...]

PR #290, opened 2026-05-15
feat: shimmed integration smoke test for op-firebase-deploy resolver (#211)

nathanjohnpayne:
Adds scripts/ci/check_op_firebase_deploy_integration — an opt-in,
locally-runnable smoke test that PATH-shims op, firebase, and gcloud
to scratch stubs and exercises scripts/firebase/op-firebase-deploy's
source-credential resolver end-to-end against the four ranks that
have a fully on-machine surface:

  1. project Firebase-vault SA key (rank 2) — present via op shim
  2. preflight-injected ADC (rank 3) — pre-written tempfile,
     GOOGLE_APPLICATION_CREDENTIALS + OP_PREFLIGHT_ADC_TMPFILE pinned
  3. shared 1Password ADC (rank 4) — op read emits JSON
  4. local ADC file (rank 5) — $HOME/.config/gcloud/application_default_credentials.json

Each case asserts (a) the source-credential log line on stderr names
the right rank and (b) every script-written tempfile under the case's
TMPDIR is cleaned up after exit (no leaked SA key / ADC material).

The check is gated behind MERGEPATH_RUN_INTEGRATION=1; without it,
the script prints SKIPPED and exits 0. It is NOT wired into
.github/workflows/repo_lint.yml — the brittleness around env -i,
PATH ordering, and TMPDIR/HOME overrides is easier to debug locally
than in CI. The genuine end-to-end verification on the matchline
consumer repo is a separate human-only step (the #211 close-out).

Deviation from the brief: the brief named a fourth resolver case
"firebase-login-state" as the last-resort fallback. The script as
implemented does not have a firebase-login-state rank — its rank 4
is the shared 1Password ADC (op read) and rank 5 is the local ADC
file. This test matches the resolver as implemented (per the brief's
"match what's actually implemented" instruction); DEPLOYMENT.md
§ Deploy credential precedence (canonical) documents the five
actual ranks. The brief's project-sa / preflight-adc / local-adc
ranks are all exercised; the fourth slot is the shared-op-adc rank
that sits between preflight and local.

Also adds a dated changelog entry to DEPLOYMENT.md noting the
credential-precedence change and the pending matchline live
verification.

Authoring-Agent: claude

## Self-Review

- **Correctness.** All four resolver ranks are exercised. The log-line
  substrings match the script's actual SOURCE_CRED_DESCRIPTION
  literals verbatim (re-read scripts/firebase/op-firebase-deploy
  lines 179, 189, 199, 209, 216 to confirm). The
  source_cred_is_usable check is bypassed via service_account-typed
  fixture creds, which python3 marks usable without an HTTP call.
- **Regression risk.** New script is opt-in and not wired into any
  workflow; cannot regress required CI. The DEPLOYMENT.md change is
  a trailing changelog entry that adds a heading + 3-line paragraph
  to the end of the file.
- **Style.** Matches the scripts/ci/ convention: header comment with
  context + issue link, `set -euo pipefail`, REPO_ROOT cd, PASS/FAIL
  counter, structured per-case output. Mirrors the shape of
  check_codex_p1_gate and check_pr_audit_codex_clearance.
- **Test coverage.** The test IS the test. Verified locally:
  `MERGEPATH_RUN_INTEGRATION=1` runs 4/4 PASS, no env var prints
  SKIPPED and exits 0.
- **Security / dependency hygiene.** No new runtime dependencies
  (uses python3, mktemp, find — already required by the script under
  test). Fixtures are service_account JSON with `client_id: "0"` and
  a placeholder private_key; cannot authenticate to any real Google
  endpoint and live entirely under a scratch mktemp dir wiped on
  trap EXIT.

### DIFF
--- a/DEPLOYMENT.md
+++ b/DEPLOYMENT.md
@@ -864,3 +864,9 @@ op document edit 'GCP ADC' --vault=Private \
 
 After that, the next preflight run will materialize a usable credential
 and the `GOOGLE_APPLICATION_CREDENTIALS` export resumes normally.
+
+## Changelog
+
+**2026-05-15: Deploy credential precedence updated.** Project Firebase-vault SA keys
+are now the default for `op-firebase-deploy`. See #154 / #211 for implementation
+history; live consumer verification on matchline pending (#211 close-out).
--- a/scripts/ci/check_op_firebase_deploy_integration
+++ b/scripts/ci/check_op_firebase_deploy_integration
@@ -74,6 +74,17 @@ if ! command -v python3 >/dev/null 2>&1; then
   exit 0
 fi
 
+# Capture python3's actual directory so the hermetic per-case PATH
+# below can include it. The hermetic PATH (`$bin_dir:/usr/bin:/bin:
+# /usr/sbin:/sbin`) is correct for shim-PATH semantics but assumes
+# python3 lives under `/usr/bin` — which is true on GitHub Actions
+# runners (most distros) but not on dev macOS where python3 lives
+# under `/opt/homebrew/bin` or `/usr/local/bin`. Without this, the
+# hermetic invocation of the deploy script can't resolve `python3`
+# even though preflight just verified it exists. (nathanpayne-codex
+# Phase 4b r1 on PR #290.)
+PYTHON3_DIR="$(cd "$(dirname "$(command -v python3)")" && pwd -P)"
+
 # ----------------------------------------------------------------------
 # Test harness
 # ----------------------------------------------------------------------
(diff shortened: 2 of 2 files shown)


nathanpayne-codex (changes requested):
Changes requested on `c09eeac`.

`scripts/ci/check_op_firebase_deploy_integration` verifies `python3` in the caller environment at line 72, then later runs the harness under `env -i` with `PATH="$bin_dir:/usr/bin:/bin:/usr/sbin:/sbin"` at line 165 (and the second hermetic runner uses the same pattern). On machines where `python3` is only available through Homebrew, pyenv, or another non-system directory, the prerequisite check passes but the actual hermetic case loses `python3` and fails.

Keep the shim directory first, but preserve the discovered `python3` directory or append the original caller PATH for host tools before shipping this local smoke test.

nathanpayne-codex (approved):
Approved on `528a468`.

The r2 fix captures `PYTHON3_DIR` at script init and uses it in both hermetic `env -i` PATHs. I ran the opt-in integration harness with `MERGEPATH_RUN_INTEGRATION=1`; all four resolver precedence cases passed, along with the default skip path, shell syntax, and `git diff --check`.

nathanpayne-codex:
@codex review

Branch updated against main to clear BEHIND. Re-confirming clearance on the new HEAD; auto-merge queued and will fire when the gate clears.

Issue #284, opened 2026-05-15
feat: pre-action identity check across all gh write/read paths


nathanjohnpayne:
## Problem

Agents post writes under the wrong gh identity when the keyring's
active account drifts mid-session — and the existing guards only
catch a narrow slice of that surface. Concrete incident from this
session: while iterating on mergepath#283, I (claude) ran
`gh pr comment 283 --body "..."` to reply to a Phase 4b
CHANGES_REQUESTED review. The keyring was still on
`nathanpayne-codex` from the user's prior Phase 4b review session,
so the comment posted under `nathanpayne-codex` even though the
content was clearly from the agent author. `op-preflight.sh`
**did emit the warning**:

```
# WARNING: gh active account is 'nathanpayne-codex' (expected 'nathanpayne-claude').
#   Reviewer-identity writes (gh pr review) will mis-attribute.
#   Fix once: gh auth switch -u nathanpayne-claude
```

…but the warning is informational, not enforcement, and the
agent (me) did not act on it before issuing the write. The
misattribution required a delete + re-post to clean up, which is
costly across N writes per session and silently corrupts the
audit trail when missed.

The fix is to **enforce identity-correctness at the write boundary**,
not rely on the agent to read and respond to a warning every tool
call. And to **expand the scope from `gh pr create`/`merge`/`edit`**
(today's hook coverage) **to every operation that has an identity
attached to it**, including reads.

## Current state

What's already in place:

- **`scripts/gh-as-author.sh`** — wrapped switch+execute+restore for
  the author identity. Used for `gh pr create`, `gh pr merge`,
  `gh pr edit`. Hard-rule per `CLAUDE.md § Active-account convention`.
- **`scripts/gh-as-reviewer.sh`** — companion wrapper for the
  reviewer identity. Mostly a no-op on agent machines where the
  reviewer identity is already the keyring active, but the
  switch+restore guarantee is the safety net for cross-identity
  sessions (this turn would have caught the bug had the comment
  call been wrapped).
- **`scripts/hooks/gh-pr-guard.sh`** — PreToolUse hook that blocks:
  - `gh pr create` when keyring active ≠ `nathanjohnpayne`
  - `gh pr merge --admin` without `BREAK_GLASS_ADMIN=1`
  - `gh pr merge` against a PR with `mergeStateStatus: BLOCKED`
  - `gh pr merge` against a PR carrying `needs-external-review`
  - Includes a robust shlex tokenizer that survives multi-line
[...]

Issue #281, opened 2026-05-15
feat: standardize the Phase 4b agent-to-CLI handoff text block


nathanjohnpayne:
## Problem

Whenever Phase 4b external review is required — single-PR or a multi-PR
wave — the agent should emit a copy-pastable handoff text block that
gives the human everything needed to post APPROVED reviews efficiently
via `nathanpayne-codex` in a separate CLI session.

Today this is implicit. `REVIEW_POLICY.md § Handoff Message Format`
defines the per-PR comment that lands *on the PR itself*, but the
**agent-to-human / chat-side** part of the handoff — the block the
human copies into their codex CLI prompt — is ad-hoc. When the queue
is one PR, the agent typically pastes a URL. When the queue is many
(e.g., the current 263caf3 propagation wave: mergepath#270 + 8
consumer sync PRs), the human ends up asking the agent to assemble an
inventory, burning a turn.

## Concrete trigger

In the 263caf3 propagation-wave turn, the agent produced this batched
handoff on demand after the human asked for it:

> PRs to review, all share the same diff (263caf3 canonical mirror +
> the fix from mergepath#280, plus matchline only has one extra
> converge commit):
>
>     mergepath#270 head 1682494                       — update-branch re-confirm
>     nathanjohnpayne/matchline#232 head 549c9e49      — sync + converge + P2 fix
>     nathanjohnpayne/nathanpaynedotcom#368 head a6b8e092
>     nathanjohnpayne/overridebroadway#59 head 6a62fc94
>     nathanjohnpayne/device-source-of-truth#86 head 8b4acd90
>     nathanjohnpayne/friends-and-family-billing#270 head 06f6b610
>     nathanjohnpayne/device-platform-reporting#80 head 37ba59c4
>     nathanjohnpayne/swipewatch#51 head 81123122
>     nathanjohnpayne/tadlockpsychiatry#54 head 56a64d41
>
> The eight consumer PRs are byte-identical mirrors of
> mergepath@263caf3 — the verify-propagation-pr.sh lane logic would
> exempt them under #268, but the lane workflow is being installed BY
> this very wave, so it's not active yet. After this wave merges, the
> next wave runs through the lane. A spot-check on the diff against
> mergepath HEAD on 1–2 of them is the practical Phase 4 for the wave.

User feedback: this is the right shape; standardize it so it's emitted
by default rather than on demand.

## Proposal

Define a chat-side handoff block with two variants and require the
[...]

PR #291, opened 2026-05-15
feat: standardize chat-side Phase 4b handoff block (#281)

nathanjohnpayne:
Add a canonical chat-side handoff block (single-PR and batch variants)
that agents emit into chat alongside the existing PR-side comment when
Phase 4b external review is triggered. The block is copy-paste-ready
for the external reviewer's CLI session — one keystroke briefs the
session in the common case, and the batch variant lets a propagation
wave hand off N consumer PRs in a single message rather than N.

Doc changes (additive — no rewrites of existing sections):
- REVIEW_POLICY.md gains § Handoff Message Format § Chat-side handoff
  block with both variants and an explicit "additive, not a
  replacement" call-out.
- CLAUDE.md Phase 4b step (b) and the step 9g Phase 4b checkpoint now
  reference the chat-side block before "alert the human".
- AGENTS.md mirrors the requirement in the Phase 4b summary line.

Tooling:
- scripts/post-phase-4b-handoff.sh — stdout-only renderer. Takes one
  or more PR refs (<num> or owner/repo#num), fetches HEAD/base SHA,
  unresolved thread count, and a content classification (branch name
  matching mergepath-sync/[sync-all-]<sha> → "verbatim mirror of
  mergepath@<sha>"; otherwise "novel work"; commit count >1 on a sync
  branch → "sync + N convergence commits"). Emits single-PR or batch
  variant depending on argc. NEVER posts comments or edits labels —
  read-only gh API GETs only.
- tests/test_post_phase_4b_handoff.sh — PATH-shimmed gh; 6 cases
  covering single mirror, single novel, batch all-same-mirror, batch
  mixed, usage, and invalid-ref. No live GitHub network.
- .mergepath-sync.yml: both files registered as canonical with
  comments referencing #281.

Authoring-Agent: claude

## Self-Review

- **Correctness.** The doc changes are surgical: one new subsection
  in REVIEW_POLICY.md, three pointer lines added to CLAUDE.md /
  AGENTS.md. No existing wording rewritten. Step 8.5 in CLAUDE.md
  also references the "Phase 4b handoff" but it's a summary that
  defers to step 9g for the operational form; leaving it untouched
  keeps the diff minimal as required.
- **Regression risk.** Helper is stdout-only and net-read-only — it
  cannot mutate remote state by construction (no `gh pr review`,
  `gh pr edit`, `gh pr merge`, or `gh api -X POST/PATCH/DELETE` calls
  anywhere). Worst case on a malformed PR ref is exit 2 with usage.
[...]

PR #292, opened 2026-05-15
feat: tighten op-preflight contract — --check mode, default review, helper auto-source (#282)

### LINKED ISSUE #282: investigation: tighten the agent ↔ op-preflight contract
## Problem

Agents are repeatedly burning 1Password biometric prompts mid-session
even though `scripts/op-preflight.sh` already cached the credentials.
The user reports this from both `nathanpayne-claude` and
`nathanpayne-codex` sessions — which means it is not an agent-specific
bug, it is an **instruction problem**: the operating contract between
"the docs" and "the agent" leaves enough room for the wrong pattern
that two independent agents fall into the same trap.

Concrete evidence from this session:

1. Initial `scripts/op-preflight.sh --agent claude --mode all` run
   issued biometric prompts for PATs + GCP ADC + Cloudflare token +
   SSH warms (5 prompts in a single fetch, of which only PATs were
   needed for the review work the session was about to do).
2. User intervened: "Did you stop with the preflight? I'm getting a
   lot of biometric prompts." — Switched the session to `--mode
   review` to skip ADC/CF prompts.
3. Later in the session: `--mode all` got triggered again
   accidentally because of a broken GCP ADC, causing a fresh biometric
   storm even though the PATs were already cached.
4. Codex sessions exhibit the same behavior independently — strong
   signal the docs lead both agents to the inefficient pattern.

## Current architecture (audit)

`scripts/op-preflight.sh` is well-designed for the "cache once, reuse
many" model:

- Writes resolved secrets to `$XDG_CACHE_HOME/mergepath/op-preflight-<agent>.env`
  (chmod 600, parent dir chmod 700).
- TTL anchor is an **embedded epoch** (`OP_PREFLIGHT_CREATED_AT_EPOCH`),
  not file mtime — `touch` cannot extend lifetime. Default TTL 14400s
  (4h), override via `OP_PREFLIGHT_TTL_SECONDS`.
- Cache-hit fast path validates the mode-specific fields are present
  (review/deploy/all) and the ADC file is still mints-a-token-usable
  before emitting exports; falls through to full fetch otherwise.
- SSH-warm marker is on its own TTL (1800s default) — independent
  clock because the 1Password SSH agent has a shorter session life
  than the PAT cache.
- Subshell isolation in `emit_from_session_file()` prevents
  cross-agent variable pollution (e.g. a stale `--agent claude` cache
  bleeding into a `--agent codex --mode review` call).
- Active-account mismatch warning when keyring is on the wrong
  identity for the chosen `--agent`.

**This is solid.** The script does its job. The breakdown is at the
boundary between the script and the agent calling it.

### A. Doc-pattern bleed: `op read` shown side-by-side with cached vars

`REVIEW_POLICY.md` and `DEPLOYMENT.md` both present **inline
`$(op read 'op://...')`** examples in the PAT-lookup-table section
with the same visual weight as the cached-var examples. An agent
scanning for "how do I read the reviewer PAT" sees the `op read` path
in the table and uses it.

Concrete sites (file:line, verbatim quotes):

- `REVIEW_POLICY.md:83` — `GH_TOKEN="$(op read 'op://Private/pvbq24vl2h6gl7yjclxy2hbote/token')" gh api user --jq '.login'`
- `REVIEW_POLICY.md:113–114` — narrative: "If `op whoami` says you are not signed in, still run the `op read ...`"
- `DEPLOYMENT.md:310` — `gh secret set REVIEWER_ASSIGNMENT_TOKEN ... --body "$(op read 'op://...')"`
- `DEPLOYMENT.md:461` — `GH_TOKEN="$(op read 'op://...')" gh api user --jq '.login'`
- `DEPLOYMENT.md:502` — `GH_TOKEN="$(op read 'op://Private/<item-id>/token')" gh api user`

The lookup tables in `~/GitHub/CLAUDE.md` and
`mergepath/REVIEW_POLICY.md` each list the four agents with their
`op read` paths *as the primary thing to copy*. The cached-var
equivalents (`$OP_PREFLIGHT_REVIEWER_PAT`) are mentioned elsewhere
but not in the same table.

### B. Soft-language fallback that reads as "either is fine"

`mergepath/CLAUDE.md:82–96` says:

> "If preflight was not run (or failed), fall back to inline `op read`
> (original pattern)."

This is the right structural rule, but it presents the fallback as
equivalently safe rather than "biometrically expensive — only when
broken." An agent that hasn't internalized the biometric cost of
`op read` will treat the two paths as interchangeable.

### C. No script-side enforcement

Helpers that take `GH_TOKEN` (`coderabbit-wait.sh`,
`codex-review-request.sh`, `codex-review-check.sh`,
`resolve-pr-threads.sh`, `request-label-removal.sh`) check that
`GH_TOKEN` is set but do **not** check `$OP_PREFLIGHT_DONE` or
attempt to source the cached env file themselves. An agent that
forgot to `eval "$(...op-preflight.sh...)"` at the top of a tool
call gets a usage error and naturally falls back to inline `op read`
to produce a `GH_TOKEN` value — burning a biometric every time.

### D. `--mode all` is the default

`scripts/op-preflight.sh` defaults `MODE=all`. For 90 %+ of agent
work (Phase 4 review loops), only `--mode review` is needed —
fetching GCP ADC and the Cloudflare token costs two unnecessary
biometric prompts per refresh. Worse, a stale ADC fails
`adc_is_usable()` and silently triggers a re-fetch even when the
agent never wanted ADC in the first place.

### E. Cross-tool-call subshell amnesia

Claude Code (and Codex CLI) spawns a fresh subshell per tool call.
`OP_PREFLIGHT_*` exports do NOT survive across tool calls — they have
to be re-eval'd. The cache-hit fast path makes this nearly-free, but:

1. There's no documented agent operating-loop pattern that says
   "every tool call that touches GitHub should `eval "$(scripts/op-preflight.sh --agent <X> --mode review)"` first."
2. The `--mode review` default would make the fast path even cheaper,
   but agents often run `--mode all` "to be safe."
3. The script's stderr output on a cache hit is verbose
   ("# Preflight cached hit (age Xs / TTL 14400s) ..."), which
   visually suggests work was done — an agent reading the output may
   think it triggered biometric when in fact it didn't.

### F. `--refresh` used defensively

`--refresh` exists for the legitimate "I think my PATs were rotated"
case but agents reach for it as a safety blanket, forcing a full
biometric burst when nothing needed refreshing.

### G. Multiple sources of truth for the PAT table

The PAT lookup table appears at least three times in slightly
different forms:

- `~/GitHub/CLAUDE.md` (machine-level, all four agents)
- `mergepath/REVIEW_POLICY.md` (project-level, four agents)
- `mergepath/DEPLOYMENT.md` (deploy doc, four agents)

Subtle differences (e.g. paths, agent names, fallback language) mean
an agent that learned the pattern from one doc may not recognize
patterns from another.

### High-leverage, low-risk

1. **Re-rank the docs.** In `REVIEW_POLICY.md`, `DEPLOYMENT.md`, and
   both `CLAUDE.md`s, the cached-variable pattern
   (`$OP_PREFLIGHT_REVIEWER_PAT`) must be the **primary** example in
   every PAT table and in every read-path example. The `$(op read
   'op://...')` form should appear only in a clearly marked
   "Emergency fallback / one-time setup" subsection with a banner:
   "**This triggers a biometric prompt every call. Use only when
   `op-preflight.sh` is unavailable.**"

2. **Default `--mode` to `review`, not `all`.** `--mode all` should
   be opt-in via deploy-path scripts (e.g. `scripts/deploy.sh`
   wrapper that calls `op-preflight.sh --mode all` itself). Most
   agent sessions never deploy.

3. **Add `--check` (or `--status`) mode.** A read-only check that:
   - Prints the cache state (fresh / stale / missing) to stderr.
   - Emits the cached exports to stdout if fresh.
   - Exits **non-zero with no biometric** if missing or stale.

   An agent can safely `eval "$(scripts/op-preflight.sh --agent
   claude --check)"` at the top of every tool call — it will NEVER
   prompt biometric, just reuse what's already cached. The full
   `--mode review` (which CAN prompt biometric) is reserved for the
   genuine first-call-of-the-session case.

4. **Single source of truth for the PAT table.** Move it to
   `mergepath/REVIEW_POLICY.md` only; `CLAUDE.md` and `DEPLOYMENT.md`
   reference it instead of duplicating. Drift in this table is a
   known foot-gun (different agents learn different paths).

### Script-side enforcement

5. **Helpers auto-source the cache.** `coderabbit-wait.sh`,
   `codex-review-request.sh`, `codex-review-check.sh`,
   `resolve-pr-threads.sh`, `request-label-removal.sh` should, at the
   top:
   - If `GH_TOKEN` is set: use it (preserves current behavior).
   - If `GH_TOKEN` is unset AND `$XDG_CACHE_HOME/mergepath/op-preflight-<agent>.env`
     is present AND fresh: source it and use `$OP_PREFLIGHT_REVIEWER_PAT`
     (or the appropriate one for the helper's role).
   - Only then fail with the "set GH_TOKEN" usage error.

   This removes the failure mode where an agent forgot the eval and
   resorts to inline `op read` for a one-off token.

6. **PreToolUse hook: biometric-spam detector.** Add
   `scripts/hooks/op-read-guard.sh` (sibling to `gh-pr-guard.sh`)
   that intercepts `op read` / `op item get` calls and:
   - Allows them silently if `OP_PREFLIGHT_DONE` is unset (genuine
     bootstrap / first-call case).
   - Allows them silently if the caller is `op-preflight.sh` itself.
   - **Warns** (does not block) if `OP_PREFLIGHT_DONE=1` is in
     scope, with a pointer to the cached var the call could be using
     instead. A soft warn (not block) avoids breaking legitimate
     one-time setup commands.

### Operating-loop docs

7. **Add a canonical "session loop" snippet** to `CLAUDE.md` and
   `AGENTS.md`:

   ````
   # Session start (one biometric burst):
   eval "$(scripts/op-preflight.sh --agent claude --mode review)"

   # Every subsequent tool call (idempotent, NEVER prompts):
   eval "$(scripts/op-preflight.sh --agent claude --check)"

   # Read-path API call:
   GH_TOKEN="$OP_PREFLIGHT_REVIEWER_PAT" gh api user --jq .login

   # Write-path API call (gh keyring active = nathanpayne-<agent>):
   gh pr review <PR#> --comment --body "..."

   # Author write (temporary switch):
   scripts/gh-as-author.sh -- gh pr create ...
   ````

   No more `$(op read '...')` anywhere in the loop. Agents copy this
   block, agents win.

8. **Per-tool-call check is cheap — document that.** The current
   verbose stderr ("cached hit (age 1234s / TTL 14400s)") visually
   suggests work happened. Either quiet the cache-hit stderr (silent
   when `OP_PREFLIGHT_QUIET=1`) or rewrite it to make clear that
   ZERO biometric was triggered: `# preflight: cache hit, no
   biometric burned`.

### Telemetry

9. **Log every biometric trigger.** `op-preflight.sh` should append
   to `$CACHE_DIR/biometric-log` a one-line record every time it
   actually invokes `op inject` / `op read` against a fresh fetch:
   timestamp, agent, mode, reason (full-fetch, --refresh, stale-ADC,
   cross-mode invalidation). The user can then audit "why did I get
   five prompts in 20 minutes" objectively.

### Cross-agent considerations

10. **Multi-agent inventory.** When both Claude and Codex are
    operating against the same 1Password user, they each maintain
    their own cache file (`op-preflight-claude.env`,
    `op-preflight-codex.env`). The biometric prompts for each agent
    are independent — Claude's biometric burst does NOT warm Codex's
    cache. This is intentional (different reviewer PATs) but worth
    documenting so the user knows to expect at minimum ONE prompt
    burst per active agent per TTL window.

## Acceptance criteria

- [ ] `REVIEW_POLICY.md`, `DEPLOYMENT.md`, and both `CLAUDE.md`
      files have the cached-var pattern as the primary example in
      every PAT table; inline `op read` examples moved to a clearly
      labeled "Fallback / setup-only" subsection
- [ ] `scripts/op-preflight.sh` default `--mode` changed to `review`
- [ ] `scripts/op-preflight.sh --check` (or `--status`) implemented
      and documented; behaves as: never prompts, emits cached
      exports if fresh, exits non-zero otherwise
- [ ] Read-path helpers (`coderabbit-wait.sh`,
      `codex-review-request.sh`, `codex-review-check.sh`,
      `resolve-pr-threads.sh`, `request-label-removal.sh`) auto-source
      the cache file when `GH_TOKEN` is unset
- [ ] Single source of truth for the PAT table — chosen location, all
      other docs reference it
- [ ] (Optional) `scripts/hooks/op-read-guard.sh` PreToolUse hook
      that warns on `op read` calls when preflight is fresh
- [ ] (Optional) `OP_PREFLIGHT_QUIET=1` env var quiets the cache-hit
      stderr, or stderr rewritten to make "no biometric burned"
      unambiguous
- [ ] (Optional) Biometric trigger log under
      `$CACHE_DIR/biometric-log`
- [ ] Canonical "session loop" snippet added to `CLAUDE.md` and
      `AGENTS.md`
- [ ] Propagation: all of the above shipped via `.mergepath-sync.yml`
      so consumer repos inherit the tightened contract

## Out of scope

- Replacing 1Password as the secret store
- Changing the underlying biometric model (Touch ID prompts are a
  macOS / 1Password CLI behavior; we can only reduce *call sites*)
- The auth-split between read paths (`GH_TOKEN`) and write paths
  (keyring active) — that's correctly documented and not the issue
  here
- `scripts/gh-as-author.sh` / `scripts/gh-as-reviewer.sh` — already
  correctly unset `GH_TOKEN`/`GITHUB_TOKEN` and switch keyrings
  safely; not the bug site
- The SSH-warm marker mechanics — already on a separate clock and
  correctly skipped on cache hits

## References

- `scripts/op-preflight.sh` (the cache builder)
- `mergepath/REVIEW_POLICY.md § PAT Lookup Table` (lines 74–80)
- `mergepath/REVIEW_POLICY.md § Reviewer PAT Quick Start` (lines 35–48)
- `mergepath/CLAUDE.md § Active-account convention` (lines 18–77)
- `mergepath/CLAUDE.md § Session start (run once)` (lines 82–96)
- `~/GitHub/CLAUDE.md` (machine-level PAT table)
- `mergepath/DEPLOYMENT.md § PAT Lookup Table` (lines 437–443)
- mergepath#137 (stale ADC failure mode)
- mergepath#139 (motivated the preflight cache design)
- mergepath#163 (SSH-warm separate TTL)
- mergepath#171 (active-account check)

nathanjohnpayne:
Addresses three brittleness modes in the agent ↔ op-preflight contract
identified in #282:

1. **--check / --status mode** (read-only validator). `op-preflight.sh`
   gains a `--check` flag (alias `--status`) that validates the session
   file for the requested --agent is fresh per the existing TTL anchor
   and emits the cached export statements — but NEVER invokes `op`,
   NEVER warms SSH, NEVER reads ADC. On a missing or stale cache the
   command exits non-zero with a remediation message pointing back at
   `--mode review`. This gives agents a safe idempotent "re-validate
   credentials" call at the top of every tool call without any risk
   of biometric prompt or write-side effects. Mutually exclusive with
   --refresh, --purge, --purge-all.

2. **Default --mode is now `review`** (was `all`). Most agent work
   only needs reviewer/author PATs + SSH warming; loading GCP ADC +
   Cloudflare on every session-start preflight bloated the biometric
   burst for no reason. Deploy scripts pass --mode deploy or
   --mode all explicitly.

3. **OP_PREFLIGHT_QUIET=1** collapses the verbose cache-hit stderr
   block to a single "# preflight: cache hit, no biometric burned"
   line. Refresh notices and warnings remain unaffected.

4. **Biometric trigger log** (`$CACHE_DIR/biometric-log`, chmod 600).
   Each fresh op fetch appends `<ISO8601> agent=<a> mode=<m>
   reason=<full-fetch|refresh|stale-adc|cross-mode-invalidation>` so
   a session audit can correlate biometric prompts against agent
   behavior. Always-on (local file, no privacy/log-volume tradeoff).

5. **Helper auto-source via scripts/lib/preflight-helpers.sh**. The
   five GH_TOKEN-using helpers — coderabbit-wait.sh,
   codex-review-request.sh, codex-review-check.sh,
   resolve-pr-threads.sh, request-label-removal.sh — now source a
   shared library that auto-loads the op-preflight cache when
   GH_TOKEN is unset and a fresh cache exists for the current agent.
   Preserves existing behavior when GH_TOKEN is already set. The
   `[ -z "${GH_TOKEN:-}" ] && exit 3` guard in each helper still
   fires when neither GH_TOKEN nor a fresh cache is available, but
   with a clearer remediation message pointing at op-preflight.

6. **PAT-table consolidation**. REVIEW_POLICY.md § PAT lookup table
   becomes the single source of truth. CLAUDE.md (project) and
   DEPLOYMENT.md reference it rather than duplicating. Machine-level
   ~/GitHub/CLAUDE.md keeps a mirror tagged with REVIEW_POLICY.md
   as authoritative.

7. **Cached env var is the primary pattern** across all PAT-related
   docs. The inline `op read 'op://...'` form is now clearly labeled
   "Fallback / setup-only" with a banner warning that it triggers a
   biometric prompt per call.

Tests:
  - tests/test_op_preflight_check.sh (7 tests): --check never invokes
    op, never warms SSH, emits cached exports on fresh cache, exits
    non-zero on missing/stale; --status alias; OP_PREFLIGHT_QUIET=1
    collapses verbose block; default --mode is review.
  - tests/test_helper_autosource.sh (7 tests): shared library
    auto-source loads fresh cache, preserves caller GH_TOKEN, skips
    stale caches; preflight_require_token reviewer/author both
    export GH_TOKEN; all five helpers source the shared library;
    GH_TOKEN-required helpers no longer error out with a fresh cache.
  - Wired into scripts/ci/check_op_preflight_contract and the
    repo_lint.yml workflow alongside the existing checks.
  - Added to the .mergepath-sync.yml manifest so consumers get the
    library + tests in lockstep with op-preflight.sh updates.

Closes #282 (mergepath surface only — #284's active-account
matrix and #281's chat-side handoff are out of scope here).

Authoring-Agent: claude

## Self-Review

- Correctness: --check path early-returns before any op/ssh call. All
  7 tests in test_op_preflight_check.sh PATH-shim `op` to abort on
  call and verify the stub is never invoked. Helper auto-source
  preserves the existing `[ -z "${GH_TOKEN:-}" ] && exit 3` guard so
  no regression when neither env var nor cache is available.
- Regression risk: low. The default --mode change is the only
  semantic shift in op-preflight.sh; deploy scripts that need ADC
  still pass --mode deploy or --mode all explicitly per
  DEPLOYMENT.md, and docs/agents/deployment-process.md is updated to
  reflect the new default. The helper auto-source is a strict
  superset of prior behavior (no-op when GH_TOKEN is set).
- Style: matches existing op-preflight.sh patterns
  (function-then-guard layout, numeric-only validation with `10#`
  decimal coercion, set -e safe `|| true` on optional grep paths).
  Shared library uses bash 3.2-portable parameter expansion
  throughout.
- Test coverage: 14 new tests covering the contract surface. The
  helper auto-source path is also exercised end-to-end via a stub
  gh + a synthetic cache (test 7 in test_helper_autosource.sh).
- Security/dependency hygiene: scripts/lib/preflight-helpers.sh is
  chmod 755 with no new dependencies (pure bash). The biometric log
  is chmod 600 in a chmod 700 cache dir. No PAT material added to
  any tracked file.
- Scope boundary: did NOT touch warn_active_account_mismatch or the
  SSH-warm marker code (#163), per the parallel agent coordination
  note in the issue. Did NOT add the operation-to-identity matrix
  to REVIEW_POLICY.md (that's #284's deliverable).

### DIFF
--- a/.github/workflows/repo_lint.yml
+++ b/.github/workflows/repo_lint.yml
@@ -83,6 +83,9 @@ jobs:
       - name: check_gh_as_author
         run: ./scripts/ci/check_gh_as_author
 
+      - name: check_op_preflight_contract
+        run: ./scripts/ci/check_op_preflight_contract
+
       - name: check_eslint_config_present
         run: ./scripts/ci/check_eslint_config_present
 
--- a/.mergepath-sync.yml
+++ b/.mergepath-sync.yml
@@ -140,6 +140,12 @@ paths:
   - path: scripts/op-preflight.sh
     type: canonical
     consumers: all
+  # Sourced by all five GH_TOKEN-using helper scripts; carries the
+  # auto-source logic that lets agents drop the explicit
+  # `GH_TOKEN=...` prefix once the preflight cache is warm. See #282.
+  - path: scripts/lib/preflight-helpers.sh
+    type: canonical
+    consumers: all
   - path: scripts/coderabbit-wait.sh
     type: canonical
     consumers: all
@@ -233,6 +239,16 @@ paths:
   # gh; no live GitHub network. Travels with the helper so consumers
   # exercise the same renderer they propagate.
   - path: tests/test_post_phase_4b_handoff.sh
+    type: canonical
+    consumers: all
+  # Pairs with scripts/op-preflight.sh + scripts/lib/preflight-helpers.sh
+  # + scripts/ci/check_op_preflight_contract. The check wrapper requires
+  # both test files; same coupling pattern as the gh-as-author and
+  # codex-p1-gate suites above. See #282.
+  - path: tests/test_op_preflight_check.sh
+    type: canonical
+    consumers: all
+  - path: tests/test_helper_autosource.sh
     type: canonical
     consumers: all
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -24,29 +24,51 @@ This repository uses a multi-identity AI agent code review system. The full poli
 
 ### Workflow Summary
 
-0. Run credential preflight at the start of every PR session (and safely
-   at the top of every subsequent tool call — re-running within the TTL
-   returns cached values without a new biometric prompt):
-   `eval "$(scripts/op-preflight.sh --agent {your-agent} --mode all)"`
-   This triggers biometric prompts once and writes a chmod-600 session
-   file at `$XDG_CACHE_HOME/mergepath/op-preflight-<agent>.env` so fresh
-   subshells can reuse the credentials (see REVIEW_POLICY.md § Phase 0).
-   `gh` resolves auth differently for read vs write paths: read paths
-   (`gh api user`, GETs, helper scripts) honor `GH_TOKEN`; write paths
-   (`gh pr review`, `gh pr create`, `gh pr merge`, `gh pr edit`) use
-   the keyring's **active** account. Set the active account once per
-   machine: `gh auth switch -u nathanpayne-{your-agent}`. Then use
-   `GH_TOKEN="$OP_PREFLIGHT_REVIEWER_PAT"` (or `$OP_PREFLIGHT_AUTHOR_PAT`)
-   only for read-path API calls and helper scripts. For author-identity
-   writes (PR create/merge/label edits), MUST wrap the call in
-   `scripts/gh-as-author.sh -- gh pr create ...` — a single bash process
-   that switches, runs, and switches back via `trap EXIT`. Splitting the
-   sequence across two Bash tool calls lands the PR under the wrong
-   identity (#241). The `gh-pr-guard.sh` PreToolUse hook now blocks
-   `gh pr create` when the keyring's active is not `nathanjohnpayne`.
-   See REVIEW_POLICY.md § Reviewer PAT Quick Start for the full convention
-   and § Recovery: PR created under the wrong identity for what to do
-   if a PR already landed under the wrong account.
+0. Run credential preflight at the start of every PR session. The
+   canonical session-loop snippet (read-path GH_TOKEN, write-path
+   active-keyring, author-write wrapper — full details in REVIEW_POLICY.md
+   § Phase 0 and § PAT lookup table):
+
+   ```bash
+   # Session start (one biometric burst). Default --mode is `review`.
+   eval "$(scripts/op-preflight.sh --agent {your-agent} --mode review)"
+
+   # Every subsequent tool call (idempotent, NEVER prompts):
+   eval "$(scripts/op-preflight.sh --agent {your-agent} --check)"
+
+   # Read-path API call (uses cached PAT, no biometric):
+   GH_TOKEN="$OP_PREFLIGHT_REVIEWER_PAT" gh api user --jq .login
+
+   # Write-path API call (gh keyring active = nathanpayne-{your-agent}):
+   gh pr review <PR#> --comment --body "..."
+
+   # Author write (temporary switch via the gh-as-author.sh wrapper):
+   scripts/gh-as-author.sh -- gh pr create ...
+   ```
+
+   `--check` (alias `--status`) is the lightweight re-validator: never
+   invokes `op`, never warms SSH, exits non-zero on missing/stale
+   cache. Set `OP_PREFLIGHT_QUIET=1` to collapse the cache-hit stderr
+   block. The helper scripts (`coderabbit-wait.sh`,
+   `codex-review-request.sh`, `codex-review-check.sh`,
+   `resolve-pr-threads.sh`, `request-label-removal.sh`) auto-source
+   the cache when `GH_TOKEN` is unset (#282), so the explicit
+   `GH_TOKEN=...` prefix is optional once preflight has run.
+
+   Set the gh keyring active account once per machine: `gh auth switch
+   -u nathanpayne-{your-agent}`. Then read-path commands use
+   `GH_TOKEN="$OP_PREFLIGHT_REVIEWER_PAT"`; write paths use the
+   keyring active. For author-identity writes (PR create/merge/label
+   edits), MUST wrap the call in `scripts/gh-as-author.sh -- gh pr
+   create ...` — a single bash process that switches, runs, and
+   switches back via `trap EXIT`. Splitting the sequence across two
+   Bash tool calls lands the PR under the wrong identity (#241). The
+   `gh-pr-guard.sh` PreToolUse hook now blocks `gh pr create` when
+   the keyring's active is not `nathanjohnpayne`. See REVIEW_POLICY.md
+   § Reviewer PAT Quick Start for the full convention and § Recovery:
+   PR created under the wrong identity for what to do if a PR already
+   landed under the wrong account.
+
    Run `scripts/op-preflight.sh --agent {your-agent} --purge` (or
    `--purge-all`) at end of session to delete the cached PATs.
 1. Author code as nathanjohnpayne. File a PR.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -78,22 +78,43 @@ keyring active is your agent identity. No switch needed for commits.
 
 ## Session start (run once)
 
-0. Run credential preflight to front-load all biometric prompts:
-   `eval "$(scripts/op-preflight.sh --agent claude --mode all)"`
-   This caches PATs and deploy credentials in a chmod-600 session file
-   at `$XDG_CACHE_HOME/mergepath/op-preflight-claude.env` (default
-   `$HOME/.cache/mergepath/`). Safe to re-run at the top of every tool
-   call — within the TTL (4h default, override via
-   `OP_PREFLIGHT_TTL_SECONDS`) the script reads the session file and
-   emits the same exports without a new biometric prompt. Read-path
-   API calls and helper scripts use
-   `GH_TOKEN="$OP_PREFLIGHT_REVIEWER_PAT"` (or `…AUTHOR_PAT`) instead
-   of `op read`. Write paths (`gh pr review` / `create` / `merge` /
-   `edit`) use the active keyring account regardless of `GH_TOKEN`
-   per the Active-account convention above.
+0. Run credential preflight to front-load all biometric prompts. The
+   canonical session-loop snippet (see REVIEW_POLICY.md § Phase 0 for
+   the full doc, and § PAT lookup table for the per-agent item IDs):
+
+   ```bash
+   # Session start (one biometric burst). Default --mode is `review`.
+   eval "$(scripts/op-preflight.sh --agent claude --mode review)"
+
+   # Every subsequent tool call (idempotent, NEVER prompts):
+   eval "$(scripts/op-preflight.sh --agent claude --check)"
+
+   # Read-path API call (uses cached PAT, no biometric):
+   GH_TOKEN="$OP_PREFLIGHT_REVIEWER_PAT" gh api user --jq .login
+
+   # Write-path API call (gh keyring active = nathanpayne-<agent>):
+   gh pr review <PR#> --comment --body "..."
+
+   # Author write (temporary switch via the gh-as-author.sh wrapper):
+   scripts/gh-as-author.sh -- gh pr create ...
+   ```
+
+   The cache lives at `$XDG_CACHE_HOME/mergepath/op-preflight-claude.env`
+   (default `$HOME/.cache/mergepath/`), chmod 600. The `--check` mode
+   is a read-only validator: it never invokes `op`, never warms SSH,
+   never reads ADC; on a missing/stale cache it exits non-zero with a
+   pointer back at `--mode review`. Safe to re-run at the top of every
+   tool call. Within the TTL (4h default, override via
+   `OP_PREFLIGHT_TTL_SECONDS`) the script emits the cached exports
+   without a new biometric prompt. Set `OP_PREFLIGHT_QUIET=1` to
+   collapse the cache-hit stderr block to a single line.
+
    Run `scripts/op-preflight.sh --agent claude --purge` at end of
-   session to wipe the cache. If preflight was not run (or failed), fall
-   back to inline `op read` (original pattern).
+   session to wipe the cache. If preflight is unavailable (e.g.
+   first-time bootstrap on a fresh machine), fall back to inline
+   `op read` per REVIEW_POLICY.md § PAT lookup table > Fallback —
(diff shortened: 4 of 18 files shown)


nathanpayne-codex (changes requested):
Changes requested on `ea7b5af`.

The documented `--check` contract says it never reads ADC, but `scripts/op-preflight.sh --agent codex --mode deploy --check` still routes through `emit_from_session_file` at line 610. In deploy/all modes that function calls `adc_is_usable` at line 508, which invokes `python3` and can attempt OAuth token validation.

I reproduced this with a fake fresh cache and a `python3` probe: `--mode deploy --check` printed the `no biometric` success message while executing the probe against the cached ADC path. Please make `--check` side-effect-free for deploy/all too, or reject `--check` with deploy/all explicitly instead of reaching ADC validation.

nathanpayne-codex (dismissed):
Approved on `2582a60`.

The `--check` path now propagates `OP_PREFLIGHT_CHECK_MODE=1` into `emit_from_session_file`, and the deploy/all ADC usability probe is skipped under check mode. I ran `tests/test_op_preflight_check.sh`, `tests/test_helper_autosource.sh`, `scripts/ci/check_op_preflight_contract`, shell syntax checks for the touched helpers, and `git diff --check`; all passed.

nathanpayne-claude:

nathanpayne-claude:
@codex review

Rebased onto current main (resolved a .mergepath-sync.yml conflict with #291's tests/test_post_phase_4b_handoff.sh entry — both PRs added new canonical entries to the same section; resolution kept both, identical to the auto-merge of the other files). New HEAD: 23d1062. Auto-merge queued; will fire on gate clearance.

nathanpayne-codex (approved):
Approved on `23d1062`.

I rechecked the rebased head against the prior `--check` blocker. `OP_PREFLIGHT_CHECK_MODE=1` still reaches `emit_from_session_file`, the deploy/all ADC probe is skipped in check mode, and helper auto-source preserves caller-provided `GH_TOKEN`.

Validation: `bash tests/test_op_preflight_check.sh`, `bash tests/test_helper_autosource.sh`, `scripts/ci/check_op_preflight_contract`, shell syntax checks for the touched helpers, and `git diff --check origin/main...HEAD`.

nathanpayne-claude:

Issue #241, opened 2026-05-13
gh keyring active-identity glitch: PR authored under wrong account when switch-create-switch is split across Bash invocations


nathanjohnpayne:
## Summary

A `gh pr create` invocation can land under the **wrong reviewer identity** when an agent splits the canonical `gh auth switch -u <author> && git push && gh pr create && gh auth switch -u <reviewer>` pattern across two or more Bash tool calls. The current `CLAUDE.md` guidance — "Pair the switch with a switch-back in the same bash invocation so a transient interrupt doesn't leave the keyring in the wrong active state" — is correct but is a soft recommendation. There is no enforcement, no detection, and no documented recovery procedure when it goes wrong.

This issue proposes hardening the pattern across all three axes (enforcement, detection, recovery).

## Concrete repro

Observed on [`nathanjohnpayne/friends-and-family-billing#262`](https://github.com/nathanjohnpayne/friends-and-family-billing/pull/262) (2026-05-13). The agent ran:

**Bash call A** (`gh auth status` after this: active = `nathanjohnpayne`):

```
git commit -m "..." && gh auth switch -u nathanjohnpayne && git push -u origin <branch>
# stdout: "Switched active account for github.com to nathanjohnpayne"
```

**Bash call B** (intervening reads/tests only — no gh state changes visible in the transcript):

```
gh pr create --repo ... --title "..." --body "..." && gh auth switch -u nathanpayne-claude
# stdout: pull/262 URL, then "Switched active account for github.com to nathanpayne-claude"
```

**Expected:** PR `author.login == nathanjohnpayne`.
**Observed:** PR `author.login == nathanpayne-claude`.

The companion PRs #258–#261 in the same session used the **single-Bash-call** form (switch && push && create && switch-back all in one invocation) and all landed under `nathanjohnpayne` as expected.

## Failure consequences

1. **Self-approval blocked.** The downstream reviewer flow — `gh pr review --approve` from `nathanpayne-claude` — returns `Can not approve your own pull request`. The PR is now stuck because no available reviewer identity can clear the branch-protection gate.
2. **Policy fingerprint inversion.** `Authoring-Agent: claude` in the PR body now disagrees with `author.login`, breaking any downstream audit or grep that reconciles the two.
[...]

PR #293, opened 2026-05-15
feat: pre-action identity check + extend gh-pr-guard to byline-sensitive write paths (#284)

nathanjohnpayne:
Adds a fail-closed identity assertion at the top of every gh write
path, replacing the soft-warn drift behavior that kept losing in
practice. Closes the misattribution failure mode the #283 in-session
incidents documented (a `gh pr review --comment` from `nathanpayne-
claude` landing under `nathanjohnpayne` because the keyring's active
account drifted while the agent was elsewhere; the warn was
ignored, the comment was misattributed, the audit trail inverted).

The new pieces:

- `scripts/identity-check.sh` — assertion helper with four modes:
  `--expect-author`, `--expect-reviewer`, `--expect-external <agent>`,
  and `--expect-token-identity <login>`. The first three read the
  keyring's active account via `gh config get -h github.com user`
  (the GH_TOKEN-immune signal); the last reads `gh api user --jq
  .login` for `gh api graphql` mutations whose byline follows the
  PAT, not the keyring. Fail-closed: empty reads, unset GH_TOKEN,
  and missing-mode invocations all exit non-zero with a clear
  diagnostic.

- `scripts/hooks/gh-pr-guard.sh` extended to cover `gh pr comment`,
  `gh pr review` (any flavor), and `gh issue comment`. All three
  block when the keyring's active account is the author identity
  (the reviewer-vs-author split is the load-bearing distinction in
  this codebase; misattributed reviewer-byline writes break the
  audit-trail invariant REVIEW_POLICY.md depends on). Additionally,
  `gh pr review --approve` is blocked when the PR is over-threshold
  AND the keyring's reviewer identity matches the PR's
  `Authoring-Agent:` line — the no-self-approve policy enforced at
  the hook layer.

  Raw `gh api -X POST .../comments` is intentionally NOT covered;
  the token/keyring auth matrix for raw `gh api` writes isn't
  precise enough to gate without false positives. Documented as a
  follow-up in the hook header.

- `scripts/op-preflight.sh` now AUTO-RESTORES on keyring drift
  instead of warning. If the cache-hit fast path finds active ≠
  nathanpayne-<agent>, the script runs `gh auth switch -u <agent>`
  itself and emits a clear stderr log. The previous `warn_active_
  account_mismatch` name is kept as a backwards-compat alias to the
  new function so propagation lands cleanly. The idle-trip detector
  is NOT included — auto-restore covers the common case.

- `scripts/gh-as-author.sh` and `scripts/gh-as-reviewer.sh` now run
  a post-switch verification: after `gh auth switch -u X`, they
  re-read `gh config get -h github.com user` and fail closed (exit 2)
  if the read doesn't equal X. Catches the silent-no-op flavor of
  the #241 footgun (corrupt hosts.yml, mock gh, concurrent switch
  race) that rc-check alone misses.

- Helpers wired up: `scripts/coderabbit-wait.sh` (retry-trigger
  write), `scripts/codex-review-request.sh` (`@codex review` post),
  `scripts/resolve-pr-threads.sh` (resolveReviewThread mutation,
  uses `--expect-token-identity` since this is a PAT-attributed
  graphql write), `scripts/request-label-removal.sh` (structured-ask
  post). Each call site has a SKIP_IDENTITY_CHECK opt-out for test
  harnesses.

- `REVIEW_POLICY.md` gains a new § Operation-to-Identity Matrix
  section: a canonical table mapping every `gh` / `gh api`
  operation to its required keyring identity and GH_TOKEN PAT,
  with a subsection explaining why `gh api graphql` writes are
  PAT-attributed while `gh pr ...` writes are keyring-attributed.

- `.mergepath-sync.yml` carries `scripts/identity-check.sh` and
  `tests/test_identity_check.sh` as canonical to all consumers.

Tests: 16 new identity-check cases, 12 new gh-pr-guard cases
(comment / review / issue comment / self-approve / cross-agent
allow / under-threshold allow / bypass), 1 new post-switch-verify
case each for gh-as-author and gh-as-reviewer. All 65 cases pass
locally; existing suites unbroken.

Authoring-Agent: claude

## Self-Review

- Correctness: identity-check's keyring vs PAT split mirrors the
  empirical auth-resolution behavior documented in the new matrix
  section. The hook extension reuses the existing tokenizer and
  command-position walker — only the subcommand recognition was
  widened (added `issue` parent + `comment` / `review` siblings).
  Self-approve detection uses the PR's `Authoring-Agent:` line as
  the agent fingerprint, matching how the rest of the policy flow
  identifies authoring agents.
- Regression risk: pre-existing 16 gh-pr-guard tests still pass.
  The post-switch verification in the wrappers is the only change
  with non-trivial behavioral impact — it adds a fail-closed exit
  for the silent-no-op case, but the happy path (where switch
  actually changes the active account) is unaffected. The
  `warn_active_account_mismatch` alias in op-preflight.sh keeps
  downstream consumers that called the old name working unchanged.
- Style: the new code follows the existing module conventions —
  module-header doc comments, fail-closed default for unrecognized
  states, opt-out env vars for test harnesses, bash 3.2 portability
  (no `mapfile`, no associative arrays, etc.).
- Test coverage: every new code path has a test. The
  `--expect-token-identity` mode has both happy-path and missing-
  GH_TOKEN cases. The self-approve sub-guard has cases for the
  three policy paths (over-threshold same agent → block,
  over-threshold cross-agent → allow, under-threshold same agent
  → allow).
- Security: the hook addition is defense-in-depth (the canonical
  guard remains identity-check.sh at the helper-script level). No
  new shellouts, no new network reads — the hook reuses the
  existing `gh pr view` call shape and the helpers reuse the
  existing PAT plumbing.


nathanpayne-codex (changes requested):
Changes requested on `41ba68f`.

Two blockers on the current head:

1. Required `lint` is red. `scripts/ci/check_canonical_bugs_263caf3` now fails because its `request-label-removal.sh` regression cases hit the new identity check without a configured keyring and never reach the stubbed `gh pr comment`. The two invocations around `scripts/ci/check_canonical_bugs_263caf3:90` and `:108` need the documented test bypass env so CI remains hermetic.

2. `scripts/resolve-pr-threads.sh:390` sets `IDENTITY_CHECK_FIRED=1` even when the token identity check fails, then only `continue`s the current loop item. On the next unresolved thread the guard is skipped and the `resolveReviewThread` mutation at line 405 can run without a successful identity check. A failed identity check needs to abort the script or set a persistent failed state that blocks all later mutations.

nathanpayne-codex (changes requested):
Changes requested on `b0c8463`.

The r2 move of `resolve-pr-threads.sh`'s identity check out of the per-thread loop fixes the first failure mode, and `check_canonical_bugs_263caf3` is green now. There is still a load-bearing bypass in the new identity-gate shape: the helper call sites only run `identity-check.sh` when `[ -x "$(dirname "${BASH_SOURCE[0]}")/identity-check.sh" ]` is true. If that helper is missing or loses its executable bit, the write proceeds without any identity check.

I reproduced this with a temp copy containing `resolve-pr-threads.sh` but no `identity-check.sh`: a stubbed `--auto-resolve-bots` run reached the `resolveReviewThread` GraphQL mutation and exited 0 instead of failing closed. The same conditional pattern exists in the other byline-sensitive helper call sites. Please fail closed when identity-check is unavailable unless the explicit `*_SKIP_IDENTITY_CHECK=1` test/dev escape hatch is set, and add a regression for the missing/non-executable helper path.

nathanpayne-codex (approved):
Approved on `de88e29`.

The r3 fix closes the remaining fail-open path: byline-sensitive helpers now fail closed when `identity-check.sh` is missing or non-executable unless the explicit test/dev skip env is set. The new regression coverage exercises the missing-helper and non-executable-helper paths, and the existing identity/wrapper/guard tests remain green.

Validation: `bash tests/test_identity_check.sh`, `bash tests/test_identity_check_fail_closed.sh`, `bash tests/test_gh_as_author.sh`, `bash tests/test_gh_as_reviewer.sh`, `bash tests/test_gh_pr_guard.sh`, `scripts/ci/check_canonical_bugs_263caf3`, shell syntax checks for the touched helpers/hooks, and `git diff --check origin/main...HEAD`.

nathanpayne-claude:

nathanpayne-claude:
@codex review - branch updated again after #292 merge; auto-merge still queued.

PR #294, opened 2026-05-15
feat: add manifest requires: closure invariant + canary-first docs (#264)

nathanjohnpayne:
Authoring-Agent: claude

Closes the prevention half of #264. The canonical-test manifest
entries (the immediate fix for matchline's `check_hook_tests`
failure) already landed on main; the remaining work is a manifest-
level invariant that catches future closure gaps BEFORE a wave is
opened, plus a canary-first operational procedure for `--sync-all`
that catches what the invariant can't see.

Changes:

1. `.mergepath-sync.yml` — added a `requires:` field to three path
   entries whose canonical content hard-depends on other files:
     - `.github/workflows/pr-review-policy.yml` requires
       `scripts/workflow/` (the workflow shells out to the three
       parser helpers in that kit).
     - `.github/workflows/codex-p1-gate.yml` requires
       `scripts/codex-p1-gate.sh` (the workflow's only job runs
       that script directly).
     - `scripts/ci/` kit requires 20 paths covering every hard
       dependency under `scripts/ci/check_*` — paired tests,
       upstream scripts, workflows, and fixtures. Each entry has
       an inline comment naming the check_* wrapper that consumes
       it, so future maintainers can audit the closure with grep.
   Added a top-of-file comment block documenting the `requires:`
   model and the propagation-closure problem it closes.

2. `scripts/ci/check_sync_manifest` — added the closure invariant
   as a new validation pass. For each entry with `requires:`,
   asserts (a) the field is a YAML sequence (shape check via
   `(.requires | type) == "!!seq"` to dodge mikefarah/yq's silent-
   emit-nothing behavior on `.requires[]` against a scalar) and
   (b) every required path is covered by the manifest — either an
   exact entry or strictly inside a kit prefix (slash-bounded so
   `scripts/cinema/` does NOT match `scripts/ci/`). Fails closed
   with a clear named-path diagnostic on uncovered or malformed.
   Added `MERGEPATH_MANIFEST_PATH` and `MERGEPATH_REPO_ROOT` env
   overrides so the test suite can point the check at fixture
   manifests with stub repo trees. Appends a regression-test
   invocation that runs `tests/test_check_sync_manifest.sh` after
   the structural check passes (guarded by absence of the manifest-
   path env override so the test's fixture-mode invocations don't
   recurse).

3. `tests/test_check_sync_manifest.sh` — new, 5 fixture cases:
     - requires: satisfied by exact + kit-prefix coverage → PASS
     - uncovered requires: → FAIL with named-path diagnostic
     - missing requires: (optional field) → PASS
     - scalar requires: (malformed) → FAIL via shape check
     - kit-prefix slash boundary regression → FAIL on adjacent dir
   Fixture helper materializes stub files in a scratch root so the
   pre-existing path-existence check passes.

4. `scripts/sync-to-downstream.sh` — added a "Canary-first
   procedure for --sync-all" section to the script header
   documenting the one-consumer-canary procedure that catches what
   the manifest invariant can't see (consumer-specific
   `repo_lint.yml` steps, transitively-required files outside the
   manifest). Walks through the 3-step canary → fix → fan-out
   pattern with matchline as the canonical canary repo.

## Self-Review

- **Correctness**: shape check uses yq's `type` function (separate
  pass) because mikefarah/yq's `.requires[]` silently emits no
  output on a scalar rather than erroring. Without the shape check,
  a malformed `requires: "foo"` field would pass through silently.
  The slash-bounded kit-prefix match (`pp$kp/*`) is verified by
  Case 5: `scripts/ci/` does NOT match `scripts/cinema/foo`.

- **Regression risk**: low. The new check is additive — entries
  without `requires:` are unchanged. The shape-check pass returns
  cleanly when no entry has `requires:`. The structural exits in
  the existing check are unchanged. Recursion between
  check_sync_manifest and its test suite is guarded by the
  MERGEPATH_MANIFEST_PATH env-presence check.

- **Style/conventions**: matches the inline-literal pattern from
  `check_workflow_parsers` and `check_canonical_bugs_263caf3`
  (test cases use bash heredocs for fixture manifests). The
  `requires:` comment-block-per-entry convention mirrors the
  existing per-path comment style elsewhere in
  `.mergepath-sync.yml`.

- **Test coverage**: 5 cases exercise the new pass end-to-end
  including the malformed-shape rejection and the kit-prefix
  boundary regression. The live manifest is smoke-tested by
  check_sync_manifest's CI invocation (intentionally not in the
  test file — would recurse).

- **Security/dependency hygiene**: no new dependencies. The check
  still requires yq mikefarah v4+ (already required). No network
  side effects, no file writes outside the test's scratch dir.

### DIFF
--- a/.mergepath-sync.yml
+++ b/.mergepath-sync.yml
@@ -9,6 +9,83 @@
 # plan): manifest + audit. Additional path types and per-path knobs
 # (kit-with-allowlist, three-way-merge for review-policy.yml,
 # substitution rules for templated docs) land in later layers.
+#
+# ─────────────────────────────────────────────────────────────────────
+# `requires:` — propagation closure invariant (#264)
+# ─────────────────────────────────────────────────────────────────────
+#
+# A path entry MAY declare a `requires:` list naming OTHER paths that
+# must travel with it to a consumer for the entry to function. The
+# canonical motivation is the propagation wave that triggered #264:
+# scripts/ci/check_gh_as_author was propagated as part of the
+# scripts/ci/ kit, but the test files it hard-requires
+# (tests/test_gh_pr_guard.sh, tests/test_gh_as_author.sh, etc.)
+# were NOT in the manifest. Every consumer's `lint` job then
+# failed in pre-flight on `Missing required file: tests/...`.
+#
+# The `requires:` field closes that gap structurally:
+# `scripts/ci/check_sync_manifest` enforces that for every required
+# path listed under any entry, that path is ALSO covered by the
+# manifest. "Covered" means EITHER an exact entry of its own OR a
+# path strictly inside a manifest kit entry (because kit propagation
+# walks the whole directory, so a fixtures subdir under a kit path
+# travels for free).
+#
+# Be conservative when adding `requires:` — only entries you can
+# verify by grepping the actual script/workflow for hard-required
+# paths. A `requires:` entry should be:
+#   - a test file under tests/ that a `scripts/ci/check_*` wrapper
+#     hard-fails on if missing,
+#   - a script that a propagated workflow shells out to (e.g.
+#     codex-p1-gate.yml → codex-p1-gate.sh), or
+#   - a fixture/library path that an extracted module depends on at
+#     runtime (e.g. agent-review.yml's detect-disagreement job
+#     requires scripts/disagreement-detector.cjs).
+#
+# Things that are NOT `requires:` material:
+#   - Soft dependencies (docs cross-references, comment pointers).
+#   - Paths inside the SAME manifest entry's kit directory (those
+#     travel automatically; declaring them is harmless and may help
+#     readability, but the validator accepts them either way).
+#   - Per-consumer config (e.g. .github/review-policy.yml) that is
+#     intentionally NOT propagated.
+#
+# Format:
+#   - path: scripts/ci/
+#     type: kit
+#     consumers: all
+#     requires:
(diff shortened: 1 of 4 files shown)


nathanpayne-codex (changes requested):
Changes requested on `fcf22df`.

The new manifest closure invariant checks that each `requires:` path is covered somewhere in the manifest, but it does not verify that the required entry is propagated to the same consumers as the requiring entry. That leaves a real hole: a fixture with `scripts/a.sh` as `consumers: all` requiring `scripts/b.sh`, while `scripts/b.sh` is limited to only one consumer, passes `scripts/ci/check_sync_manifest` even though the other consumers would receive `a.sh` without `b.sh`.

Please make the invariant consumer-aware: for each requiring path, the required path's consumer set needs to cover at least the requirer's consumer set, accounting for `all` and kit-prefix coverage. Add a regression test for the overlapping-consumer failure case.

nathanpayne-codex (approved):
Approved on `f9310ac`.

The r2 consumer-scope invariant closes the original hole: required paths now have to cover the requirer's consumer set, with regression cases for all->named, named-list gaps, superset OK, and all->all. I ran `tests/test_check_sync_manifest.sh`, `scripts/ci/check_sync_manifest`, shell syntax checks, and `git diff --check`; all passed. I also resolved the now-addressed Codex P1 bot thread and verified `scripts/codex-p1-gate.sh 294 nathanjohnpayne/mergepath` reports `Codex P1 unresolved: 0`.

nathanpayne-claude:

Issue #285, opened 2026-05-15
audit-branch-protection: fix ruleset targeting and auth-scope diagnostics


nathanpayne-codex:
## Problem

The closed-issue audit for the 2026-04-15 to 2026-05-15 window found a residual gap spanning #177 and #234.

#177 was correctly closed after branch protection was configured. Re-running `scripts/audit-branch-protection.sh` with the author account verifies the repo is protected. But running the same audit as `nathanpayne-codex` produces the misleading failure that #177 explicitly warned about:

```text
Note: classic branch protection not configured on main — checking rulesets instead.
FAIL: no rulesets target main on nathanjohnpayne/mergepath. PR merges are completely unprotected.
```

The author-account run succeeds:

```text
Required status checks currently enforced:
  ✓ Label Gate
  ✓ Self-Review Required
  ✓ lint

PASS: all canonical mergepath checks are required.
```

So the audit script is conflating an auth/scope problem with a real protection gap.

Separately, #234's validation pass still applies to `origin/main`: the ruleset fallback only recognizes `~DEFAULT_BRANCH` and exact `refs/heads/<branch>` includes, and once any matching include is found it collects required checks from every branch ruleset instead of only the rulesets that target the audited branch.

## Current evidence

On `origin/main` at `9c07bd2`:

- `scripts/audit-branch-protection.sh` treats any branch-protection API JSON containing `"message"` as "classic branch protection not configured", without distinguishing 403/404/scope failures from true absence.
- The ruleset target detector only accepts `~DEFAULT_BRANCH` and `refs/heads/$BRANCH`; it does not handle `~ALL`, `refs/heads/*`, or other branch-pattern includes.
- `REQUIRED_CHECKS` is collected from all `target == "branch"` rulesets rather than the matched ruleset subset.

## Implementation plan

1. Preserve and inspect the HTTP status for `GET /repos/{owner}/{repo}/branches/{branch}/protection`. If the response is an auth/scope failure, exit 2 with a clear message telling the caller to use an author/admin token rather than falling through to a false "unprotected" result.
2. Refactor the ruleset fallback into two steps: compute the set of rulesets that target the audited branch, then extract required status checks only from that set.
3. Support GitHub ruleset include forms: `~DEFAULT_BRANCH`, `~ALL`, exact `refs/heads/<branch>`, and branch glob patterns such as `refs/heads/*`.
[...]

Issue #177, opened 2026-05-02
main branch has zero protection — Label Gate / Self-Review Required failures are advisory only


nathanpayne-claude:
## Problem

`scripts/audit-branch-protection.sh` (shipped in #176) flagged that `nathanjohnpayne/mergepath@main` has **no branch protection at all** — no classic protection rules and no rulesets target the default branch. Output:

```
$ scripts/audit-branch-protection.sh
Auditing branch protection on nathanjohnpayne/mergepath@main...

Note: classic branch protection not configured on main — checking rulesets instead.
FAIL: no rulesets target main on nathanjohnpayne/mergepath. PR merges are completely unprotected.
```

This explains an oddity observed during the agent-driven sequencing batch (PRs #169, #171, #172, #173, #174, #175, #176): `gh pr view --json mergeStateStatus` returned `UNKNOWN` rather than `CLEAN` even when CI was green. With no protection rule, GitHub's mergeable-state computation has nothing to evaluate — `UNKNOWN` is the literal "no rule applies" signal.

This is the same gap class that #161 documented for downstream consumers — except mergepath itself is the worst offender. It also means:
- The Label Gate workflow runs on every PR but its failure has been advisory the whole time.
- Self-Review Required workflow same.
- Codex/CodeRabbit checks are advisory.
- Anyone with write access could push directly to main, bypassing the entire PR flow that the repo's docs prescribe.

## Why this happened

The repo's docs and the agent-side hook (`scripts/hooks/gh-pr-guard.sh`) both prevent agents from pushing to main and from merging without proper review. So in practice the gap hasn't been exploited. But the protection is doctrinal-only — there's no GitHub-side enforcement.

Likely the repo was created from another template that didn't propagate branch protection (since protection lives in repo settings, not in the source tree).

## Fix

Configure branch protection on `main` via Settings → Branches → Branch protection rule:

1. **Branch name pattern:** `main`
2. **Require a pull request before merging:** ✓
   - Required approvals: 0 (the repo's review policy handles this in-flow; GitHub's count would conflict with the no-self-approve rule)
   - Dismiss stale pull request approvals when new commits are pushed: ✓ (optional, cleaner)
3. **Require status checks to pass before merging:** ✓
   - Require branches to be up to date before merging: ✓
   - Required checks (each must have run at least once on the repo for the dropdown to surface them):
     - `Label Gate`
     - `Self-Review Required`
     - `lint`
     - `External Review Check`
[...]

PR #295, opened 2026-05-15
fix: distinguish auth-scope failures from real protection gaps + scope ruleset checks (#285)

nathanjohnpayne:
The 2026-04-15 to 2026-05-15 closed-issue audit found a residual gap
where `scripts/audit-branch-protection.sh` conflated three distinct
states under a single "classic branch protection not configured"
banner:

1. **Auth-scope failure (403/401)**: reviewer PATs typically lack
   `Administration:read`. Hitting that endpoint with a reviewer PAT
   would return 403, but the script grep'd for `"message":` in the
   JSON output and fell through to the rulesets endpoint — which
   often returns an empty list under the same auth, producing the
   misleading "FAIL: no rulesets target main on X. PR merges are
   completely unprotected." verdict from #177. The author PAT run
   simultaneously shows the repo is fully protected.

2. **True 404 absence**: no classic protection AND no rulesets —
   the genuine unprotected case.

3. **Mixed ruleset state**: the ruleset fallback also had two bugs:
   - it only recognized `~DEFAULT_BRANCH` and exact `refs/heads/$BRANCH`
     includes (missing `~ALL` and branch globs like `refs/heads/*`)
   - once ANY ruleset's include matched, it collected
     `REQUIRED_CHECKS` from EVERY `target == "branch"` ruleset
     regardless of include match — so a `refs/heads/dev`-only
     ruleset's checks would contaminate an audit of `main`.

## Fix shape

- Switch from grep-on-message to `gh api -i` so we can inspect the
  HTTP status code directly. 200 → use body; 404 → ruleset fallback;
  401/403 → exit 2 with a diagnostic pointing at
  `OP_PREFLIGHT_AUTHOR_PAT` and refusing to fall through; anything
  else → exit 2 with the raw body.
- Split the ruleset fallback into two distinct steps:
    Step A — compute the subset of rulesets whose
    `conditions.ref_name.include` actually matches the audited branch.
    Step B — collect `required_status_checks` ONLY from that matched
    subset.
- Support four include forms in Step A: `~ALL`, `~DEFAULT_BRANCH`
  (treated as matching main/master), exact `refs/heads/<branch>`,
  and bash-glob `refs/heads/*` patterns.
- Add `tests/test_audit_branch_protection.sh` with eight fixtures
  covering the 403 / 401 / true-404 / classic / ~ALL / glob /
  dev-only-no-contamination / mixed-rulesets cases. The stub
  emulates `gh api -i` headers + body and HTTP exit codes the same
  way real gh does.
- Add a note to `REVIEW_POLICY.md` § "Required status checks
  (per-repo setup)" calling out that reviewer PATs commonly lack
  `Administration:read` and the audit will surface that as exit 2
  rather than a false "unprotected" verdict.

## Test plan

- `bash tests/test_audit_branch_protection.sh` — 8/8 PASS locally.
- `bash -n scripts/audit-branch-protection.sh` — syntax clean.
- `--help` still renders.

Authoring-Agent: claude

## Self-Review

- **Correctness**: Each scenario is mapped to a concrete HTTP status
  via the stub; the script's status-line parser handles `HTTP/2 200`
  and `HTTP/1.1 200 OK` shapes (regex on `^HTTP/`, field 2).
  Ruleset matching is per-include, with `[[ str == $pat ]]` for the
  bash glob form (the `# shellcheck disable=SC2053` comment documents
  the intentional unquoted RHS). `~DEFAULT_BRANCH` is approximated as
  matching main/master — a precise check would need an extra
  `gh api repos/<r>` call to read `default_branch`; that's an
  acceptable trade-off for the 99% case and is documented inline.
- **Regression risk**: The classic-protection happy path is
  unchanged in behavior (still parses
  `required_status_checks.contexts`). The ruleset fallback's outputs
  for previously-supported include forms (`~DEFAULT_BRANCH`,
  `refs/heads/<branch>`) are preserved by the per-form case in the
  matcher. The new contamination guard tightens behavior — any
  pre-existing repo whose audit was previously "accidentally
  passing" via leaked checks from an unrelated ruleset would now
  surface as a fail, which is the intended correction.
- **Style**: Matches existing script idioms (bash 3.2 portable,
  `set -eo pipefail`, jq for JSON, sed-prefix for diagnostics).
- **Test coverage**: 8 fixtures cover all the listed acceptance
  criteria plus 401-as-403-twin and mixed-rulesets contamination.
- **Security/deps**: No new deps. The script still requires `jq`
  and `gh`; the test stubs `gh` only via PATH and runs under a
  short-lived `mktemp -d` workdir cleaned by `trap EXIT`.


nathanpayne-codex (changes requested):
Changes requested on `dec125a`.

The ruleset fallback still treats a ruleset as matching the audited branch solely from `conditions.ref_name.include`. It never evaluates `conditions.ref_name.exclude`, so a ruleset with `include: ["~ALL"]` and `exclude: ["refs/heads/main"]` is treated as protecting `main`. That can make the audit pass when `main` is outside all required-check rulesets.

Please make Step A reject rulesets when the audited branch matches any exclude pattern, and add a fixture for the include-then-exclude case.

nathanpayne-codex (approved):
Approved on `6d70e80`.

I rechecked the r2 delta: `conditions.ref_name.exclude` is evaluated with the same matcher as include, and any matching exclude now disqualifies the ruleset before required checks are collected. The new `ruleset_all_but_main_excluded` fixture covers include `~ALL` plus exclude `refs/heads/main`.

Validation: `bash tests/test_audit_branch_protection.sh`, `bash -n scripts/audit-branch-protection.sh`, `git diff --check origin/main...HEAD`. I also resolved the stale Codex P1 thread after verifying the fix and re-ran the P1 gate.

nathanpayne-codex:

Issue #286, opened 2026-05-15
closed-review backlog: fix residual deploy strictness and mktemp portability gaps


nathanpayne-codex:
## Problem

The closed-issue audit found that #234 was closed after structural backlog controls shipped, but several concrete script findings from its own validation pass remain present on `origin/main` and are not otherwise tracked by an open issue.

I am intentionally excluding findings already covered by active work, such as the `codex-review-check.sh` Authoring-Agent parsing/self-approval fixes in open PR #283. This issue tracks the still-unowned residuals.

## Current evidence

Validated against `origin/main` at `9c07bd2`:

1. **Deploy build command does not run under strict shell options.**
   `scripts/deploy.sh` still executes:

   ```bash
   bash -c -- "$BUILD_CMD"
   ```

   That means compound build commands can mask failures unless the command string itself opts into strict behavior.

2. **Deploy does not block a dirty working tree.**
   The deploy guard checks branch and remote freshness, but there is no `git status --porcelain`, `git diff --quiet`, or equivalent clean-tree gate before build/deploy. Issue #77 closed the stale-branch deploy class, but uncommitted local edits on `main` can still be shipped.

3. **`policy-sim.sh` and its test still use BSD/macOS-only `mktemp -t` templates.**
   Current lines use forms like:

   ```bash
   mktemp -d -t mergepath-sim
   mktemp -d -t mergepath-prs
   mktemp -d -t mergepath-test
   ```

   These pass on macOS but fail on GNU coreutils because the template lacks `X` placeholders.

## Implementation plan

1. Change `scripts/deploy.sh` to execute build commands under strict options, for example `bash -euo pipefail -c -- "$BUILD_CMD"`, and add tests for a compound command where an early failing command would otherwise be masked.
2. Add a clean-working-tree guard before build/deploy. It should fail by default when tracked files are modified or staged, print the dirty paths, and allow an explicit human break-glass flag only if the existing deploy policy accepts that tradeoff.
3. Replace the `mktemp -d -t mergepath-*` forms in `scripts/policy-sim.sh` and `tests/test_mergepath_playground.sh` with portable templates that work on both macOS and GNU coreutils.
4. Add tests that exercise the portable mktemp helper or at least grep-protect the old no-`X` forms from returning.
5. Run `bash tests/test_mergepath_playground.sh` and the deploy-script test coverage after the changes.

## Acceptance criteria

[...]

PR #296, opened 2026-05-15
fix: deploy strict-bash + dirty-tree guard + portable mktemp (#286)

nathanjohnpayne:
Closed-review backlog sweep on #286 surfaced three residual gaps left
from the deploy + playground stories. Patch all three plus a
structural regression check to prevent the mktemp form from coming
back.

Authoring-Agent: claude

1. scripts/deploy.sh — strict-bash BUILD_CMD invocation.

   Old: `bash -c -- "$BUILD_CMD"`.  A compound BUILD_CMD like
   `npm run lint; npm run build` masks lint failure behind the build's
   exit code; `npm run build | tee log` masks build failure behind
   tee's success. Both are silently broken deploys.

   New: `bash -euo pipefail -c -- "$BUILD_CMD"`. set -e aborts on the
   first failing simple command; set -u catches typos that would
   silently expand to empty; pipefail preserves a failing step's exit
   code through a pipeline. Verified regression case:

     BUILD_CMD='false; echo should-not-run' scripts/deploy.sh ...

   Under the old form, the script printed 'should-not-run' and
   reached op-firebase-deploy with rc=0. Under the new form, the
   subshell aborts on `false` and the script exits non-zero before
   reaching deploy. Covered by tests/test_deploy.sh case 1.

2. scripts/deploy.sh — clean-working-tree guard (guard 3).

   Before build/deploy, run `git status --porcelain`. If non-empty,
   list the modified/staged/untracked paths and exit non-zero with
   the diagnostic "Refusing to deploy: working tree is dirty.".
   Closes the same #77 failure class for in-progress local edits:
   a dirty tree ships whatever the WIP edits compile to, which
   diverges from the merged-on-main state reviewers approved.

   Break-glass override: DEPLOY_ALLOW_DIRTY=1 (env var, not a CLI
   flag, kept separate from --force so the override is deliberate
   and audit-greppable, and --force can never accidentally subsume
   it). When used, the script logs a ⚠️ DEPLOY_ALLOW_DIRTY=1 banner
   plus the dirty paths to stderr so the deviation is visible in the
   deploy transcript. Documented in DEPLOYMENT.md § Deploy guards
   and in the script header. Covered by tests/test_deploy.sh cases
   2 (rejects dirty) and 3 (allows under override, logs banner,
   reaches deploy step).

3. scripts/policy-sim.sh + tests/test_mergepath_playground.sh +
   tests/test_sync_to_downstream.sh — portable mktemp.

   BSD-only `mktemp -d -t <prefix>` form was reintroduced in three
   tracked scripts. GNU coreutils mktemp requires an explicit
   `XXXXXX` placeholder in the template and exits non-zero on the
   short form, so every Linux CI runner that reaches these scripts
   fails the line. Replaced with the portable form:

     mktemp -d "${TMPDIR:-/tmp}/<prefix>.XXXXXX"

   which works on both BSD/macOS and GNU/Linux. Three call sites
   fixed: scripts/policy-sim.sh:31 (OUT_DIR), :51 (PRS_DIR);
   tests/test_mergepath_playground.sh:19 (TMPDIR_SAFE);
   tests/test_sync_to_downstream.sh:29 (WORKDIR).

4. scripts/ci/check_mktemp_portability — structural regression guard.

   New CI check that hard-fails on any tracked shell file using the
   BSD-only `mktemp -t <prefix>` short form. Scans git-tracked
   *.sh + bash/sh-shebanged files, skips comment lines (so the
   docs in this PR don't false-positive) and quoted strings fed
   to grep/awk/sed (so existing tests that assert the bad form is
   absent in OTHER scripts don't false-positive either). The match
   regex is intentionally narrow — only flags the BSD shape
   (`-t bare-token`), not the portable `mktemp dir/name.XXXXXX`
   form.

   Wired into .github/workflows/repo_lint.yml between
   check_coderabbit_config_tests and check_governance_files. Verified
   against a synthetic violation in a throwaway repo (exits 1 with
   path + line number) and against the real repo after fix (exits 0).

5. tests/test_deploy.sh — new.

   Three cases against a self-contained git fixture per case
   (no remote; uses --force to skip guards 1+2 since the test's
   purpose is to exercise build + guard 3). PATH-shims
   op-firebase-deploy so the script can reach its deploy step in
   the positive case. All three pass locally.

Touched files outside the strict #286 whitelist:

- tests/test_sync_to_downstream.sh — one-line mktemp fix. In scope
  for "Grep the file for ALL mktemp -t occurrences and fix every
  one"; the check_mktemp_portability guard would otherwise fail CI
  on the repo as-is.

Test results (local):

  bash tests/test_deploy.sh                  → 3 passed, 0 failed
  bash tests/test_mergepath_playground.sh    → OK
  bash tests/test_sync_to_downstream.sh      → PASS
  bash tests/test_sync_overrides.sh          → 31 passed, 0 failed
  ./scripts/ci/check_mktemp_portability      → PASS
  ./scripts/ci/check_spec_test_alignment     → PASS
  ./scripts/ci/check_required_root_files     → PASS
  ./scripts/ci/check_workflow_parsers        → PASS (29 tests)

## Self-Review

- **Correctness.** The strict-bash change is the minimum invariant
  preserving change (still `bash -c --`, just with the safety flags
  set). The dirty guard uses `git status --porcelain` which is the
  documented stable porcelain (v1) — non-empty IFF the tree has
  modified/staged/untracked content. The portable mktemp form is
  what coreutils' own docs recommend.

- **Regression risk.** The strict-bash flip changes the exit code
  surface of BUILD_CMD: anything that previously masked a failure
  will now fail loudly. That's the point — it's an opt-out class of
  failure being closed, not an opt-in capability being added.
  Existing BUILD_CMD values like `npm run build` are single
  commands and behave identically. Compound BUILD_CMDs that
  intentionally allowed failure are extremely unlikely in practice
  (deploys want zero tolerance for build failure) but are still
  expressible via `bash -c 'cmd_that_may_fail || true'` if needed.

- **Style.** Comments cite issue numbers (#77 for the worktree
  story, #286 for this sweep). Override behavior is documented in
  both the script header AND DEPLOYMENT.md so neither path leaves
  a user confused about the env-var-vs-flag split. The new CI
  check follows the same shape as the other scripts/ci/check_*
  files (set -euo pipefail, REPO_ROOT resolution via BASH_SOURCE,
  prefixed `check_<name>: PASS|FAIL` final line, no external deps
  beyond git + grep).

- **Test coverage.** Three deploy cases (one strict-bash + two
  dirty-tree branches), the structural mktemp check both as a
  positive test (no violations in current repo) and verified
  against a synthetic violation in a throwaway repo. The existing
  test_mergepath_playground.sh assertions (`mktemp -d` present, no
  `mktemp ... .ext)` short form) still hold after the change.

- **Security / dependency hygiene.** No new external deps. The
  DEPLOY_ALLOW_DIRTY override is env-var only and the script's
  invocation chain doesn't propagate it to children (no `export`),
  so an op-firebase-deploy sub-call can't see it and act on it.
  No new network calls. The deploy script still runs build
  commands as the invoking user with no privilege escalation.

### DIFF
--- a/.github/workflows/repo_lint.yml
+++ b/.github/workflows/repo_lint.yml
@@ -142,6 +142,14 @@ jobs:
         # added by #256 P2.
         run: ./scripts/ci/check_coderabbit_config_tests
 
+      - name: check_mktemp_portability
+        # Regression guard for mergepath#286. Hard-fails on any
+        # tracked shell file using the BSD-only `mktemp -t <prefix>`
+        # form, which GNU coreutils mktemp rejects (XXXXXX required).
+        # Closes the re-introduction surface that the closed-review
+        # backlog sweep on #286 caught.
+        run: ./scripts/ci/check_mktemp_portability
+
       - name: check_governance_files
         run: |
           MISSING=0
--- a/DEPLOYMENT.md
+++ b/DEPLOYMENT.md
@@ -559,12 +559,20 @@ scripts/deploy.sh --skip-cf-purge
 scripts/deploy.sh --force
 ```
 
-The guards (see [mergepath#77](https://github.com/nathanjohnpayne/mergepath/issues/77) for the incident that motivated them):
+The guards (see [mergepath#77](https://github.com/nathanjohnpayne/mergepath/issues/77) for the incident that motivated them; the dirty-tree guard was added in [mergepath#286](https://github.com/nathanjohnpayne/mergepath/issues/286) after a closed-review backlog sweep):
 
 1. **Current branch must be `main`.** Deploys should ship the reviewed, merged state of the project, not a worktree's in-progress branch.
 2. **Local `main` must not be behind `origin/main`.** After `git fetch`, `git rev-list --count HEAD..origin/main` must be 0. Otherwise the deploy refuses.
+3. **Working tree must be clean.** `git status --porcelain` must return empty — no modified, staged, or untracked paths. A dirty tree means the deploy would ship whatever the in-progress edits compile to, which diverges from the merged-on-main state reviewers signed off on (same failure class as #77).
 
-Both guards can be bypassed with `--force` for break-glass scenarios. Never use `--force` during routine deploys.
+Guards 1 and 2 are bypassed with `--force`. Guard 3 is bypassed by the dedicated env var `DEPLOY_ALLOW_DIRTY=1` — kept separate from `--force` so the override is deliberate, audit-greppable, and `--force` doesn't accidentally subsume the dirty-tree check:
+
+```bash
+# Break-glass: deploy with uncommitted changes (NEVER for routine deploys)
+DEPLOY_ALLOW_DIRTY=1 scripts/deploy.sh
+```
+
+When the override is used, the script logs the dirty paths to stderr under a `⚠️  DEPLOY_ALLOW_DIRTY=1` banner so the deviation is visible in the deploy transcript. Never use `--force` or `DEPLOY_ALLOW_DIRTY=1` during routine deploys.
 
 Cloudflare cache purge runs when `CF_API_TOKEN` and `CF_ZONE_ID` are set in the environment. `CF_API_TOKEN` is sourced automatically by `scripts/op-preflight.sh --mode deploy` (or `--mode all`) from the shared "All Domains — Cache Purge API token" 1Password item — no `op read` needed in your shell. `CF_ZONE_ID` is per-repo; each downstream consumer sets its own zone ID (e.g., in the repo's bootstrap or as a hardcoded value in its `scripts/deploy.sh` wrapper) since one CF token covers all domains but each domain has its own zone. Without both variables the purge step no-ops with a clear log line.
 
--- a/scripts/ci/check_mktemp_portability
+++ b/scripts/ci/check_mktemp_portability
@@ -86,8 +86,9 @@ VIOLATION_LIST=""
 
 # The pattern targets `mktemp` followed by zero or more whitespace-
 # separated short-option clusters, ending in a cluster that contains
-# `-t` (either bare `-t` OR a compact combined form like `-dt`),
-# followed by a bare-token prefix arg.
+# `-t` (bare `-t` OR a compact combined form like `-dt`), followed
+# by a prefix-like argument — bare token, quoted string, OR variable
+# expansion.
 #
 # The cluster regex `-[a-zA-Z]*t` matches:
 #   - bare `-t`                  (zero leading letters, then 't')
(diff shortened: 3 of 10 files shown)


nathanpayne-codex (changes requested):
Changes requested on `de04eb3`.

The new portability guard misses compact short-option form `mktemp -dt <prefix>`, even though the script header lists that as a BAD example. I reproduced this by adding a tracked shell script containing `mktemp -dt mergepath-bad` in a temporary repo copy; `./scripts/ci/check_mktemp_portability` still printed PASS and exited 0.

Because this PR is adding the structural guard meant to prevent the Linux mktemp regression from returning, the guard should catch both `mktemp -d -t prefix` and `mktemp -dt prefix`, with a regression case.

nathanpayne-codex (changes requested):
Changes requested on `5a1528b`.

The compact `-dt` regex itself is present, and `bash tests/test_check_mktemp_portability.sh` passes, but the required `lint` job is still red on this head. `scripts/ci/check_mktemp_portability` now scans the newly added tracked regression test file and flags its fixture lines (`mktemp -t`, `mktemp -d -t`, `mktemp -dt`, `mktemp -Pdt`) before it can run the regression suite. I reproduced the same failure locally with `./scripts/ci/check_mktemp_portability`.

Please make the repo-wide scan ignore the test fixture file, or otherwise keep fixture-mode bad examples out of the production scan, then confirm the required `lint` jobs pass.

nathanpayne-codex (changes requested):
Changes requested on `abd45b1`.

The r3 self-exclusion fixes the previous blocker: the paired test file is no longer scanned as a production violation, and the existing regression suite plus repo-wide `check_mktemp_portability` pass. Two current-head blockers remain:

1. `scripts/ci/check_mktemp_portability` still misses quoted or variable BSD `-t` prefixes. I ran an isolated fixture containing `mktemp -t "mergepath-bad"`, `mktemp -t '$prefix'`, `mktemp -t "$prefix"`, and `mktemp -t ${prefix}` through `MERGEPATH_MKTEMP_SCAN_ROOT=... bash scripts/ci/check_mktemp_portability`; it printed `check_mktemp_portability: PASS` and exited 0. Those are the same non-portable `mktemp -t <prefix>` shape on GNU mktemp, so the guard can still be bypassed by quoting.

2. `scripts/deploy.sh` now has a Bash 3.2 failure on the normal no-extra-deploy-args path. On this machine (`GNU bash, version 3.2.57`), `scripts/deploy.sh --force --skip-build --skip-cf-purge` exits before invoking `op-firebase-deploy` with `DEPLOY_ARGS[@]: unbound variable`, because line 182 expands `"${DEPLOY_ARGS[@]}"` under `set -u` when the array is empty. The current test avoids that path by always passing `-- --only hosting`; please make the deploy invocation Bash-3.2-safe and add a no-trailing-args regression.

Validation run: `bash tests/test_check_mktemp_portability.sh`, `./scripts/ci/check_mktemp_portability`, `bash tests/test_deploy.sh`, `bash tests/test_mergepath_playground.sh`, `bash tests/test_sync_to_downstream.sh`, shell syntax checks, and `git diff --check origin/main...HEAD` all pass before the manual probes above.

nathanpayne-codex (changes requested):
Changes requested on `8de56cb`.

The r4 fixes close the two prior findings: the mktemp regex now catches quoted/variable prefix forms while leaving portable quoted templates alone, and `scripts/deploy.sh` uses the Bash 3.2-safe empty-array expansion for `DEPLOY_ARGS`. I verified the PR head with `bash tests/test_check_mktemp_portability.sh` (13/13), `./scripts/ci/check_mktemp_portability`, `bash tests/test_deploy.sh` (4/4), `bash tests/test_mergepath_playground.sh`, `bash tests/test_sync_to_downstream.sh`, `bash -n ...`, and `git diff --check origin/main...HEAD`.

The remaining blocker is the required `lint` job on the current PR merge ref. Reproducing `refs/pull/296/merge` locally, the newly wired live guard fails on `scripts/worktree-cleanup.sh` from current `main`:

```
scripts/worktree-cleanup.sh
217:GONE_FILE=$(mktemp -t wcleanup-gone.XXXXXX)
218:REC_FILE=$(mktemp -t wcleanup-rec.XXXXXX)
```

Because this PR makes `check_mktemp_portability` required in `repo_lint.yml`, the PR still cannot pass the protected merge-ref CI until that current-main violation is fixed, deliberately exempted/scoped, or otherwise removed from the merge ref. Please rebase/update after that surface is clean and confirm the live required `lint` job passes.

nathanpayne-codex (approved):
APPROVED on `ef6f34c`.

The r5 cross-PR fix closes the remaining blocker: `scripts/worktree-cleanup.sh` now uses portable `${TMPDIR:-/tmp}/wcleanup-*.XXXXXX` templates, with the inline comment explaining why this was a cross-PR cleanup from #298 after #296 wired the structural guard.

Validation run: `./scripts/ci/check_mktemp_portability` on the PR head, `./scripts/ci/check_mktemp_portability` on `refs/pull/296/merge`, `bash tests/test_worktree_cleanup.sh` (20/20), `bash tests/test_deploy.sh` (4/4), `bash tests/test_check_mktemp_portability.sh` (13/13), `bash tests/test_mergepath_playground.sh`, `bash tests/test_sync_to_downstream.sh`, `bash -n ...`, and `git diff --check origin/main...HEAD`. Live GitHub `lint` checks are green on the current head/merge ref.

I also checked the current CodeRabbit advisory threads. The `--help` range and Case 4 fixture-signing comments are small docs/test-hardening notes; I am not treating them as blockers for #286 because they do not change the deploy behavior or the newly required mktemp portability guard, and the affected suites pass.

nathanpayne-claude:

PR #303, opened 2026-05-16
feat: daily deferred-feedback rollup workflow + classifier (#299 v1)

nathanjohnpayne:
Implements the first slice of #299 — the daily cron that scans yesterday's merged PRs and surfaces bot review threads that were resolved without an associated fix commit or substantive reply, so the deferred-and-forgotten class of feedback stops accumulating into the next "closed-review backlog" incident (#234, #286, #287).

Architecture mirrors `scripts/sweep-unresolved-feedback/enumerate.sh` + `render.sh` but inverted: weekly sweep catches **unresolved** feedback on closed PRs; daily rollup catches **resolved** feedback on yesterday-merged PRs that wasn't actually addressed.

## What's in this slice

| File | Purpose |
|------|---------|
| `scripts/daily-feedback-rollup.sh` | Workhorse. Fetches merged PRs in 24h window, walks resolved threads, classifies via `[mergepath-resolve:<class>]` tag (preferred) or heuristics (`addressed-via-fix` / `addressed-via-reply` / `stale-head`), routes by severity into `substantive` / `polish` tracks, creates or appends to per-track rollup issues per self-throttling thresholds. `--dry-run` mode for offline inspection. |
| `scripts/lib/daily-feedback-rollup-helpers.sh` | Pure-function classifier helpers. Sourced by main script + unit tests. Bash 3.2 safe. |
| `.github/workflows/daily-feedback-rollup.yml` | Daily cron at `55 23 * * *` UTC (4:55pm Pacific) + `workflow_dispatch` with `since`/`until`/`dry_run` inputs. Auth via `REVIEWER_ASSIGNMENT_TOKEN` (same PAT weekly-feedback-sweep.yml uses). |
| `scripts/ci/check_daily_feedback_rollup` | Structural check. Layer 1: runs paired unit tests. Layer 2: runs `--dry-run` against a `gh`-shimmed scratch env with a 3-thread fixture, asserts classification + routing. |
| `tests/test_daily_feedback_rollup.sh` | 46 cases — `classify_severity` (Codex P0-P3 + CodeRabbit Major/Nitpick/Trivial/Outside-diff-range + body-head anchor), `severity_to_track` routing table, `item_id_for` (12 chars + deterministic + collision-resistant), `extract_tag_class` (canonical form + whitespace tolerance + first-tag wins + malformed rejection), `tag_class_action` (surface/skip with err-on-surface default), `is_agent_author` (override-at-call-time), `body_excerpt` (newline collapse + cap honored). |
[...]

PR #307, opened 2026-05-16
feat(rollup): dedupe daily rollup against prior-rollup triage state (#304)

### LINKED ISSUE #304: feat: daily rollup — triage-state persistence / dedupe-against-prior-rollups (#299 follow-up)
## Background

Follow-up to #299. v1 (PR #303) emits stable `<!-- mp-id:... -->` markers on every line item so dedupe can layer on without a data migration, but the dedupe logic itself is deferred. **Implication today:** an item left `[ ]` on yesterday's rollup WILL re-list in today's rollup, even if the upstream thread is still in `resolved-without-fix` state. The self-throttling logic (append to existing issue if ≥N unchecked items) bounds the issue-count blast radius until this lands.

## Proposal

Add a **dedupe pass** at the top of `scripts/daily-feedback-rollup.sh` that filters today's candidates against items already triaged in prior rollups.

### Triage-state semantics (per #299 v4 spec § Triage-state persistence)

Three checkbox markers, plain-markdown-editable on the rollup issue:

| Marker | State | Rollup re-listing |
|---|---|---|
| `- [ ]` | Open — not yet triaged | Re-list in tomorrow's rollup if upstream thread remains in scope |
| `- [x]` | Triaged: fix landed, won't-fix accepted, OR follow-up issue filed | Never re-listed |
| `- [~]` | Triaged: N/A or not-relevant | Never re-listed |

Also count as triaged:
- **Strikethrough** (`~~- [ ] ...~~`) — preserves item visibility while excluding from re-list
- **Follow-up issue reference** (a reply on the checklist line that includes a `#N` issue reference) — the follow-up issue is now where the work lives
- **Closed host issue** — the implicit "I'm closing this, the rest are won't-fix" signal

### Dedupe pass

Before emitting today's rollup, the workflow:

1. Fetches all open + recently-closed (last 14 days) issues with either `deferred-feedback-rollup` or `polish-feedback-rollup` label
2. Parses each line item's `<!-- mp-id:... -->` marker
3. Builds the set of already-triaged item IDs (from `[x]`, `[~]`, strikethrough, follow-up-link, or closed-host)
4. Filters today's candidate items to exclude already-triaged IDs
5. Surfaces "skipped: N items previously triaged" in the methodology footer

## Acceptance criteria

- [ ] `scripts/daily-feedback-rollup.sh` runs a dedupe pass against prior rollup issues
- [ ] Parses `[x]`, `[~]`, strikethrough, follow-up-link, and closed-host signals
- [ ] 14-day window cap (items older than that re-list because the host issue is presumed stale)
- [ ] New test cases in `tests/test_daily_feedback_rollup.sh` covering each dedupe state
- [ ] Methodology footer in the rollup body reports the dedupe-skipped count
- [ ] Manual smoke run via `workflow_dispatch` to validate dedupe on real prior-day rollup issues

## Out of scope

- Cross-repo dedupe (each repo's rollup remains independent in v2 too)
- Backfilling stable item IDs onto historical rollup issues (the v1 PR already started emitting them; the dedupe pass just reads what's there going forward)
- Changing the v1 self-throttling logic (the append-to-existing-issue threshold is complementary to dedupe, not replaced by it)

## References

- Parent: #299 (v1 close-out at https://github.com/nathanjohnpayne/mergepath/issues/299#issuecomment-4465053596)
- v1 PR: #303
- v4 spec section: § Triage-state persistence + § Dedupe pass

nathanjohnpayne:
Closes #304.

Authoring-Agent: claude

## Summary

Adds the v2 dedupe pass to the daily-feedback rollup workflow shipped in #303. v1 emitted stable `<!-- mp-id:... -->` markers on every line item but did NOT consult prior rollups, so an item left `[ ]` yesterday re-listed today. This PR layers the dedupe-against-prior-rollups logic on top of the v1 generator without any data migration.

### What changed

- **New pure helpers** in `scripts/lib/daily-feedback-rollup-helpers.sh`:
  - `parse_triaged_ids_from_body <body>` → newline-separated mp-ids. Recognises `[x]`/`[X]`/`[~]`/`[-]` checkboxes, `~~...~~` strikethrough wrapping a bullet, and `#N` follow-up issue references on the same line.
  - `parse_all_ids_from_body <body>` → every mp-id regardless of triage state. Used for the implicit "closed host issue → all items triaged" branch.
- **Dedupe pass** in `scripts/daily-feedback-rollup.sh`, between the per-PR classification loop and the dry-run gate:
  - Fetches open + closed-in-14-day-window rollup issues for BOTH `deferred-feedback-rollup` and `polish-feedback-rollup` labels (cross-track scan — mp-ids are track-agnostic).
  - Builds a `TRIAGED_IDS_FILE` set; closed hosts contribute all their mp-ids, open hosts only contribute mp-ids on lines that carry a triage signal.
  - Filters both NDJSON streams (`SUBSTANTIVE_NDJSON` / `POLISH_NDJSON`) via a single `jq --slurpfile` pass per stream (avoids a per-line shell loop).
  - Increments `COUNT_DEDUP_SKIPPED` and surfaces it both in the stderr summary line and in the methodology footer of the rendered rollup body.
- **15 new unit tests** covering each triage-signal type, dedupe of duplicate mp-ids within a body, the URL-anchor false-positive guard (`pull/999#discussion_r1` must NOT count as `#N`), and a realistic mixed-state rollup-body fixture.
- **Integration smoke extended** in `scripts/ci/check_daily_feedback_rollup`: the `gh` shim now returns fake prior rollups carrying Thread A's mp-id as `[x]` (substantive) and Thread C's mp-id alongside a `#888` reference (polish). Post-dedupe NDJSON line count drops from 2 → 0 with `dedup=2` in the stderr summary, locking the end-to-end wiring.

### Behaviour summary

| Signal on prior rollup line | Re-listed today? |
|---|---|
| `- [ ]` (open) | yes (this is the v1 behaviour the PR preserves) |
| `- [x]` / `- [X]` | **no** (triaged) |
| `- [~]` / `- [-]` | **no** (N/A) |
| `~~- [ ] ... ~~` | **no** (visibility-preserved, excluded) |
| Same line carries `#789` | **no** (follow-up filed) |
| Host issue is closed | **no** (all items implicit won't-fix) |
| Host issue is open + older than 14 days | yes (stale host → re-list) |

### Design trade-offs (documented inline)

- **Per-item follow-up granularity.** The spec mentions reply-comment threads with `#N` references as a possible triage signal. We deliberately do NOT consume issue-comment replies — instead we require the user to drop the `#N` into the bullet text itself. This keeps the helper a pure body-string parser with no API I/O. The simpler interpretation is explicitly permitted by the #304 spec.
- **Cross-track scan.** mp-ids are deterministic on `<repo>#<pr>:<thread_id>` — they don't encode track. A thread that flipped from `Major` (substantive) to `Nitpick` (polish) due to a classifier update should still dedupe across tracks. We scan both labels accordingly.
- **Best-effort on dedupe-fetch failure.** Dedupe is an optimisation, not a correctness invariant. If `gh issue list` fails for one label, we warn and continue rather than fail-closed — the worst-case post-failure outcome is re-listing an already-triaged item (which is the pre-#304 baseline behaviour). This is opposite of the v1 post-create atomicity gate (which DOES fail-closed because partial post is worse than no post).

### Correctness
- The triage-signal regex correctly handles the URL-anchor false-positive case (`pull/999#discussion_r1` contains `#discussion` not `#N`, and the digit-only `#[0-9]+` requirement excludes it). Unit test #8 locks this.
- `parse_all_ids_from_body` for closed hosts means every item on a closed rollup is dedupe-skipped going forward — this matches the spec's implicit-won't-fix semantics.
- mp-id de-duplication happens inside the helpers (via `awk '!seen[$0]++'`) AND again in the main script (via `sort -u`), so a single mp-id appearing on the v1 throttling-append path (which can write the same item to one issue twice) doesn't double-count.
- The 14-day window is enforced server-side via `closed:>=$DEDUP_CUTOFF` in the search query, plus a `MAX_PRIOR_ROLLUPS_PER_LABEL` (default 50) safety cap.

### Regression risk
- Pre-existing classifier behaviour is untouched: every test that was green before (46 unit + 5-case integration smoke) is still green. The new dedupe step is purely additive — it runs AFTER classification and only REMOVES items, never alters or adds them.
- The dry-run path now also exercises the dedupe pass (which calls `gh issue list`). The shim was extended to handle that; production `--dry-run` against a real repo will issue read-only `gh issue list` calls, which is consistent with the read-only-in-dry-run contract.
- I considered gating dedupe on `--dry-run` (i.e. don't fetch prior rollups in dry-run mode) but rejected that — operators inspecting `--dry-run` output expect it to match what a real run would post.

### Style
- Bash 3.2 compatible (no associative arrays, no `mapfile`, no `\b` regex). The earlier draft used `\b` in awk which isn't POSIX-portable; the final regex uses an explicit `(^|[ \t(\[,;])` boundary.
- POSIX character classes for grep/awk regex consistency (`[[:space:]]` not `\s`) — matches the codebase convention enforced by `check_mktemp_portability`-style audits.
- Functions documented with header comments explaining the trade-off, signal matrix, and Bash 3.2 portability notes.

### Test coverage
- 15 new unit tests covering every triage-signal type from the #304 spec, plus the URL-anchor negative case, the dedupe-within-helper guarantee, and a realistic mixed-state rollup fragment.
- Integration smoke extended with a fake prior-rollup fixture that exercises BOTH the substantive `[x]` path AND the polish `#N` follow-up path. Asserts the dedup counter surfaces in the stderr summary.

### Security
- No new credentials, no new write paths, no new shell-injection surfaces. The `gh issue list` calls are read-only and use the same `GH_TOKEN` as the existing PR-list call. The jq invocations use `--arg`/`--slurpfile` so no shell variables are interpolated into jq programs.
- No new dependencies (still `gh`, `jq`, `shasum`).

## Test plan

- [x] `bash tests/test_daily_feedback_rollup.sh` → 61 PASS (was 46; +15 dedupe cases)
- [x] `bash scripts/ci/check_daily_feedback_rollup` → PASS (unit + integration smoke + dedupe)
- [x] `bash scripts/ci/check_ci_scripts_wired` → PASS
- [x] `bash scripts/ci/check_mktemp_portability` → PASS
- [x] `bash scripts/ci/check_sync_manifest` → PASS
- [x] `bash -n` syntax check on both shell files
- [ ] Live `workflow_dispatch` of `daily-feedback-rollup.yml` against this repo's prior rollup issues (post-merge; documented in #304 AC)

### DIFF
--- a/scripts/ci/check_daily_feedback_rollup
+++ b/scripts/ci/check_daily_feedback_rollup
@@ -212,6 +212,41 @@ cat >"$FIXTURES/threads.json" <<'JSON'
 }
 JSON
 
+# ----- Fake prior rollup body (#304 dedupe pass) -----
+#
+# Thread A's stable mp-id is the SHA1[1:12] of "owner/repo#999:PRT_thread_a"
+# — we hardcode it here so the dedupe assertion is reproducible. If
+# item_id_for ever changes its hashing strategy, this fixture must be
+# refreshed in lockstep (and the unit tests would catch that first).
+THREAD_A_MPID="aa2b2e31d8f9"
+# Thread C is referenced as a follow-up issue `#888` on a different
+# rollup line to exercise the `#N` triage signal end-to-end.
+THREAD_C_MPID="1d9cdc0a026f"
+
+cat >"$FIXTURES/prior-substantive-rollups.json" <<JSON
+[
+  {
+    "number": 7001,
+    "state": "OPEN",
+    "body": "## owner/repo#900 (merged 2026-05-14, fix: prior PR)\n- [x] [scripts/old.sh:1](https://github.com/owner/repo/pull/900#discussion_r99) — \`chatgpt-codex-connector[bot]\` P1 [untagged]: \"fix landed in commit\" <!-- mp-id:${THREAD_A_MPID} -->\n- [ ] [scripts/another.sh:5](https://github.com/owner/repo/pull/900#discussion_r100) — \`coderabbitai[bot]\` Major [untagged]: \"still open elsewhere\" <!-- mp-id:9999deadbeef -->"
+  }
+]
+JSON
+
+cat >"$FIXTURES/prior-polish-rollups.json" <<JSON
+[
+  {
+    "number": 7002,
+    "state": "OPEN",
+    "body": "## owner/repo#901 (merged 2026-05-13, fix: polish prior)\n- [ ] [scripts/x.sh:1](https://github.com/owner/repo/pull/901#discussion_r200) — \`coderabbitai[bot]\` Nitpick [untagged]: \"polish, filed #888\" <!-- mp-id:${THREAD_C_MPID} -->"
+  }
+]
+JSON
+
+cat >"$FIXTURES/prior-empty.json" <<'JSON'
+[]
+JSON
+
 # ----- gh shim — dispatches on argv -----
 cat >"$BIN/gh" <<STUB
 #!/usr/bin/env bash
@@ -232,6 +267,41 @@ case "\$1" in
     # gh api graphql -f query='...' -F owner=... -F name=... -F pr=999
     # (we don't introspect the query — just return the fixture)
     cat "$FIXTURES/threads.json"
+    exit 0
+    ;;
+  issue)
+    # gh issue list --repo ... --label <label> --state all|open --search ... --limit ... --json number,state,body
+    # The dedupe pass (mergepath#304) calls this twice per track:
+    # once with --state all + closed:>= search, once with --state open.
+    # Dispatch on the --label value so substantive vs polish prior
+    # rollups can be different (and so an unrelated label returns []).
+    label=""
+    state=""
+    for ((i=1; i<=\$#; i++)); do
+      arg="\${!i}"
+      if [ "\$arg" = "--label" ]; then
+        next_i=\$((i + 1))
+        label="\${!next_i}"
+      elif [ "\$arg" = "--state" ]; then
+        next_i=\$((i + 1))
+        state="\${!next_i}"
+      fi
+    done
+    case "\$label" in
+      deferred-feedback-rollup)
+        # Substantive label: return prior rollup ONLY on the --state all
+        # call (which the dedupe pass uses for closed-in-window). The
+        # second --state open call returns the same OPEN rollup so the
+        # dedupe pass picks it up regardless of which branch hits.
+        cat "$FIXTURES/prior-substantive-rollups.json"
+        ;;
+      polish-feedback-rollup)
+        cat "$FIXTURES/prior-polish-rollups.json"
+        ;;
+      *)
+        cat "$FIXTURES/prior-empty.json"
+        ;;
+    esac
     exit 0
     ;;
   *)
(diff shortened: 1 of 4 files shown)


nathanpayne-codex (approved):
APPROVED

Reviewed head `e1b215b3157af3caded5ec5fc50f811f07fcf0a4`.

Validation passed on the merge ref: `git diff --check`, `bash -n` on touched shell files, `bash tests/test_daily_feedback_rollup.sh`, `bash scripts/ci/check_daily_feedback_rollup`, `bash scripts/ci/check_ci_scripts_wired`, `bash scripts/ci/check_mktemp_portability`, and `bash scripts/ci/check_sync_manifest`.

Current-head comments and review threads checked; no blocking findings.

PR #319, opened 2026-05-17
fix(propagation): export_consumer_facts yq filter rejected by mikefarah/yq — broke every templated render with non-empty facts

nathanjohnpayne:
**Filed by:** `nathanpayne-claude` — hotfix for the Phase D rollout (mergepath#250).

## Summary

`scripts/sync-to-downstream.sh`'s `export_consumer_facts()` used a yq `if (tag == "!!seq") then ... else ... end` form that mikefarah/yq v4 rejects at the lexer. The error went to stderr (dropped by process substitution), so the function emitted NO `MERGEPATH_FACT_*` exports and the lib (#313) rendered every templated entry as if the consumer had no frameworks.

**Phase D's swipewatch canary fell-positive** because swipewatch's declared `frameworks: []` produces identical output whether facts loaded or not. The bug surfaced on the second consumer (`device-platform-reporting`, `frameworks: [react]`) — the rendered `eslint.config.js` lacked the React block entirely.

## Fix

Switch to the documented mikefarah idiom for type-branching:

```yq
.value | (select(tag == "!!seq") | join(" ")) // (.value | tostring)
```

`select` filters on the yaml tag; on a !!seq the join wins, otherwise `//` falls back to `tostring`. Verified against all four real fact-value shapes (multi-element list, single-element list, empty list, scalar) plus the no-facts-block consumer case.

## Regression coverage

- `tests/test_export_consumer_facts.sh` — 6 cases, 4 verifying yq filter output across value-type shapes, 2 grep-asserting the lib script contains the new idiom and does NOT contain the broken form
- `scripts/ci/check_export_consumer_facts` — CI wrapper
- Wired into `repo_lint.yml` between `check_template_substitution` and `check_sync_manifest`
- In-source comment on `export_consumer_facts` documents the incident

## Self-Review

- [x] **Correctness:** new yq filter verified empirically against the live manifest for all 6 ESLint-rollout consumers. Each produces the expected facts shape (`frameworks <TAB> react typescript` for matchline, etc.).
- [x] **Regression risk:** scoped change — one function in one script. The lib's contract is unchanged; only the yq extraction was broken. The 6-case test pins the correct idiom (Cases 5 + 6 grep-assert the script's source) so a future contributor can't silently revert.
- [x] **Style and conventions:** bash 3.2 portable. Test follows the same fixture-runner pattern as `tests/test_export_consumer_facts.sh`'s siblings. CI wrapper mirrors `scripts/ci/check_sync_overrides`. Wired into the workflow in the natural ordering (after the other lib + sync-related checks).
- [x] **Test coverage:** 6 cases covering all four value-type shapes (multi-seq, single-seq, empty seq, scalar) plus the no-facts case plus the source-code grep assertions.
- [x] **Security and dependency hygiene:** no new dependencies. The yq filter operates on the manifest's content only (no shell-out, no eval). Failure mode of the OLD filter was silent + permissive (rendered to baseline) which is technically safer than failing-closed but masked the bug for a full canary cycle — the new test's Case 5/6 grep assertions ensure the filter syntax stays correct.

## Phase D follow-up after this lands

- swipewatch (PR #53, merged): unaffected, no action needed
- device-platform-reporting (PR #82, OPEN with WRONG rendered output): close + re-render after this hotfix merges
- Remaining 4 consumers (fanout): proceed normally once hotfix lands

### DIFF
--- a/.github/workflows/repo_lint.yml
+++ b/.github/workflows/repo_lint.yml
@@ -200,6 +200,16 @@ jobs:
         # independently of integration.
         run: ./scripts/ci/check_template_substitution
 
+      - name: check_export_consumer_facts
+        # Exercises tests/test_export_consumer_facts.sh — regression-
+        # guards the yq filter inside scripts/sync-to-downstream.sh's
+        # export_consumer_facts() against the mikefarah-yq-incompatible
+        # `if (tag == "!!X") then ... else ... end` form that silently
+        # broke Phase D's first real-frameworks consumer (every
+        # templated render fell through to the "no frameworks"
+        # baseline because MERGEPATH_FACT_* was never exported).
+        run: ./scripts/ci/check_export_consumer_facts
+
       - name: check_coderabbit_config
         # Validates the template-level .coderabbit.yml stays on the
         # nit-suppressing profile (chill). Shares the yq install with
--- a/scripts/ci/check_export_consumer_facts
+++ b/scripts/ci/check_export_consumer_facts
@@ -1,0 +1,21 @@
+#!/usr/bin/env bash
+# scripts/ci/check_export_consumer_facts — CI entry point for the
+# tests/test_export_consumer_facts.sh regression test.
+#
+# Guards the yq filter inside scripts/sync-to-downstream.sh's
+# export_consumer_facts() against the mikefarah-yq-incompatible
+# if/then/else form that silently broke Phase D's first
+# real-frameworks consumer (device-platform-reporting) — see the
+# test header for the full incident.
+
+set -euo pipefail
+
+REPO_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd)"
+TEST_SCRIPT="$REPO_ROOT/tests/test_export_consumer_facts.sh"
+
+if [ ! -f "$TEST_SCRIPT" ]; then
+  echo "check_export_consumer_facts: ERROR (missing $TEST_SCRIPT)" >&2
+  exit 1
+fi
+
+bash "$TEST_SCRIPT"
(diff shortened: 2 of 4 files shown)


nathanjohnpayne:
@codex review

nathanpayne-claude (commented):
**Filed by:** `nathanpayne-claude` — reviewer self-review.

This is a single-function hotfix for a real bug found mid-Phase-D rollout. The bug's failure mode (silent fall-through to JS-baseline render) was masked by the swipewatch canary because swipewatch's `frameworks: []` produces identical output regardless of whether facts loaded. The dpr propagation PR (#82) surfaced the real failure on the second consumer.

**What I'd flag fresh:**

1. **Test Cases 5 + 6 are grep-assertions on the lib's source code.** Pragmatic given the function isn't independently testable without sourcing the whole 38K-token script, but feels slightly meta. Alternative would be to extract the yq filter into a separate `scripts/lib/export-consumer-facts.sh` for proper unit testing — larger refactor, not urgent.

2. **The fix is empirically correct against mikefarah/yq v4.53.2** (the version the test runs against) but yq's syntax has churned across major versions. Worth confirming the select+// idiom works on whatever yq version CI's runner has — that's what `check_export_consumer_facts` in repo_lint.yml exercises on every PR.

3. **No retroactive fix for in-flight consumer PRs.** This PR closes the bug going forward; the dpr propagation PR (#82) will need to be closed + re-rendered manually after this hotfix lands. Documented in PR body's "Phase D follow-up" section. Could mention in #82 directly.

PR is small (4 files, ~200 lines) and under-threshold, so Codex 4a + my reviewer approval should clear the merge gate. No need for 4b unless the bot finds anything.

PR #321, opened 2026-05-17
fix(propagation): 3 check_* wrappers SKIP on consumer-side when mergepath prereqs absent (Phase D 4b CHANGES_REQUESTED)

nathanjohnpayne:
**Filed by:** `nathanpayne-claude`.

## Summary

nathanpayne-codex's Phase D 4b CHANGES_REQUESTED on dpr#83 (round 2) caught an architectural gap from #319 + #320: three check_* wrappers under `scripts/ci/` get propagated to consumers as part of the kit, but their target artifacts are mergepath-internal:

- `check_template_substitution` → tests `scripts/lib/template-substitution.sh` (mergepath-only)
- `check_export_consumer_facts` → tests `scripts/sync-to-downstream.sh`'s yq filter (mergepath-only)
- `check_sync_manifest` → validates `.mergepath-sync.yml` (deliberately not propagated per #268)

All three fail-loudly on consumers (`check_sync_manifest` has had this behavior since day 1; just wasn't a problem because consumer CI's `repo_lint.yml` is consumer-specific and never wired the check). nathanpayne-codex's ad-hoc 4b run from a consumer checkout hit it.

## Fix

Each wrapper now detects the absence of its mergepath-internal prereq and exits `0` with a clear `SKIP` diagnostic before attempting any work. On mergepath itself the prereq is present and tests run normally.

## Why SKIP not "remove from kit"

scripts/ci/ is a kit with allow-extras semantics — no per-file exclusion without changing the kit model. SKIP-on-absence is the minimum-scope fix. Future check_* wrappers added under scripts/ci/ that need mergepath-internal prereqs should follow the same pattern.

## Verification

- mergepath checkout: all three pass as before (45 + 6 + 25 test cases respectively)
- Fixture consumer checkout (no scripts/lib/, no scripts/sync-to-downstream.sh, no .mergepath-sync.yml): all three exit 0 with explicit SKIP diagnostics

## Phase D follow-up after this lands

Push the 3 updated wrapper scripts to all 5 Phase D consumer PR branches (same pattern as the test-file push from #320 follow-up). Clears nathanpayne-codex's CHANGES_REQUESTED on dpr#83 + preempts the same blocker on the other 4 propagation PRs.

## Self-Review

- [x] **Correctness:** verified empirically against both mergepath checkout (pass) and fixture consumer checkout (skip). Each SKIP path emits a diagnostic naming the missing prereq.
- [x] **Regression risk:** only changes the error path. Happy path (mergepath checkout with all files present) is byte-equivalent to before. check_sync_manifest's existing FAIL-on-missing was semantically wrong (consumers always failed); flipping to SKIP corrects the long-standing behavior.
- [x] **Style and conventions:** each wrapper documents the rationale inline. Pattern is replicable for future check_* additions.
- [x] **Test coverage:** the underlying test files (test_template_substitution.sh, test_export_consumer_facts.sh) are unchanged. Wrapper SKIP path is exercised by the consumer-fixture verification above (not codified as a test case because the only thing to test is "does the wrapper exit 0 when prereq missing" — trivially demonstrable).
- [x] **Security and dependency hygiene:** no new dependencies. The SKIP path is loud (writes to stdout); an operator inspecting CI output sees explicitly that the check was skipped, not silently passed.

### DIFF
--- a/scripts/ci/check_export_consumer_facts
+++ b/scripts/ci/check_export_consumer_facts
@@ -12,6 +12,17 @@ set -euo pipefail
 
 REPO_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd)"
 TEST_SCRIPT="$REPO_ROOT/tests/test_export_consumer_facts.sh"
+SYNC_SCRIPT="$REPO_ROOT/scripts/sync-to-downstream.sh"
+
+# Skip when running in a consumer checkout — the script being
+# tested (scripts/sync-to-downstream.sh's export_consumer_facts
+# function) is mergepath-internal and not propagated. Same
+# rationale as check_template_substitution above; same Phase D 4b
+# review surfacing.
+if [ ! -f "$SYNC_SCRIPT" ]; then
+  echo "check_export_consumer_facts: SKIP (mergepath-only: scripts/sync-to-downstream.sh not present in this checkout)"
+  exit 0
+fi
 
 if [ ! -f "$TEST_SCRIPT" ]; then
   echo "check_export_consumer_facts: ERROR (missing $TEST_SCRIPT)" >&2
--- a/scripts/ci/check_export_consumer_facts
+++ b/scripts/ci/check_export_consumer_facts
@@ -14,14 +14,24 @@ REPO_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd)"
 TEST_SCRIPT="$REPO_ROOT/tests/test_export_consumer_facts.sh"
 SYNC_SCRIPT="$REPO_ROOT/scripts/sync-to-downstream.sh"
 
-# Skip when running in a consumer checkout — the script being
-# tested (scripts/sync-to-downstream.sh's export_consumer_facts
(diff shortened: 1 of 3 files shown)


nathanjohnpayne:
@codex review

nathanjohnpayne:
@codex review

nathanjohnpayne:
@codex review

nathanpayne-codex (approved):
Approved current head `1b3c7f4b6c864b8a21b4aa33e66cbde8c10844d0`.

Validated the three wrapper changes in both directions:
- Mergepath checkout: `check_export_consumer_facts`, `check_template_substitution`, and `check_sync_manifest` all pass.
- Consumer fixture without `.mergepath-sync.yml` / mergepath-only helpers: all three wrappers exit 0 with explicit SKIP diagnostics.
- Mergepath misconfig cases still fail closed when `.mergepath-sync.yml` is present but the canonical sync script, template lib, or manifest is missing.

Also ran `check_ci_scripts_wired` and `check_workflow_parsers`, and `scripts/coderabbit-wait.sh 321` cleared the current head with zero Potential issue markers.

PR #327, opened 2026-05-17
feat(eslint-template): fact-driven defaults for testing globals + JSX-in-.js + React Compiler (#322)

nathanjohnpayne:
Closes #322.

## Summary

Adds three new template facts to `examples/eslint.config.{js,cjs.js}` so consumers stop hand-editing the same 7 overrides into every propagation PR:

- **\`facts.testing\`** (closed vocab: \`jest | vitest | mocha | node | none\`) — emits the appropriate test-runner globals block.
- **\`facts.jsx_in_js: true\`** — extends the React file glob from \`**/*.{jsx,tsx}\` to \`**/*.{js,jsx,tsx}\` for repos that babel-/vite-transpile JSX in .js files.
- **\`facts.react_compiler: true\`** — when false (the default), disables the React Compiler advisory rules (\`react-hooks/set-state-in-effect\`, \`preserve-manual-memoization\`, \`refs\`, \`immutability\`).

All three default OFF so existing rendered output doesn't drift. Each consumer opts in via \`.mergepath-sync.yml\` when their ground truth requires it.

Also rolls in template-wide baseline improvements that every consumer ended up adding by hand during the Phase D fanout:
- \`.claude/worktrees/**\` added to default ignores
- \`no-unused-vars\` + \`@typescript-eslint/no-unused-vars\` default to \`^_\`-prefix convention
- \`no-empty: { allowEmptyCatch: true }\`
- \`@typescript-eslint/no-explicit-any: warn\`

## Why

Must land before the next ESLint policy change — otherwise that change ships as a churn diff fighting every consumer's overrides. See #250 Phase D fanout where 5 of 6 consumers added the same vitest globals block, \`^_\` pattern, etc.

## Manifest changes

Per the audit in the implementation plan, only two consumers needed new facts to match their current ground truth:
- **friends-and-family-billing**: \`testing: vitest\`, \`jsx_in_js: true\`
- **device-platform-reporting**: \`testing: jest\`, \`jsx_in_js: true\`

Other consumers (matchline, nathanpaynedotcom, tadlockpsychiatry, swipewatch) have no new facts because their current rendered configs match the new template defaults exactly (^_ pattern, no-explicit-any: warn, etc. were universal; test-globals blocks were absent in those four).

## Nested-conditional limitation

[...]

PR #333, opened 2026-05-17
fix(eslint-template): remove orphan disable block + `<<<` from CJS template

nathanjohnpayne:
## Summary

Caught by today's first \`scripts/sync-to-downstream.sh --sync-all\` run after the #327 wave merged: 2 of 8 consumers (\`friends-and-family-billing\`, \`swipewatch\`) failed render with \`unexpected \`<<<\` (no open \`>>> if\`)\`. Both are the CJS-template consumers. Root cause: the #327 round-3 codex P1 fix (inline React Compiler disables into the React blocks, delete the standalone block) landed correctly on \`examples/eslint.config.js\` but only partially on \`examples/eslint.config.cjs.js\` — only the \`// >>> if frameworks contains react\` opener was removed; the body + closing \`// <<<\` were left behind, producing both a parse error AND a latent unconditional rules block.

This PR deletes the orphan content from the CJS template (the templates were supposed to be byte-identical modulo import/export shape) and adds two guard tests so this drift class can't recur silently.

## Changes

- \`examples/eslint.config.cjs.js\` — delete the orphan rules block (lines 314-321) + the orphan \`// <<<\` (line 322). The pointer-comment block remains, matching ESM.
- \`tests/test_template_substitution.sh\`:
  - **13a**: marker-count parity check across both templates. ESM and CJS must each have balanced \`// >>> if\` / \`// <<<\` counts, AND the totals must match between the two files.
  - **13b**: render-success check on the actual \`examples/eslint.config.{js,cjs.js}\` files under every load-bearing fact combination from the manifest. Render-time parsing is the load-bearing proof.

## Why the existing tests didn't catch it

The previous tests (12a–12h) exercised the template lib against \*synthetic fixtures\*, not against \`examples/eslint.config.*.js\`. The orphan-\`<<<\` was in the real template but invisible to fixture-based tests. 13b closes that gap; 13a is the cheaper marker-count check that would have flagged the imbalance even without invoking the renderer.

## Test plan

- [x] \`bash tests/test_template_substitution.sh\` — 55/55 PASS (was 53/53; +2 new guard cases).
- [x] Marker counts equal: 22=22 (11 open + 11 close in each file).
- [ ] Post-merge: re-run \`scripts/sync-to-downstream.sh --sync-all\` to fan out to friends-and-family-billing and swipewatch (the 2 consumers that failed in the first round).

Authoring-Agent: claude

## Self-Review

- **Correctness:** The deleted lines exactly match the orphan content that has no opener in the template-substitution lib's marker grammar. Confirmed by diff against the ESM template (which is now structurally aligned with the CJS template post-fix).
- **Regression risk:** Low. The deleted lines were rendering an unconditional rules block referencing react-hooks/* — a latent correctness bug that would have broken consumer ESLint runs if the parse error hadn't blocked rendering entirely. Removing it brings CJS to parity with ESM's correct shape.
- **Style:** Matches existing test structure (\`PASS:\` / \`FAIL:\` output, \`reset_facts\` + sourced \`\$LIB\` pattern, render in subshell for fact isolation).
- **Test coverage:** Two new guards. 13a is a cheap text-grep check; 13b actually invokes the renderer to confirm parse-clean under real consumer facts. Either would have caught the orphan-\`<<<\` bug at the unit-test layer instead of in prod.
- **Security/dependency hygiene:** No new dependencies. No secrets. The fix removes content, doesn't add any.

### DIFF
--- a/examples/eslint.config.cjs.js
+++ b/examples/eslint.config.cjs.js
@@ -311,15 +311,6 @@ module.exports = [
 // don't reference react-hooks/* on files where the plugin isn't
 // loaded — codex P1 #327 round 3). The standalone block previously
 // at this position has been removed.
-  {
-    rules: {
-      "react-hooks/set-state-in-effect": "off",
-      "react-hooks/preserve-manual-memoization": "off",
-      "react-hooks/refs": "off",
-      "react-hooks/immutability": "off",
-    },
-  },
-// <<<
 
 // >>> if testing contains vitest
   // Vitest globals — applied to common test file patterns. Without
--- a/tests/test_template_substitution.sh
+++ b/tests/test_template_substitution.sh
@@ -693,6 +693,78 @@ react-hooks/set-state-in-effect: off
 expect_output "12h react_compiler=true does NOT suppress disable block (v1 limitation)" "$r" "react-hooks/set-state-in-effect: off"
 
 # ---------------------------------------------------------------------------
+# 13. ESM/CJS template parity: render the real examples/eslint.config.{js,
+#     cjs.js} files against every meaningful facts combination. The render
+#     itself must succeed (parse-level marker balance is the load-bearing
+#     check) AND the ESM and CJS rendered bodies must match modulo the
+#     import/export shape. Without these guards, an edit that drifts the
+#     CJS template's `>>> if` / `<<<` structure away from the ESM (e.g.
+#     deletes the opener but leaves an orphan closer — observed during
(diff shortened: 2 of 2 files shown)


nathanpayne-claude (approved):
Reviewer-identity approval. Under-threshold template fix (72 LOC, no `.github/**` touch). CodeRabbit cleared on c353dfe with 0 findings. Two new drift guards (13a marker-count parity + 13b render-success on real templates) prevent recurrence. 55/55 tests pass.

PR #389, opened 2026-05-23
fix(sync): propagate op-preflight token-mode tests

nathanjohnpayne:
Fixes the closure gap caught by the downstream sync canary on nathanjohnpayne/device-source-of-truth#95.

## Changes

- Add `tests/test_op_preflight_token_mode.sh` to `.mergepath-sync.yml` so `scripts/ci/check_op_preflight_contract` does not propagate without all required tests.
- Rename the pre-provisioned Firebase deploy integration fixture from `preflight-firebase-sa.json` to `preflight-cached-project-sa.json` so the test harness-owned file does not match the tempfile leak detector pattern.

Authoring-Agent: codex

## Self-Review
- Correctness: closes the missing-test propagation gap and preserves the same preflight/deploy test behavior.
- Regression risk: low; manifest-only propagation addition plus a fixture filename that is internal to one integration test.
- Style: matches existing manifest comments and shell test naming.
- Test coverage: ran `scripts/ci/check_sync_manifest`, `scripts/ci/check_op_preflight_contract`, `MERGEPATH_RUN_INTEGRATION=1 scripts/ci/check_op_firebase_deploy_integration`, `scripts/ci/check_ci_scripts_wired`, and `git diff --check`.
- Security: no secrets or new credential paths; improves propagation of token-mode test coverage.

### DIFF
--- a/.mergepath-sync.yml
+++ b/.mergepath-sync.yml
@@ -377,9 +377,12 @@ paths:
     consumers: all
   # Pairs with scripts/op-preflight.sh + scripts/lib/preflight-helpers.sh
   # + scripts/ci/check_op_preflight_contract. The check wrapper requires
-  # both test files; same coupling pattern as the gh-as-author and
-  # codex-p1-gate suites above. See #282.
+  # all three test files; same coupling pattern as the gh-as-author and
+  # codex-p1-gate suites above. See #282/#374.
   - path: tests/test_op_preflight_check.sh
+    type: canonical
+    consumers: all
+  - path: tests/test_op_preflight_token_mode.sh
     type: canonical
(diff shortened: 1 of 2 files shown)


nathanpayne-codex (approved):
Approved. Small upstream closure fix for the sync canary findings: the op-preflight token-mode test now travels with check_op_preflight_contract, and the Firebase deploy integration fixture no longer matches the leak-detector pattern. Verified locally with check_sync_manifest, check_op_preflight_contract, check_op_firebase_deploy_integration, check_ci_scripts_wired, and git diff --check; CI and CodeRabbit are green on this HEAD.

Issue #402, opened 2026-06-01
Pre-existing: daily-rollup throttle/append path (gh issue comment) is dedupe-invisible


nathanjohnpayne:
## Pre-existing: daily-rollup throttle/append path is dedupe-invisible

Surfaced during the #400 / PR #401 review (Phase 4b, nathanpayne-codex). **Pre-existing**, not introduced by #401.

### The gap

`scripts/daily-feedback-rollup.sh` → `post_or_append`: when an existing rollup issue has `≥ throttle` unchecked items, today's findings are appended as a **comment** (`gh issue comment`). But the dedupe pass (`collect_triaged_ids_for_label`) reads prior rollups with `gh issue list --json number,state,body` — **body only, never comments**. So:

- mp-ids that land only in a throttle-path **comment** are never read by `parse_all_ids_from_body` / `parse_triaged_ids_from_body` on subsequent days.
- The daily window is strictly per-day (`merged:>=SINCE merged:<UNTIL`), so those PRs are never re-enumerated.
- Net: findings appended via the comment path are surfaced once but never tracked for dedupe/re-surfacing, and a human checking `[x]` in a comment is invisible to the throttle/dedupe loop.

This is the same *class* as #400 (a finding not in the dedupe-readable surface), but on a different (throttle) path.

### Why not fixed in PR #401

The natural fix (switch the append from `gh issue comment` to a `gh issue edit` body-merge) is a read-modify-write that introduces a **concurrency last-writer-wins** silent-loss vector the append-only comment path does not have (`schedule` + `workflow_dispatch` can interleave; no `concurrency:` guard on the workflow). Bundling it into #401 would trade one safe property for a new loss path. Deferred deliberately per maintainer decision.

### Options for the fix

1. Mirror appended mp-ids into the issue **body** via a single `gh issue edit` to a machine-managed `<!-- mp-overflow-ids -->` region, guarded by optimistic concurrency (re-read + compare `updatedAt` before write, fail-closed on mismatch), **and** add a `concurrency:` group to `.github/workflows/daily-feedback-rollup.yml`.
2. Keep comment-append but have the next run's dedupe also read recent comments on open rollups (widens the read; bounded by the 14-day window).

Both touch the synced `daily-feedback-rollup.sh` (and option 1 the synced `.yml`), so this needs the canary-first `--sync-all` propagation after merge.

Filed per REVIEW_POLICY.md § post-review observation.

PR #404, opened 2026-06-01
fix(rollup): paginate dedupe comment reads beyond the ~100 inline cap (#403)

### LINKED ISSUE #403: daily-rollup dedupe reads only first ~100 comments per rollup (#402 follow-up)
## `collect_triaged_ids_for_label` only reads the first ~100 comments per rollup

Found by Codex (P2) reviewing the #401 propagation canary (matchline#250); applies to the merged #402 fix in `scripts/daily-feedback-rollup.sh`.

### The gap

The #402 fix made the throttle/append path dedupe-visible by having `collect_triaged_ids_for_label` fetch `gh issue list --json …,comments` and parse body + comment bodies. But `gh`'s nested `comments` connection in a **list** query is **page-bounded (~100)** — it does not auto-paginate. So for a rollup issue that has accumulated **>100 comments** (many throttled-append days on one open issue), comment-resident mp-ids beyond the first ~100 are not read by the dedupe → they can be re-listed (or, for closed-host, not marked triaged). A narrower re-occurrence of #402 for high-comment-count issues.

### Realistic exposure

Bounded: a rollup issue needs >100 appended comments (>100 throttled days on the same open issue) to hit it. Rollups are normally triaged/closed well before that. P2, not a blocker — but real.

### Fix options

1. For issues whose `comments.totalCount > 100`, do a paginated per-issue fetch (`gh api --paginate repos/{owner}/{repo}/issues/{n}/comments`) instead of the inline list field.
2. Cap throttled appends per issue (roll over to a fresh rollup issue after N comments), so a single issue never exceeds the readable window.
3. Re-evaluate Option 1 from #402 (mirror appended mp-ids into a machine-managed body region with optimistic concurrency) — no comment pagination, but a read-modify-write.

### Minor nitpicks from the same review (fold in or dismiss)

- CodeRabbit Minor: `mktemp …-XXXXXX.md` suffix portability on BSD/macOS. NOTE: `check_mktemp_portability` passes and the suffix pattern predates this PR — likely a false positive; confirm.
- CodeRabbit Trivial: an unguarded `grep … | wc -l` under `set -euo pipefail` (a no-match `grep` exits 1 and trips the pipe). Add `|| true` for parity with the dedupe counts.

Touches the synced `daily-feedback-rollup.sh`, so the fix will need another canary-first `--sync-all`/per-commit propagation to the 8 consumers.

Filed per REVIEW_POLICY.md § post-review observation (surfaced during #401 propagation).

nathanjohnpayne:
## What & why

Follow-up to #401/#402, closes **#403**. The #402 fix made the throttle/append **comment** path dedupe-visible by folding comment bodies into `collect_triaged_ids_for_label` via the inline `gh issue list --json comments` field — but that nested field is GitHub **page-capped at ~100**. A rollup issue with >100 appended comments (heavy throttle use over time) would lose dedupe visibility for the later mp-ids — a narrower re-occurrence of #402. Codex flagged it (P2) on the #401 propagation canary, matchline#250.

## Change

`collect_triaged_ids_for_label` now uses the inline `comments` length only as a cheap **"are there any comments?"** gate; when an issue has comments it fetches **every** one via `gh api --paginate repos/{owner}/{repo}/issues/{n}/comments` (best-effort — warn + body-only on failure, consistent with the existing dedupe fail-soft), then parses body + all comments. The common **0-comment rollup skips the extra call**. The append mechanism is unchanged (still an append-only comment — no read-modify-write concurrency risk).

Also folded in the CodeRabbit nits from the same review:
- `|| true` on two test `grep | … | wc -l` pipelines for `set -euo pipefail` parity (clean assertion failure instead of a `set -e` abort).
- The mktemp-suffix Minor is a **non-issue** — `check_mktemp_portability` passes and CI runs GNU mktemp; left as-is.

## Verification (73 tests pass)

- #402 parse contract (unchanged): body + comments concatenation surfaces comment-resident triaged mp-ids (open host) and the full set (closed host).
- #403 regression guard: asserts the comment source is the **paginated** REST endpoint, not the capped inline field. (The fetch itself is best-effort gh-integration; the parse contract is the unit-tested part.)
- `check_daily_feedback_rollup` green (unit + dry-run smoke that runs `collect_triaged_ids` + #304 dedupe); `check_mktemp_portability` green.

## Post-merge

Script-only change to the synced `daily-feedback-rollup.sh` → propagates to the 8 consumers via canary-first per-commit propagation after merge.

Closes #403.

Authoring-Agent: claude

## Self-Review

- **Correctness**: Paginated REST fetch reads all comments regardless of the inline field's cap (verified `gh api --paginate` returns full comment sets). Best-effort: a per-issue fetch failure warns + degrades to body-only for that issue (doesn't abort — matches the dedupe's existing fail-soft posture; the all-or-nothing gate protects the post path, not the read-only dedupe). 0-comment rollups skip the call.
- **Regression risk**: Low. 48-line change, isolated to the dedupe read loop; the parse path is unchanged. Append mechanism unchanged (no concurrency exposure). Adds N best-effort API calls only for issues that actually have comments.
- **Style**: Matches the function's existing jq + fail-soft idioms; bash 3.2-safe.
- **Test coverage**: parse contract + paginated-source structural guard; test pipefail parity fixed.
- **Security/deps**: no new deps; read-only `gh api` GET; one extra paginated read per commented issue.

### DIFF
--- a/scripts/daily-feedback-rollup.sh
+++ b/scripts/daily-feedback-rollup.sh
@@ -614,16 +614,32 @@ collect_triaged_ids_for_label() {
   n=$(printf '%s' "$merged" | jq 'length')
   local x=0
   while [ "$x" -lt "$n" ]; do
-    local issue_state issue_text
+    local issue_state issue_num issue_body inline_comment_count comment_text issue_text
     issue_state=$(printf '%s' "$merged" | jq -r ".[$x].state")
+    issue_num=$(printf '%s' "$merged" | jq -r ".[$x].number")
+    issue_body=$(printf '%s' "$merged" | jq -r ".[$x].body // \"\"")
+    inline_comment_count=$(printf '%s' "$merged" | jq -r ".[$x].comments // [] | length")
     # Parse the BODY *and* all comment bodies (#402). The throttle path
     # appends today's findings as a COMMENT (append-only, atomic — we keep
     # that rather than a read-modify-write body edit that could clobber a
-    # concurrent run). So mp-ids and triage signals can live in comments;
-    # folding them in here makes the comment-append path dedupe-visible.
-    issue_text=$(printf '%s' "$merged" | jq -r "
-      .[$x] | ((.body // \"\") + \"\n\" + ((.comments // []) | map(.body // \"\") | join(\"\n\")))
-    ")
+    # concurrent run). The inline `comments` list field is page-capped
+    # (~100, #403), so when an issue has any comments we paginate the REST
+    # endpoint to read EVERY one — otherwise a rollup with >100 appended
+    # comments would lose dedupe visibility for the later mp-ids. The
+    # inline count is only a cheap "are there comments?" gate so the
+    # common (0-comment) rollup skips the extra call.
+    comment_text=""
+    if [ "$inline_comment_count" -gt 0 ]; then
+      if ! comment_text=$(gh api --paginate \
+           "repos/$OWNER/$NAME/issues/$issue_num/comments" \
+           --jq '.[].body // empty' 2>/dev/null); then
+        echo "daily-feedback-rollup: WARN dedupe: could not paginate comments for #$issue_num (best-effort, continuing)" >&2
+        DEDUP_FETCH_FAILURES=$((DEDUP_FETCH_FAILURES + 1))
+        comment_text=""
+      fi
+    fi
+    issue_text="${issue_body}
+${comment_text}"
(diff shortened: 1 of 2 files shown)


nathanpayne-claude (approved):
Reviewer approval (nathanpayne-claude). Under-threshold (48 lines, no protected paths). Closes #403 — dedupe now paginates comments past the ~100 inline cap; append mechanism unchanged (no concurrency risk). CodeRabbit clear; 73 tests + check_daily_feedback_rollup + check_mktemp_portability green.

PR #430, opened 2026-06-09
fix(eslint-template): extend TS unused-vars tuning to .astro frontmatter

nathanjohnpayne:
## Summary

The ESLint template's `@typescript-eslint/no-unused-vars` (and `no-explicit-any`) tuning block was scoped to `files: ["**/*.{ts,tsx}"]`, so `.astro` frontmatter never received the `^_` ignore-pattern convention — even though astro frontmatter is TypeScript that `@typescript-eslint` lints via the astro parser. The rule still fired (from the recommended preset), just *without* the ignore patterns, so intentionally-unused `_`-prefixed locals errored on a full `eslint .`.

This extends the glob to `["**/*.{ts,tsx,astro}"]` in **both** the ESM (`examples/eslint.config.js`) and CJS (`examples/eslint.config.cjs.js`) template variants, and adds a regression guard.

Authoring-Agent: claude

## Why it lives in the typescript-gated block

The `.astro` extension is added to the **existing `typescript`-gated block**, not a new `astro`-gated one. That keeps the `@typescript-eslint/*` rule IDs co-gated with the `typescript` fact — the only place the `@typescript-eslint` plugin is imported — so they can never render into a config where the plugin is absent (the rule-without-plugin break fixed in #327). For a TS consumer with no `.astro` files (matchline, tadlockpsychiatry) the extra extension matches nothing and is inert: behavior-neutral, the rendered config text gains only the `,astro` token plus the explanatory comment.

(The alternative — an `astro`-gated block referencing `@typescript-eslint/*` — would reintroduce the #327 risk for any future `astro`-without-`typescript` consumer, and v1 templating can't express the `astro && typescript` compound gate.)

## Verification

Rendered both templates with each ESM consumer's real manifest facts via `scripts/lib/manifest-fact-helpers.sh` + `scripts/lib/template-substitution.sh`:

- `render(old, nathanpaynedotcom)` is **byte-identical** to its committed `eslint.config.js` (render harness faithful).
- `render(new, nathanpaynedotcom)` → `npx eslint .` reports **0 errors** (the two latent `_slug` / `_err` errors are gone; only the pre-existing `no-explicit-any` *warnings* in a test remain). This also confirms the `astro/flat-recommended` block (spread later) does not override the setting.
- matchline / tadlockpsychiatry renders differ only by the comment + `,astro` glob token — behavior-neutral (no `.astro` files).
- `tests/test_template_substitution.sh`: **59/59** pass (Case 14 added — below). `tests/test_eslint_policy_check.sh`: **10/10**. Both templates pass `node --check`.

## Self-Review

- **Correctness:** Additive glob change; `.astro` frontmatter is parsed as TS by the astro parser, so the rule applies cleanly — proven by the clean nathanpaynedotcom render-lint. The ESM and CJS edits are identical.
- **Regression risk:** Low. Inert for non-astro TS consumers. New **Case 14** in `test_template_substitution.sh` renders the *real* templates and asserts (a) astro+ts carries `**/*.{ts,tsx,astro}`, and (b) a consumer **without** the `typescript` fact never emits `@typescript-eslint/*` rule IDs — locking in the co-gating safety property. Marker-balance + ESM/CJS symmetry tests still pass.
- **Style:** Matches the file's heavily-commented house style; the comment explains both the `.astro` rationale and the co-gating decision inline.
- **Test coverage:** +4 assertions (Case 14a/14b across ESM + CJS), rendering the real templates.
- **Security/dependency hygiene:** None affected. Comment + one glob token + a shell test; no new dependencies, no credential surface.

## Downstream

After merge this propagates to **nathanpaynedotcom** (the only `astro` consumer) via `scripts/sync-to-downstream.sh`, regenerating its `eslint.config.js` and making `npm run lint` clean there. matchline / tadlockpsychiatry will show a behavior-neutral propagation diff (comment + glob token) on their next sync.

### DIFF
--- a/examples/eslint.config.cjs.js
+++ b/examples/eslint.config.cjs.js
@@ -191,8 +191,18 @@ module.exports = [
   // Both demotions were hand-added by every TS consumer during the
   // Phase D fanout (#250); folding into the template removes the
   // per-consumer churn.
+  //
+  // `.astro` is in the glob because astro frontmatter (the `---` fence)
+  // is TypeScript, linted by @typescript-eslint via the astro parser —
+  // without it the `^_` convention silently fails on astro consumers
+  // (the rule still fires from the recommended preset, just without
+  // these ignore patterns). Kept in this typescript-gated block, NOT a
+  // separate astro-gated one, so these `@typescript-eslint/*` rule IDs
+  // only render where the plugin is imported — cf. the #327 rule-
+  // without-plugin break. A TS consumer with no `.astro` files matches
+  // none, so the added extension is inert for them.
   {
-    files: ["**/*.{ts,tsx}"],
+    files: ["**/*.{ts,tsx,astro}"],
     rules: {
       "@typescript-eslint/no-unused-vars": ["error", {
         argsIgnorePattern: "^_",
--- a/examples/eslint.config.js
+++ b/examples/eslint.config.js
@@ -185,8 +185,18 @@ export default [
   // Both demotions were hand-added by every TS consumer during the
   // Phase D fanout (#250); folding into the template removes the
   // per-consumer churn.
+  //
+  // `.astro` is in the glob because astro frontmatter (the `---` fence)
+  // is TypeScript, linted by @typescript-eslint via the astro parser —
+  // without it the `^_` convention silently fails on astro consumers
+  // (the rule still fires from the recommended preset, just without
+  // these ignore patterns). Kept in this typescript-gated block, NOT a
+  // separate astro-gated one, so these `@typescript-eslint/*` rule IDs
+  // only render where the plugin is imported — cf. the #327 rule-
+  // without-plugin break. A TS consumer with no `.astro` files matches
+  // none, so the added extension is inert for them.
   {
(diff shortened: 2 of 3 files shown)


nathanpayne-claude (approved):
Reviewing as nathanpayne-claude (PR authored by nathanjohnpayne).

**Verified**
- **Correct:** rendered the new template with nathanpaynedotcom's real manifest facts → `eslint .` reports **0 errors** (the latent `_slug`/`_err` gone). Confirmed the astro `flat/recommended` block (spread later) does not override the TS `no-unused-vars` setting.
- **Safe co-gating:** the `.astro` extension lives in the `typescript`-gated block, so `@typescript-eslint/*` rule IDs never render where the plugin is absent — Case 14b locks this in. ESM/CJS edits are identical; marker-balance + symmetry tests pass (59/59).
- **Behavior-neutral** for non-astro TS consumers (matchline/tadlockpsychiatry): the glob matches no `.astro` files.
- CI green; CodeRabbit: no actionable comments. Under-threshold (58 lines, no protected paths) — no Phase 4 required.

Approving.

PR #459, opened 2026-06-16
fix(sync): fail closed in export_consumer_facts on yq parse error (#457)

### LINKED ISSUE #457: manifest-fact-helpers: export_consumer_facts fails open on yq parse error
## Summary

`export_consumer_facts()` in `scripts/lib/manifest-fact-helpers.sh` swallows
`yq` parse failures. The `yq` invocation runs inside a process substitution
feeding a `while` loop, so its exit code is discarded. On a malformed/unparseable
manifest, `yq` prints an error to stderr but the function still returns `rc=0`
with **no** `MERGEPATH_FACT_*` exported — it fails *open* instead of closed.

## Location

`scripts/lib/manifest-fact-helpers.sh`, the export block (~lines 52–63):

```bash
while IFS=$'\t' read -r key value; do
  ...
  export "MERGEPATH_FACT_$upper=$value"
done < <(MERGEPATH_CONSUMER_NAME="$consumer_name" yq -r '
  ... filter ...
' "$manifest")
```

## Root cause

The exit status of `while ... done < <(cmd)` is the status of the last `read`
(or `0` if the loop body ran), never the status of `cmd` inside the process
substitution. A `yq` failure is therefore masked.

## Impact

`scripts/workflow/verify-propagation-pr.sh` sources this helper from the trusted
mergepath checkout (`FACTS_HELPER="$MERGEPATH_DIR/scripts/lib/manifest-fact-helpers.sh"`)
and calls `export_consumer_facts` immediately before templated re-rendering. A
manifest that `yq` cannot parse yields empty facts silently, so templated
verification can proceed with empty/wrong `MERGEPATH_FACT_*` state instead of
failing closed:

- For consumers whose facts are empty anyway (e.g. `frameworks: []`) the
  re-render byte-compare still matches — the failure is fully masked.
- For consumers with real facts the re-render diverges and fails for the *wrong*
  reason (looks like render drift, not a parse error).

Either way a verification helper must fail closed on a parse error.

**Secondary (same fail-closed gap at the call site):** in
`verify-propagation-pr.sh` the render subshell is

```bash
render_rc=0
(
  export_consumer_facts "$CONSUMER_NAME" "$MANIFEST"
  template_substitution::render "$mp_source_abs"
) > "$rendered" 2> "$render_err" || render_rc=$?
```

The exit status of `( a; b )` is `b`'s status, so even after the helper is fixed
to `return 1`, a non-zero from `export_consumer_facts` would be masked by
`template_substitution::render`'s rc. Harden both: have the helper return
non-zero, and have the subshell propagate it (e.g. `set -e` in the subshell or an
explicit rc check on the export step).

## Suggested fix (helper)

Capture `yq`'s output into a variable, check its exit code, `return 1` on
failure, then iterate via a here-string. **Do not** convert to a pipe
(`yq | while`) — that subshells the loop and the `MERGEPATH_FACT_*` exports would
not survive into the caller's shell. Keep the exact `yq` filter string intact —
`tests/test_export_consumer_facts.sh` greps the file for
`select(tag == "!!seq") | join(" ")` and the absence of the broken
`if ... then ... else ... end` form — and add a regression case asserting
fail-closed behavior on malformed YAML.

```bash
local facts_tsv
if ! facts_tsv=$(MERGEPATH_CONSUMER_NAME="$consumer_name" yq -r '
  ... unchanged filter ...
' "$manifest"); then
  echo "export_consumer_facts: yq failed to parse $manifest (consumer=$consumer_name)" >&2
  return 1
fi
while IFS=$'\t' read -r key value; do
  [ -z "$key" ] && continue
  local upper
  upper=$(printf '%s' "$key" | tr '[:lower:]' '[:upper:]' | tr '-' '_')
  export "MERGEPATH_FACT_$upper=$value"
done <<< "$facts_tsv"
```

(A here-string adds a trailing newline; the existing `[ -z "$key" ] && continue`
already absorbs the empty trailing record, so empty-facts consumers behave as
before.)

## Provenance

Surfaced by a `nathanpayne-codex` external review on
[overridebroadway#86](https://github.com/nathanjohnpayne/overridebroadway/pull/86#pullrequestreview-4502170272)
(a mirror-sync PR). The bug is in the canonical helper here, so it is mirrored
into every consumer and is currently **live** on overridebroadway `main` (synced
to `mergepath@cc3487a`, which still carries this pattern). overridebroadway#86
itself was closed as superseded (it targeted the older `mergepath@165da95`); this
issue tracks the upstream fix that will propagate the correction to all consumers
on the next sync.

nathanjohnpayne:
Authoring-Agent: claude

## Summary
- Fixes #457: `export_consumer_facts()` in `scripts/lib/manifest-fact-helpers.sh` swallowed `yq` parse failures. The `yq` ran inside `while ... done < <(yq ...)`, so the loop's exit status (last `read`) masked yq's — a manifest yq couldn't parse exported **no** `MERGEPATH_FACT_*` yet returned `0` (**fail open**), letting the templated-render verifier proceed on empty/wrong facts.
- **Helper:** capture yq output into a variable, `return 1` on parse failure, iterate via a here-string. A `yq | while` pipe would subshell the loop and the `MERGEPATH_FACT_*` exports would not survive into the caller — so the here-string is deliberate. The yq filter string is **byte-unchanged** (the test's `select(tag == "!!seq") | join(" ")` shape guard and broken-form-absence guard still hold).
- **Call site** (`scripts/workflow/verify-propagation-pr.sh`): `export_consumer_facts ... || exit $?` so the render subshell propagates the fail-closed rc. `set -e` alone is unsafe — the subshell is the left operand of `|| render_rc=$?`, a context where bash suppresses `set -e` for the commands inside it, so the helper's non-zero would otherwise be masked by `template_substitution::render`'s rc.

No user-visible behavior change on the happy path; the success path still exports facts identically (verified against the live manifest: `matchline` → `MERGEPATH_FACT_FRAMEWORKS=react typescript`).

## Testing
- [x] Tests pass — `tests/test_export_consumer_facts.sh` now 9 cases incl. new **Case 8** (sources the lib, asserts non-zero rc + no exports on an unparseable manifest, with a precondition that the fixture is genuinely unparseable so it can't pass vacuously).
- [x] CI checks pass locally: `check_export_consumer_facts`, `check_verify_propagation_pr`, `check_workflow_verify_propagation_templated`, `check_mktemp_portability`.
- [x] Manual verification: success path unchanged; malformed manifest → rc=1, no facts exported.

## Self-Review
- [x] Correctness: matches #457's suggested helper fix and the secondary call-site fail-closed hardening.
- [x] Regression risk: filter string unchanged; here-string trailing newline absorbed by the existing empty-key guard, so empty-facts consumers behave as before.
- [x] Style and conventions: follows repository standards.
- [x] Test coverage: new fail-closed path covered; existing cases passing.
- [x] Security and dependency hygiene: no new deps; the change makes a verification helper fail **closed**.

Closes #457

### DIFF
--- a/scripts/lib/manifest-fact-helpers.sh
+++ b/scripts/lib/manifest-fact-helpers.sh
@@ -49,18 +49,35 @@ export_consumer_facts() {
     unset "$var"
   done
 
-  while IFS=$'\t' read -r key value; do
-    [ -z "$key" ] && continue
-    local upper
-    upper=$(printf '%s' "$key" | tr '[:lower:]' '[:upper:]' | tr '-' '_')
-    export "MERGEPATH_FACT_$upper=$value"
-  done < <(MERGEPATH_CONSUMER_NAME="$consumer_name" yq -r '
+  # Capture yq's output up front so a parse failure surfaces as a
+  # non-zero return (fail CLOSED). A `while ... done < <(yq ...)` process
+  # substitution masks yq's exit code — the loop's status is the last
+  # `read` (or 0 if the body ran), never yq's — so a malformed manifest
+  # would export no facts yet return 0 (fail OPEN), letting a verification
+  # caller proceed on empty/wrong MERGEPATH_FACT_* state. See #457.
+  local facts_tsv
+  if ! facts_tsv=$(MERGEPATH_CONSUMER_NAME="$consumer_name" yq -r '
     env(MERGEPATH_CONSUMER_NAME) as $cn
     | .consumers[] | select(.name == $cn) | .facts // {} | to_entries[]
     | .key + "\t"
       + ((.value | select(tag == "!!seq") | join(" "))
          // (.value | tostring))
-  ' "$manifest")
+  ' "$manifest"); then
+    echo "export_consumer_facts: yq failed to parse $manifest (consumer=$consumer_name)" >&2
+    return 1
+  fi
+
+  # Iterate via a here-string, NOT a `yq | while` pipe: a pipe subshells
+  # the loop body and the MERGEPATH_FACT_* exports would not survive into
+  # the caller's shell. The here-string adds a trailing newline, which the
+  # empty-key guard below absorbs, so empty-facts consumers behave exactly
+  # as before.
+  while IFS=$'\t' read -r key value; do
+    [ -z "$key" ] && continue
+    local upper
+    upper=$(printf '%s' "$key" | tr '[:lower:]' '[:upper:]' | tr '-' '_')
+    export "MERGEPATH_FACT_$upper=$value"
+  done <<< "$facts_tsv"
   # NOTE on the value-typed serialization: mikefarah/yq v4 does NOT
   # accept a top-level `if`-then-else expression in this filter
   # position (the lexer rejects `if` at column-after-pipe). An
--- a/scripts/workflow/verify-propagation-pr.sh
+++ b/scripts/workflow/verify-propagation-pr.sh
@@ -298,7 +298,13 @@ if [ "$TEMPLATED_SURFACE_ACTIVE" = "1" ] && [ -n "$CONSUMER_NAME" ]; then
       render_err=$(mktemp "${TMPDIR:-/tmp}/verify-prop-render-err.XXXXXX")
       render_rc=0
       (
-        export_consumer_facts "$CONSUMER_NAME" "$MANIFEST"
+        # `|| exit $?` explicitly propagates a fail-closed export rc as the
+        # subshell's exit status. Relying on `set -e` alone is unsafe here:
+        # the subshell is the left operand of `|| render_rc=$?`, a context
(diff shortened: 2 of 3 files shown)


nathanpayne-claude (approved):
Reviewer pass (nathanpayne-claude), under-threshold (~69 LOC, no protected paths).

Verified the fail-closed fix: capturing yq output into a variable surfaces the parse-error rc that the prior `done < <(yq ...)` masked; here-string iteration (not a pipe) keeps the MERGEPATH_FACT_* exports in the caller's shell; and the `|| exit $?` at the verify-propagation-pr.sh call site propagates the rc past the set-e-in-|| subshell context. yq filter string is byte-unchanged (filter-shape regression guards intact). New Case 8 has a precondition that the malformed fixture is genuinely unparseable, so it can't pass vacuously. CodeRabbit StatusContext clean. LGTM.

Issue #471, opened 2026-06-16
Templated propagation: preserve source mode + switch protected-path globs to segment-local semantics


nathanpayne-claude:
Source rollup: nathanjohnpayne/mergepath#456

## Decision (2026-06-16)

Both questions from the original issue have been decided. This issue is now agent-ready.

---

### 1. Templated output modes → preserve source template's git mode

**Decision:** Rendered templated tree entries must preserve the git mode of the source template in `mergepath@<sha>`. The current `100644` hardcode is wrong for executable templates.

**Evidence:**
- `scripts/workflow/verify-propagation-pr.sh` hardcodes `100644 blob` as the expected rendered entry mode.
- `scripts/sync-to-downstream.sh` already treats mode as part of faithful propagation.
- Canonical/kit propagation mirrors executable bits; the verifier must match.

**Implementation notes:**
- The verifier (`verify-propagation-pr.sh`) must replace the hardcoded `100644` check with a `git ls-tree mergepath@<sha> -- <template-path>` lookup to get the actual expected mode from the source tree. Removing the constant without adding the lookup leaves verification hollow.
- Both the materializer (`sync-to-downstream.sh`) and verifier must be updated together so they stay in contract.

---

### 2. Protected-path globs → segment-local `*`, recursive `**`

**Decision:** `match_protected_paths.sh` must switch from Bash `[[ ]]` pattern semantics to repo-relative path glob semantics where `*` and `?` do not cross `/`, and `**` may cross directories.

**Evidence:**
- Current Bash `[[ $path == $pattern ]]` makes `src/*.js` match `src/nested/file.js` — too broad for a policy knob.
- The Playground's slash-aware matcher in `mergepath/playground/index.html` already implements the intended semantics. Shell and Playground must be in parity.

**Implementation notes:**
- `match_protected_paths.sh` will need a rewrite — Bash `[[ ]]` can't express the segment-local constraint natively. Use `python3 -c "import fnmatch; ..."` or a small inline helper that splits on `/` and matches segment-by-segment. Do not leave Bash `[[ ]]` as the implementation with a comment saying it's approximate.
- `.github/**` is the most-used pattern; regression tests must cover it explicitly.
- Playground JS logic is the reference implementation — shell output must match it for identical inputs.

---

## Work items

- [ ] `verify-propagation-pr.sh` — replace `100644` hardcode with `git ls-tree mergepath@<sha>` mode lookup
[...]

Issue #456, opened 2026-06-15
Unresolved reviewer feedback backlog — 2026-06-15


nathanpayne-claude:
# Unresolved reviewer feedback backlog

Automated weekly sweep — see #236. Last run **2026-06-15** found **719** unresolved review threads on closed PRs across configured target repos.

**Counts below are raw (pre-validation).** A separate validation pass (see issue rubric in #236) is required to classify each item as VALID / ALREADY-FIXED / REJECTED / MOOT / AMBIGUOUS before treating the count as actionable backlog.

Lookback window: closed PRs in the last `SWEEP_LOOKBACK_DAYS` days (default 90).

## By repo

- **nathanjohnpayne/device-platform-reporting** — 27 items
- **nathanjohnpayne/device-source-of-truth** — 93 items
- **nathanjohnpayne/friends-and-family-billing** — 54 items
- **nathanjohnpayne/matchline** — 158 items
- **nathanjohnpayne/mergepath** — 55 items
- **nathanjohnpayne/nathanpaynedotcom** — 176 items
- **nathanjohnpayne/overridebroadway** — 59 items
- **nathanjohnpayne/swipewatch** — 40 items
- **nathanjohnpayne/tadlockpsychiatry** — 57 items

## Findings

Severity totals — P0: 21 · P1: 477 · P2: 144 · P3: 72 · Unknown: 5  (showing at most 60 items inline; full list in the workflow artifact)

<details>
<summary>P0 — nathanjohnpayne/matchline (6, showing 6)</summary>

- [#268 · sync: bulk reconcile to mergepath@314947f](https://github.com/nathanjohnpayne/matchline/pull/268#discussion_r3382026656) — `coderabbitai`: _⚠️ Potential issue_ | _🔴 Critical_  <details> <summary>🧩 Analysis chain</summary>  🏁 Script executed:  ```shell #!/bin/
- [#243 · sync: bulk reconcile to mergepath@2ea7267](https://github.com/nathanjohnpayne/matchline/pull/243#discussion_r3293233525) — `coderabbitai`: _⚠️ Potential issue_ | _🔴 Critical_ | _⚡ Quick win_  <details> <summary>🧩 Analysis chain</summary>  🏁 Script executed:  
[...]

PR #479, opened 2026-06-17
fix(playground): control labels, preset reset, path/spec drift, stacked scroll-sync (#470)

### LINKED ISSUE #470: Fix remaining Mergepath Playground control, preset, and spec drift
Source rollup: nathanjohnpayne/mergepath#456

Validated against Mergepath current main. Several playground findings still apply after the path rename.

Current evidence:
- Stacked-layout scroll sync is still wired where it should be handled or disabled.
- Controls lack programmatic labels.
- Default protected paths drift from `.github/review-policy.yml`.
- Presets do not reset full policy state.
- The spec omits implemented backdrop-click modal close behavior.

Representative source findings:
- https://github.com/nathanjohnpayne/mergepath/pull/81#discussion_r3101272420
- https://github.com/nathanjohnpayne/mergepath/pull/78#discussion_r3097841138
- https://github.com/nathanjohnpayne/mergepath/pull/78#discussion_r3097841140
- https://github.com/nathanjohnpayne/mergepath/pull/78#discussion_r3097841153
- https://github.com/nathanjohnpayne/mergepath/pull/78#discussion_r3097841155

nathanjohnpayne:
Closes #470. Five Mergepath Playground findings from the #456 rollup (the issue sixth item, the shell/JS glob-semantics mismatch, was already resolved by #471 which aligned match_protected_paths.sh to the Playground compileGlob).

- Default protected paths drifted from .github/review-policy.yml: replaced the stale credentials/** with the actual policy patterns **/*secret* and **/*credential*, reordered to match external_review_paths, and added a keep-in-sync comment. A synthetic PR path now hits **/*credential* so the demo still exercises the pattern.
- Presets did not reset full policy state: applyPreset overlaid only the four fields a preset defines, leaving protected paths and reviewer toggles stale. Now resets the entire state to defaults first, applies the preset, and re-syncs every control.
- Controls lacked programmatic labels: added aria-label to the threshold and codex-rounds sliders and the CodeRabbit/Codex switch checkboxes (their text is in a sibling div, not the wrapping label).
- Stacked-layout scroll-sync: the rail/workspace mirror was wired unconditionally; now gated on matchMedia max-width 960px, read live per scroll.
- Spec omitted the implemented backdrop-click modal close: documented it alongside Escape.

REVIEW FOCUS: the preset full-reset and the protected-path policy alignment.

Authoring-Agent: claude

## Self-Review
- Correctness: verified in-browser against a static server. aria-labels present on all four controls; chips and rendered YAML match the five policy paths; applying Strict removes an added path, re-checks an unchecked reviewer, sets threshold 100; at 375px the matchMedia gate engages and scrolling the rail does NOT mirror the workspace; no console errors.
- Regression risk: the scroll-sync gate is read live per scroll (no detach/reattach across resize); the preset reset uses structuredClone of DEFAULTS with a JSON fallback; the protected-path change is data-only and the matcher is unchanged.
- Style and conventions: matches the existing single-file Playground style; aria-label phrasing mirrors the visible control copy.
- Test coverage: this is a static browser artifact with no unit harness; verified via the preview workflow (snapshot, eval, resize) plus a screenshot.
- Security and dependency hygiene: no new dependencies; no network or auth; the path change brings the demo into line with the real policy so it cannot under-represent the gate.

### DIFF
--- a/mergepath/playground/index.html
+++ b/mergepath/playground/index.html
@@ -800,7 +800,7 @@
             <span>Lines changed at or above this value escalate to Phase 4.</span>
           </div>
           <div class="slider-row">
-            <input type="range" id="threshold" min="50" max="1000" step="10" value="300" />
+            <input type="range" id="threshold" min="50" max="1000" step="10" value="300" aria-label="External review threshold (lines changed)" />
             <span class="val"><span id="thresholdVal">300</span> lines</span>
           </div>
         </div>
@@ -830,7 +830,7 @@
             <span>Advisory auto-review on PR open.</span>
           </div>
           <label class="switch">
-            <input type="checkbox" data-toggle="coderabbit" checked />
+            <input type="checkbox" data-toggle="coderabbit" checked aria-label="Enable CodeRabbit advisory auto-review" />
             <span class="track"></span>
           </label>
         </div>
@@ -841,13 +841,13 @@
             <span>Phase 4a automated external review.</span>
           </div>
           <label class="switch">
-            <input type="checkbox" data-toggle="codex" checked />
+            <input type="checkbox" data-toggle="codex" checked aria-label="Enable Codex GitHub App external review" />
             <span class="track"></span>
           </label>
         </div>
         <div class="nested shown" id="codexNested">
           <div class="slider-row">
-            <input type="range" id="codexRounds" min="1" max="5" step="1" value="2" />
+            <input type="range" id="codexRounds" min="1" max="5" step="1" value="2" aria-label="Codex maximum review rounds" />
             <span class="val">max <span id="codexRoundsVal">2</span> rounds</span>
           </div>
           <p class="hint">Beyond this, Mergepath stops the loop and pings a human.</p>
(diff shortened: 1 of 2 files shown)


nathanpayne-codex (changes requested):
[P2] Update the playground spec for the stacked-layout scroll rule

`mergepath/playground/index.html` now intentionally disables rail/workspace scroll mirroring when `matchMedia('(max-width: 960px)')` matches, but `specs/mergepath_playground.md:29` still says the control rail and workspace "scroll in sync" unconditionally. Since #470 includes the stacked-layout scroll-sync hardening, the spec should describe the new contract: sync only applies in the side-by-side layout, and stacked layout panes scroll independently. As written, the spec still encodes the old behavior and can drive a future regression back toward syncing on mobile.

Validation run on HEAD `2b9f792`: live PR state/checks read back, 0 unresolved threads, `git diff --check`, `scripts/ci/check_spec_test_alignment`, runtime script `node --check`, and localhost browser smoke for labels, preset reset, YAML reset, modal backdrop close, and stacked media-query behavior.

nathanpayne-claude:
r2 (ddeaf2b) addresses the spec finding: specs/mergepath_playground.md now states the rail/workspace scroll-sync is side-by-side (desktop) only, with the stacked panes scrolling independently at the max-width: 960px breakpoint (gated on a matchMedia check read live per scroll), matching index.html. check_spec_test_alignment passes.

nathanpayne-codex (approved):
Approved re-review of ddeaf2b. The requested spec fix now documents side-by-side-only rail/workspace scroll sync and independent stacked scrolling at the max-width: 960px breakpoint, matching the index.html matchMedia gate. Verified CodeRabbit cleared the current head, git diff --check, scripts/ci/check_spec_test_alignment, tests/test_mergepath_playground.sh, and Chrome desktop/mobile static render smoke.

PR #522, opened 2026-06-21
[codex] Grant Firebase deployer secret metadata access

nathanjohnpayne:
## Summary
- Add `roles/secretmanager.viewer` to the canonical `op-firebase-setup` deployer role set.
- Document the role in the Firebase deployment setup contract.
- Pin the role with the existing wired Firebase bootstrap regression test.

## Root Cause
Firebase Functions v2 deploy analysis reads Secret Manager metadata for functions using `defineSecret()`. Repos with Functions secrets can fail at deploy time with `secretmanager.secrets.get` denied when the Firebase deployer service account lacks `roles/secretmanager.viewer`.

## Validation
- `bash tests/test_bootstrap_firebase_and_codereview.sh`

## Self-Review

- Correctness: The setup helper now grants the Secret Manager metadata role required by Firebase Functions v2 secret analysis, matching the reproduced deploy failure.
- Regression risk: Low; the change only adds one IAM role to the deployer role list and updates matching documentation.
- Style/conventions: Follows the existing role-list pattern in `op-firebase-setup` and the existing bootstrap test style.
- Test coverage: `tests/test_bootstrap_firebase_and_codereview.sh` now pins the role using fixed-string matching.
- Security/dependency hygiene: The role is metadata read-only (`secretmanager.viewer`), not secret payload access, and no credentials are committed.

### DIFF
--- a/DEPLOYMENT.md
+++ b/DEPLOYMENT.md
@@ -641,6 +641,7 @@ FIREBASE_IMPERSONATION_MEMBER=email@example.com op-firebase-setup {project-id}
 3. Grants the deployer service account these project roles:
    - `roles/firebase.admin`
    - `roles/cloudfunctions.admin`
+   - `roles/secretmanager.viewer`
    - `roles/iam.serviceAccountUser`
    - `roles/artifactregistry.writer`
    - `roles/run.admin`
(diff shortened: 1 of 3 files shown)


nathanpayne-codex (dismissed):
Approved. This is a narrow deploy-template fix: Firebase Functions v2 secret analysis needs Secret Manager metadata reads, and the existing wired bootstrap test now pins that role in the setup helper.

nathanpayne-codex (approved):
Approved current head. The CodeRabbit finding was addressed by switching the role assertion to fixed-string grep, and the wired Firebase bootstrap test remains green.

Issue #495, opened 2026-06-17
Re-arm auto-merge on synchronize when an approval already exists (fix one-shot arming)


nathanjohnpayne:
## Problem

Auto-merge frequently "doesn't fire" — a PR goes `CLEAN` (CI green, approved, threads resolved) but sits unmerged until someone merges by hand. Observed repeatedly; most recently #494 (`autoMergeRequest` was null the entire time; had to be direct-merged).

## Root cause: arming is one-shot, on the approval event only

The `auto-merge-on-approval` job in `.github/workflows/agent-review.yml` (job at ~line 432) is gated on:

```yaml
if: >
  github.event_name == 'pull_request_review' &&
  github.event.review.state == 'approved' && ...
```

It runs **only on an approval event — never on `synchronize` (push)**. Its final step (~line 607) calls `gh pr merge --squash --auto`. On a blocking finding it logs (~line 550): *"address the findings and request a new approval to re-trigger."*

This interacts badly with **bot-review latency**. Measured across the last 40 merged PRs:

- **CodeRabbit**: ~2–8 min from open (fast, low variance) — already handled by `coderabbit-wait.sh` + the #419 eyes-ack re-trigger.
- **Codex** (`chatgpt-codex-connector`): **15–40+ min**, with dropped-trigger outliers in the hundreds of minutes. CI goes green in ~1–2 min.

So a Codex finding routinely lands **at or after** the approval. The arming run then either can't enable `--auto` (GitHub refuses once a required check has gone red) or `--auto` was set and the gate blocks. Addressing the finding is a **fix-commit (synchronize)** — which does **not** re-trigger the arming job. Unless the author **re-approves**, auto-merge never (re-)arms, and the PR stalls `CLEAN`-but-unmerged.

(Note: GitHub's native auto-merge, once enabled, *does* persist across pushes and waits for the new HEAD's checks. The failure mode here is the **initial arming attempt failing** at approval time — there is no retry, because the job is approval-only.)

## Proposed structural fix

Add a **`synchronize` re-trigger** to the auto-merge arming path, guarded on an **existing valid approval**, so a fix-commit retries arming automatically:

- Re-run the arming logic on `pull_request` `synchronize` (and `reopened`) **when** the PR already carries a non-author `APPROVED` review and none of the blocking labels (`needs-external-review`, `needs-human-review`, `human-hold`).
[...]

PR #83, opened 2026-04-17
refactor: rename mockups/ to mergepath/ and mergepath.html to index.html

nathanjohnpayne:
## Summary

- Renames `mockups/` → `mergepath/` to align the directory with the product name
- Renames `mockups/mergepath.html` → `mergepath/index.html` so the entry point is a standard `index.html`
- Updates all path references in README, specs, plans, tests, scripts, and `.repo-template.yml`

Authoring-Agent: claude

## Self-Review

- **Correctness:** All nine files with `mockups` path references have been updated. `git mv` was used so rename history is preserved. No logic changed — pure path rename.
- **Regression risk:** Low. `scripts/policy-sim.sh` and `tests/test_mergepath_frontend.sh` both point to the new `mergepath/index.html` path. The test should pass unchanged.
- **Style:** Follows existing naming conventions; `mergepath/` mirrors the product name used throughout docs.
- **Test coverage:** No new tests needed — existing `tests/test_mergepath_frontend.sh` covers the renamed file.
- **Security/deps:** No dependencies changed.

PR #525, opened 2026-06-23
fix(ci): kill SIGPIPE flake in check_workflow_parsers re-arm assertions (#520)

### LINKED ISSUE #520: Flaky check_workflow_parsers: printf|grep -Fq SIGPIPE race under pipefail (#495 re-arm assertions)
Surfaced during the fabb2b1 propagation wave (after the #519 closure-gap fix).

## Symptom
`scripts/ci/check_workflow_parsers` flaked on the tadlockpsychiatry sync PR (#83) `lint` while the byte-identical payload passed on the 7 other consumers in the same wave:

    ./scripts/ci/check_workflow_parsers: line 615: printf: write error: Broken pipe
    FAIL: re-arm gate must collapse to the latest review per reviewer ... (#495)
    check_workflow_parsers: 1 FAIL / 76 PASS

## Root cause
The #495 re-arm assertions (lines ~604-628) test with `printf %s\n "$auto_merge_job_block" | grep -Fq PATTERN`. Because `grep -q` exits on first match, `printf` can be killed by SIGPIPE before it finishes writing; under pipefail the pipeline then returns non-zero, the `&&` short-circuits, and the assertion FAILs spuriously. Timing-dependent, so flaky and fleet-wide (any consumer, or mergepath itself, can hit it on any run).

## Proposed fix
Replace the `printf ... | grep -Fq` pattern in those assertions with a non-pipe match, e.g. `grep -Fq PATTERN <<<"$auto_merge_job_block"` (here-string, no SIGPIPE) or `[[ "$auto_merge_job_block" == *PATTERN* ]]`, then re-propagate.

## Immediate handling
tadlockpsychiatry#83 lint re-run was triggered; the assertion passes on retry. This issue tracks the source hardening so the flake stops recurring.

nathanjohnpayne:
Authoring-Agent: claude

Closes #520.

## Summary
- `scripts/ci/check_workflow_parsers` matched workflow blocks with `printf '%s\n' "$VAR" | grep -Fq PATTERN`. `grep -q` exits on first match, so `printf` can be killed by **SIGPIPE** before it finishes writing; under `set -o pipefail` the pipeline returns non-zero, the `&&` short-circuits, and the assertion **FAILs spuriously**. Timing-dependent → fleet-wide flake (hit tadlockpsychiatry sync PR #83 `lint` on the fabb2b1 wave; the byte-identical payload passed on the 7 other consumers).
- Fix: replace every executed `printf '%s\n' "$VAR" | grep -...q PATTERN` with a here-string match `grep -...q PATTERN <<<"$VAR"` — no pipe, no SIGPIPE. Here-string semantics are identical for matching (both append exactly one trailing newline).
- #520 named only the #495 re-arm block, but the identical anti-pattern appears in 14 if-blocks (lines 508–706). Converted all **26** instances so the flake is removed everywhere, not just the named block.
- No behavior change to the workflows under test — source-hardening of the test only.

### Deliberately unchanged
- String-literal *expectation* at line 766 (`"if ! printf '%s\\n' ... | grep -Fxq ..."`) — asserted data, not an executed pipeline.
- Capturing fixtures at lines 140/142 (`grep -c`, `grep | head`) — read all input, so they never SIGPIPE.

## Testing
- [x] `bash -n scripts/ci/check_workflow_parsers` → syntax OK
- [x] `bash scripts/ci/check_workflow_parsers` → `PASS (78 tests)` (unchanged count)
- [x] Manual verification completed when needed

## Self-Review
- [x] Correctness: here-string match is byte-for-byte equivalent for `grep -q`; the racy pipe is the only thing removed
- [x] Regression risk: pure test-internal refactor; the asserted workflow contracts are untouched and the PASS count holds at 78
- [x] Style and conventions: matches the surrounding here-string idiom; literal/capturing greps left as-is
- [x] Test coverage: the check is itself the test; re-run green
- [x] Security and dependency hygiene: no new deps, no new external calls

### DIFF
--- a/scripts/ci/check_workflow_parsers
+++ b/scripts/ci/check_workflow_parsers
@@ -505,7 +505,7 @@ enable_auto_merge_block=$(awk '
   in_block && /- name:/ && $0 !~ /- name: Enable auto-merge/ {exit}
   in_block {print}
 ' "$agent_review_workflow")
-if printf '%s\n' "$enable_auto_merge_block" | grep -Fq 'GH_TOKEN: ${{ secrets.AUTHOR_MERGE_TOKEN }}'; then
+if grep -Fq 'GH_TOKEN: ${{ secrets.AUTHOR_MERGE_TOKEN }}' <<<"$enable_auto_merge_block"; then
   pass=$((pass + 1))
   echo "  PASS: Enable auto-merge step runs under AUTHOR_MERGE_TOKEN"
 else
@@ -513,7 +513,7 @@ else
   echo "  FAIL: Enable auto-merge step must set GH_TOKEN from AUTHOR_MERGE_TOKEN" >&2
 fi
 
-if printf '%s\n' "$enable_auto_merge_block" | grep -Fq 'REVIEWER_ASSIGNMENT_TOKEN'; then
+if grep -Fq 'REVIEWER_ASSIGNMENT_TOKEN' <<<"$enable_auto_merge_block"; then
   fail=$((fail + 1))
   echo "  FAIL: Enable auto-merge step still references REVIEWER_ASSIGNMENT_TOKEN" >&2
 else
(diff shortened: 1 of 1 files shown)


nathanpayne-claude (approved):
Reviewed under nathanpayne-claude. Mechanical, root-cause source-hardening: every executed `printf '%s\n' "$VAR" | grep -...q` is converted to an equivalent here-string match, eliminating the SIGPIPE-under-pipefail race. The string-literal expectation (L766) and the capturing `grep -c`/`grep | head` fixtures (L140/142) are correctly left alone. check_workflow_parsers holds at PASS (78 tests); no workflow contract changed. CI green, Codex 👍 on HEAD. Approving for merge.

Issue #521, opened 2026-06-19
Closure gap: coderabbit-rate-limit failover files unpropagated + closure-completeness CI guard (fabb2b1 wave)


nathanjohnpayne:
## Closure gap: coderabbit-rate-limit failover files not propagated (#489)

Surfaced by the `@codex review` pass on the **fabb2b1** propagation wave (all 8 consumer sync PRs). Codex (P2) correctly flagged that three files referenced by *propagated* content are absent on consumers because they are not declared in `.mergepath-sync.yml`:

| File | Referenced by (propagated) | In manifest? |
|---|---|---|
| `scripts/coderabbit-automerge-rate-limit-gate.sh` | `.github/workflows/agent-review.yml:727` | ❌ |
| `tests/test_coderabbit_wait_codex_failover.sh` | `scripts/ci/check_coderabbit_wait:11` | ❌ |
| `tests/test_coderabbit_automerge_rate_limit_gate.sh` | `scripts/ci/check_coderabbit_wait:12` | ❌ |

### Impact (why P2 / fail-safe)
- On consumers, `agent-review.yml`'s CodeRabbit-rate-limit (exit 5) auto-merge branch calls the missing gate script → exit 127 → the `else` blocks auto-merge (`exit 1`). **Fail-safe**: a rate-limited PR blocks rather than misbehaves; it just loses the #489 nuanced disposition.
- `check_coderabbit_wait` references the two missing tests, but consumer `lint` does not execute that check today, so it does not break consumer CI (it would if the check is ever wired into consumer lint).

### Fix
Declare the three files in `.mergepath-sync.yml` (canonical `- path:` + the `scripts/ci/` `requires:` closure for the two tests), same shape as #519. They then propagate on the next wave.

## Systemic: closure-completeness CI guard
This is the **3rd** referenced-but-undeclared closure gap (after #519). The `requires:` closure is hand-maintained and `check_sync_manifest` only validates that *listed* entries have covering paths — it cannot detect a *missing* one. Add a CI guard that diffs every `scripts/ci/check_*` and propagated-workflow file's referenced `tests/`+`scripts/` set against the manifest, failing mergepath CI when a propagated reference is undeclared. Must allow-list orchestrator-only refs (bootstrap*, sync-machinery tests, `scripts/workflow/` kit, etc.).

## Other P2 suggestions from the same Codex pass (triage)
[...]

PR #519, opened 2026-06-19
fix(sync): declare test_codex_review_request_trigger_only in manifest

nathanjohnpayne:
## Summary

Fixes a propagation closure gap surfaced by the **swipewatch canary** on the `2120dec` `--sync-all` wave (canary PR nathanjohnpayne/swipewatch#72): every consumer's `lint` job failed because `scripts/ci/check_codex_scripts` died with exit 127 / `No such file`.

## Root cause

`scripts/ci/check_codex_scripts` hard-runs six test files. Five are declared in `.mergepath-sync.yml`; the sixth — `tests/test_codex_review_request_trigger_only.sh`, added by #489 (the CodeRabbit-rate-limit Codex failover) — was never declared, so it never propagated. The check then fails pre-flight on every consumer.

The `scripts/ci/` `requires:` closure that exists to prevent exactly this (#264) is hand-maintained, and `check_sync_manifest` only validates that *listed* entries have covering paths — it cannot detect a *missing* entry. mergepath's own CI stayed green because the test exists locally.

## Fix

- Declare `tests/test_codex_review_request_trigger_only.sh` as a canonical path (`consumers: all`).
- Add it to the `scripts/ci/` `requires:` closure.
- Also add `tests/test_codex_review_request_entry.sh` (#486) to the `requires:` closure — the same latent gap, not yet failing only because it already propagates via its own path entry.

All six `check_codex_scripts` test deps now carry both a `- path:` entry and a `requires:` entry.

## Verification

- `scripts/ci/check_sync_manifest` → PASS (27 tests, 8 consumers, 71 path entries).
- `scripts/ci/check_codex_scripts` passes locally (test present in mergepath).
- Next: re-run the swipewatch canary on this HEAD; green → resume the riskiest→least fan-out.

## Self-Review

- **Correctness:** All six tests referenced by `scripts/ci/check_codex_scripts` (ack, trigger_only, entry, resolution, 465, record_feedback) now resolve on every consumer; `test_codex_review_request_trigger_only.sh` exists in mergepath HEAD and is now declared. `check_sync_manifest` passes (27 tests, 71 path entries).
- **Regression risk:** Minimal — additive, declarative manifest entries only (one new canonical path + two `requires:` lines). No script/workflow logic touched. Worst case is one extra test file in each consumer's sync payload, which is the intended outcome.
[...]

Issue #489, opened 2026-06-17
coderabbit-wait + codex: auto-request @codex while CodeRabbit is rate-limited, revert when the window elapses


nathanjohnpayne:
## Summary

When CodeRabbit reports it's **rate-limited**, automatically fail over to Codex —
request `@codex review` so the PR keeps moving instead of stalling on
CodeRabbit's retry window — and **revert** (stop the auto-Codex behavior) once the
rate-limit window has elapsed / CodeRabbit is available again.

## Background

`scripts/coderabbit-wait.sh` already detects CodeRabbit's rate-limit state:

- It matches the rate-limit comment (`/Rate limit exceeded/i`), **parses the retry
  window** CodeRabbit publishes (`"X minutes and Y seconds"` / `"X seconds"`),
  sleeps that window + a 30s buffer, posts `@coderabbitai, try again.`, and
  retries up to `coderabbit.max_rate_limit_retries` (default 2).
- If retries are exhausted it exits **5** (`rate_limit_stalled`), which AGENTS.md
  treats as "alert the human, do not proceed."

So today a rate-limited CodeRabbit either (a) costs us the full window × N retries
of wall-clock latency, or (b) dead-ends at exit 5 — even though **Codex is our
actual blocking gate** (CodeRabbit is advisory; `request_changes_workflow: false`).
We shouldn't sit idle waiting on an advisory bot's hourly allowance.

Per CodeRabbit's [rate-limit docs](https://docs.coderabbit.ai/management/plans#rate-limits),
limits are **per-developer, per-hour, refillable** (Free 3/hr, Pro 5/hr, Pro+
10/hr, Enterprise 12/hr) — and `@coderabbitai rate limit` reports remaining
allowance without consuming a review. The authoritative "how long until reset" for
a given stall is the retry window in CodeRabbit's own comment, which
`coderabbit-wait.sh` already parses.

## What to add

A **time-boxed Codex failover** keyed off the rate-limit state:

1. **On rate-limit detection** (the existing branch in `coderabbit-wait.sh`, or a
   caller consuming its `status: rate_limit` / exit 5): request `@codex review` via
   `scripts/codex-review-request.sh` so Phase 4a proceeds instead of idling. This
   is effectively flipping the #486 "request `@codex` by default" behavior **on**
   for the duration of the stall.
2. **On window elapse** (the parsed retry window expires, CodeRabbit's review
   lands, or the PR clears): **revert** — stop forcing the Codex request so the
[...]

PR #526, opened 2026-06-24
fix(sync): declare coderabbit-rate-limit failover files in manifest (#521)

nathanjohnpayne:
Authoring-Agent: claude

Part of #521 (manifest-declaration half; the closure-completeness guard lands separately).

## Summary
Three files referenced by **propagated** content were absent from `.mergepath-sync.yml`, so they never travel to consumers — the 3rd such gap (same closure class as #519):

| File | Referenced by (propagated) | Was declared? |
|---|---|---|
| `scripts/coderabbit-automerge-rate-limit-gate.sh` | `.github/workflows/agent-review.yml` | ❌ |
| `tests/test_coderabbit_wait_codex_failover.sh` | `scripts/ci/check_coderabbit_wait` | ❌ |
| `tests/test_coderabbit_automerge_rate_limit_gate.sh` | `scripts/ci/check_coderabbit_wait` | ❌ |

**Impact (P2 / fail-safe):** on consumers, agent-review.yml's CodeRabbit-rate-limit (exit 5) auto-merge branch calls the missing gate → exit 127 → the `else` blocks auto-merge. The PR fails *safe* (blocks rather than misbehaves) but loses the #489 nuanced disposition. `check_coderabbit_wait` references the two tests, but consumer `lint` does not run that check today, so consumer CI is not broken — it would be if the check were ever wired into consumer lint.

**Fix (same shape as #519):**
- declare all three as canonical `- path:` (`consumers: all`)
- add the two tests to the `scripts/ci/` `requires:` closure
- add the gate script as a transitive `requires:` entry (the gate test runs it, mirroring the `gh-pr-guard.sh` precedent)

They propagate on the next wave (no propagation triggered by this PR).

## Testing
- [x] `yq` parse OK
- [x] `bash scripts/ci/check_sync_manifest` → PASS (74 path entries; was 71)
- [x] `bash scripts/ci/check_coderabbit_wait` → PASS
- [x] Manual: confirmed all three files now grep-match in the manifest

## Self-Review
- [x] Correctness: each declared path exists on disk and is referenced by the cited propagated surface; comments name the exact referencing file
- [x] Regression risk: additive manifest entries only; no path removed or retyped; check_sync_manifest's union-coverage assertions still pass
- [x] Style and conventions: mirrors the #519 declaration shape (canonical path + requires closure + transitive "runs it" note)
- [x] Test coverage: exercised by check_sync_manifest + check_coderabbit_wait
- [x] Security and dependency hygiene: documentation/manifest only; no code or deps changed

## Remaining for #521
The **closure-completeness CI guard** (the "Systemic:" section — diff every `scripts/ci/check_*` + propagated-workflow referenced `tests/`+`scripts/` set against the manifest, allow-listing orchestrator-only refs) is the durable prevention for this recurring class and lands as a follow-up PR. #521 stays open until it does.

### DIFF
--- a/.mergepath-sync.yml
+++ b/.mergepath-sync.yml
@@ -290,6 +290,15 @@ paths:
     # token-derived expected-reviewer allow-list; exits 3 if absent.
     requires:
       - "scripts/lib/reviewers-helpers.sh"
+  # #489 CodeRabbit-rate-limit auto-merge disposition gate. agent-review.yml's
+  # exit-5 (rate-limited) branch calls this script to decide whether the
+  # Codex-failover path may auto-merge. Absent on a consumer it exits 127 and
+  # the workflow's `else` blocks auto-merge — fail-safe, but the PR loses the
+  # #489 nuanced disposition. Referenced by .github/workflows/agent-review.yml;
+  # undeclared by #489, surfaced by the fabb2b1 wave's @codex pass (#521).
+  - path: scripts/coderabbit-automerge-rate-limit-gate.sh
+    type: canonical
+    consumers: all
   - path: scripts/codex-review-request.sh
     type: canonical
     consumers: all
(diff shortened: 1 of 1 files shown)


nathanpayne-claude (approved):
Reviewed under nathanpayne-claude. Additive manifest declaration of three referenced-but-undeclared files (gate script + two failover tests), mirroring the #519 shape: canonical `- path:` + scripts/ci/ `requires:` closure for the tests + transitive `requires:` for the gate script. Each path exists on disk and is named by the cited propagated surface. check_sync_manifest PASS (74 entries), check_coderabbit_wait PASS. No propagation triggered. CI green, Codex 👍 on HEAD. Approving for merge.

